PR #1, opened 2020-08-25
create kotlin instance from KtScript

piiertho:
Create new KotlinInstance from KotlinScript::instance_create, KotlinInstance constructor call create_instance of KtClass, which has to be implemented

### DIFF
--- a/src/kotlin_instance.cpp
+++ b/src/kotlin_instance.cpp
@@ -1,5 +1,15 @@
 #include "kotlin_instance.h"
 #include "kotlin_language.h"
+#include "kt_class.h"
+
+KotlinInstance::KotlinInstance(KtClass *ktClass, Object *p_owner) :
+ktObject(ktClass->create_instance(jni::Jvm::current_env(), nullptr, 0, p_owner)) {
+
+}
+
+KotlinInstance::~KotlinInstance() {
+    delete ktObject;
+}
 
 bool KotlinInstance::set(const StringName& p_name, const Variant& p_value) {
     return false;
--- a/src/kotlin_instance.cpp
+++ b/src/kotlin_instance.cpp
@@ -2,8 +2,12 @@
 #include "kotlin_language.h"
 #include "kt_class.h"
 
-KotlinInstance::KotlinInstance(const jni::JObject &wrappedObject) : jObject(wrappedObject) {
+KotlinInstance::KotlinInstance(const jni::JObject &wrappedObject) : jObject(wrappedObject), owner(nullptr) {
 
+}
+
+KotlinInstance::KotlinInstance(const jni::JObject &wrappedObject, Object *owner) : KotlinInstance(wrappedObject) {
+    set_owner(owner);
 }
 
 bool KotlinInstance::set(const StringName& p_name, const Variant& p_value) {
@@ -23,7 +27,11 @@ Variant::Type KotlinInstance::get_property_type(const StringName& p_name, bool* 
 }
 
 Object* KotlinInstance::get_owner() {
-    return ScriptInstance::get_owner();
+    return owner;
+}
+
+void KotlinInstance::set_owner(Object *object) {
+    owner = object;
 }
 
 void KotlinInstance::get_property_state(List<Pair<StringName, Variant>>& state) {
--- a/src/kotlin_instance.cpp
+++ b/src/kotlin_instance.cpp
@@ -2,13 +2,8 @@
 #include "kotlin_language.h"
 #include "kt_class.h"
 
-KotlinInstance::KotlinInstance(KtClass *ktClass, Object *p_owner) :
-ktObject(ktClass->create_instance(jni::Jvm::current_env(), nullptr, 0, p_owner)) {
+KotlinInstance::KotlinInstance(const jni::JObject &wrappedObject) : jObject(wrappedObject) {
 
-}
-
-KotlinInstance::~KotlinInstance() {
-    delete ktObject;
 }
 
 bool KotlinInstance::set(const StringName& p_name, const Variant& p_value) {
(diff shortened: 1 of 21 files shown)


src/kotlin_instance.cpp
```diff
@@ -1,5 +1,15 @@
 #include "kotlin_instance.h"
 #include "kotlin_language.h"
+#include "kt_class.h"
+
+KotlinInstance::KotlinInstance(KtClass *ktClass, Object *p_owner) :
+ktObject(ktClass->create_instance(jni::Jvm::current_env(), nullptr, 0, p_owner)) {
```
raniejade, line 6: KotlinInstance should accept `jni::JObject` param instead instead of calling create instance here.

KtObject should also store a pointer to a `KtClass` that created it.

src/kotlin_instance.h
```diff
@@ -1,11 +1,15 @@
...
+#include "kt_object.h"
+#include "kt_class.h"
 
 class KotlinInstance : public ScriptInstance {
+private:
+    KtObject *ktObject;
 public:
-    KotlinInstance() = default;
-    ~KotlinInstance() override = default;
+    KotlinInstance(KtClass *ktClass, Object *p_owner);
```
raniejade, line 11: The script owner can be change IIRC. So this should be a setter instead, not part of the constructor

raniejade (changes requested):

src/kt_class.h
```diff
@@ -1,10 +1,10 @@
...
 #define GODOT_JVM_KTCLASS_H
 
-#include "jni/jvm.h"
-#include "core/reference.h"
-#include "java_instance_wrapper.h"
-#include "kt_object.h"
+#include <jni/jvm.h>
+#include <core/reference.h>
+#include <java_instance_wrapper.h>
+#include <kt_object.h>
```
raniejade, line 7: Let's stick with `"header.h"` for classes in our module and use `<header.h>` for classes outside of it.

src/kt_object.cpp
```diff
@@ -1,3 +1,8 @@
 #include "kt_object.h"
 
-KtObject::KtObject(jni::JObject p_wrapped, jni::JObject p_class_loader) : JavaInstanceWrapper("godot.core.KtObject", p_wrapped, p_class_loader) {}
+KtObject::KtObject(const jni::JObject &p_wrapped, const jni::JObject &p_class_loader, jni::JObject *p_ktClass)
```
raniejade, line 3: `jni::JObject *p_ktClass` shouldn't be a pointer, `jni::JObject p_ktClass` should suffice. `JObject` just wraps `jobject` which is a pointer itself.

src/kt_class.cpp
```diff
@@ -7,10 +7,10 @@ KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) : JavaIns
 }
 
 KtObject* KtClass::create_instance(jni::Env& env, const Variant** p_args, int p_arg_count, Object* p_owner) {
-    auto new_method = get_method_id(env, "new", "(J[Lgodot/core/KtVariant;)Lgodot/core/KtObject;");
+    jni::MethodId new_method = get_method_id(env, "new", "(J[Lgodot/core/KtVariant;)Lgodot/core/KtObject;");
     // TODO: send args
-    auto j_kt_object = wrapped.call_object_method(env, new_method, {reinterpret_cast<long>(p_owner)});
-    return new KtObject(j_kt_object, class_loader);;
+    const jni::JObject &j_kt_object = wrapped.call_object_method(env, new_method, {reinterpret_cast<long>(p_owner)});
```
raniejade, line 12: the returned object is not a reference nor is const: so it should be `jni::JObject j_kt_object`.

raniejade (changes requested):

src/kt_class.cpp
```diff
@@ -7,10 +7,10 @@ KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) : JavaIns
 }
 
 KtObject* KtClass::create_instance(jni::Env& env, const Variant** p_args, int p_arg_count, Object* p_owner) {
-    auto new_method = get_method_id(env, "new", "(J[Lgodot/core/KtVariant;)Lgodot/core/KtObject;");
+    jni::MethodId new_method { get_method_id(env, "new", "(J[Lgodot/core/KtVariant;)Lgodot/core/KtObject;") };
```
raniejade, line 10: compiler can elide this right? the line is a lot more longer now.

src/kotlin_script.cpp
```diff
@@ -26,7 +26,9 @@ StringName KotlinScript::get_instance_base_type() const {
 }
 
 ScriptInstance* KotlinScript::instance_create(Object* p_this) {
-    return memnew(KotlinInstance);
+    jni::Env env = jni::Jvm::current_env();
+    KtObject *wrapped = get_kt_class()->create_instance(env, nullptr, 0, p_this);
```
raniejade, line 30: `KtObject* wrapped` :P

raniejade (approved):

src/kt_object.h
```diff
@@ -1,12 +1,18 @@
...
 class KtObject : public JavaInstanceWrapper {
+
+private:
+    const jni::JObject ktClass;
 public:
-    KtObject(jni::JObject p_wrapped, jni::JObject p_class_loader);
+    KtObject(jni::JObject p_wrapped, jni::JObject p_class_loader, jni::JObject p_ktClass);
     ~KtObject() = default;
+
+    const jni::JObject &get_wrapped() const;
```
raniejade, line 15: you don't really need this, most of the time you will be only interacting with `KtObject`

PR #2, opened 2020-08-26
Fix compilation incomplete type compilation error due to forward declare

piiertho:
This needs to be tested.
I got it compiling on linux with those changes (still does not compile on windows)
Forward declarations were causing troubles, and protobuf was not linking.

### DIFF
--- a/SCsub
+++ b/SCsub
@@ -13,8 +13,12 @@ protobuf_include_dir = [
 ]
 
 protobuf_lib_dir = Dir("libprotobuf/lib").path
-env.Append(LINKFLAGS="-L" + protobuf_lib_dir)
-env.Append(LIBS="protobuf")
+if env["platform"] == "windows":
+    env.Append(LINKFLAGS="-LIBPATH:" + protobuf_lib_dir)
+    env.Append(LINKFLAGS="-DEFAULTLIB:libprotobuf")
+else:
+    env.Append(LINKFLAGS="-L" + protobuf_lib_dir)
+    env.Append(LIBS="protobuf")
 
 module_env = env.Clone()
 module_env.Append(CPPPATH=[java_include_dirs, protobuf_include_dir])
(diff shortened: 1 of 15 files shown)


src/jni/env.h
```diff
@@ -41,6 +41,13 @@ namespace jni {
         friend class JArray;
         friend class JObjectArray;
     };
+
+    template <class T>
+    inline T JObject::new_global_ref(Env& env) {
```
piiertho, line 46: Not sure that's a good way to solve the error

SCsub
```diff
@@ -13,8 +13,12 @@ protobuf_include_dir = [
 ]
 
 protobuf_lib_dir = Dir("libprotobuf/lib").path
-env.Append(LINKFLAGS="-L" + protobuf_lib_dir)
-env.Append(LIBS="protobuf")
+if env["platform"] == "windows":
+    env.Append(LINKFLAGS="-LIBPATH:" + protobuf_lib_dir)
```
raniejade, line 17: ohhh, so it's a `-` not `/`? The microsoft docs says it's a slash XD

src/kt_variant.h
```diff
@@ -3,6 +3,10 @@
 #include "wire/wire.pb.h"
 #include "core/variant.h"
 
+// must match the value order of godot_variant_type
+static void(*TO_KT_VARIANT_FROM[27 /* Variant::Type count */])(wire::Value&, const Variant&);
```
raniejade, line 7: does it need to be in the header? I mean clients of the class shouldn't really know about this arrays.

raniejade (approved):

SCsub
piiertho: It replace it by /, do you want I put / instead ?

src/kt_variant.h
piiertho: yep that's true, let me change this.

SCsub
raniejade: Yeah, I think it's better to use `/`.

piiertho: done.

PR #5, opened 2020-08-30
Feature/get method info

piiertho:
Draft, to discuss implem


src/kt_member.h
```diff
@@ -0,0 +1,13 @@
+#ifndef GODOT_JVM_KT_MEMBER_H
+#define GODOT_JVM_KT_MEMBER_H
+
+
+#include <core/string_name.h>
+
+class KtMember {
```
raniejade, line 7: Do we need this? I.e will there be a time where we are going to return a `KtMember`?

src/kt_method.h
```diff
@@ -0,0 +1,25 @@
+#ifndef GODOT_JVM_KT_METHOD_H
+#define GODOT_JVM_KT_METHOD_H
+
+
+#include <core/object.h>
+#include "kt_member.h"
+
+class KtMethod : KtMember {
```
raniejade, line 8: This should be named `KtFunction` to match the jvm side. Additionally this should wrap the jvm `KtFunction` so it should extend `JavaInstanceWrapper`. Also the info needed should be fetched in the constructor, i.e method name, number parameters etc..

src/kt_class.cpp
```diff
@@ -1,11 +1,20 @@
 #include "kt_class.h"
 
-KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) : JavaInstanceWrapper("godot.core.KtClass", p_wrapped, p_class_loader) {
+KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) :
+    JavaInstanceWrapper("godot.core.KtClass", p_wrapped, p_class_loader), method_fetched(false) {
```
raniejade, line 4: The list of KtFunctions should be fetched from jvm here.

src/kt_class.cpp
```diff
@@ -24,3 +33,17 @@ StringName KtClass::get_super_class(jni::Env& env) {
     jni::JObject ret = wrapped.call_object_method(env, getter);
     return StringName(env.from_jstring(jni::JString((jstring) ret.obj)));
 }
+
+KtMethod* KtClass::get_method(const StringName& methodName) {
```
raniejade, line 37: I prefer return a reference here, `KtClass` still owns the returned `KtMethod` - passing a pointer gives the caller an option to call delete.

src/kt_class.cpp
```diff
@@ -24,3 +33,17 @@ StringName KtClass::get_super_class(jni::Env& env) {
...
+    if (!method_fetched) {
```
raniejade, line 38: do it in the constructor

src/kotlin_script.h
```diff
@@ -8,8 +8,8 @@ class KotlinScript : public Script {
 
 private:
     String source;
+    KtClass *ktClass;
```
raniejade, line 11: KtClass shouldn't be cached here, it should be always fetched from `GDKotlin` which takes care of reloading.

raniejade: Also can we make `GDKotlin` return a `KtClass&` instead of a pointer?

src/kotlin_script.cpp
```diff
@@ -87,9 +93,20 @@ bool KotlinScript::get_property_default_value(const StringName& p_property, Vari
...
 }
 
 void KotlinScript::get_script_property_list(List<PropertyInfo>* p_list) const {
 
 }
+
+void KotlinScript::set_path(const String& p_path, bool p_take_over) {
+    //TODO: Manage path change.
+    Resource::set_path(p_path, p_take_over);
+    ktClass = GDKotlin::get_instance().find_class(get_path());
```
raniejade, line 111: as I mentioned, please don't cache `KtClass` here

src/kt_class.h
```diff
@@ -5,18 +5,28 @@
...
+    ~KtClass();
 
     KtObject* create_instance(jni::Env& env, const Variant** p_args, int p_arg_count, Object* p_owner);
 
+    KtMethod* get_method(const StringName& methodName);
+
+    const Vector<KtMethod*> get_method_list() const;
+
 private:
+    Map<StringName, KtMethod*> methods;
```
raniejade, line 25: can't we just have a map?

src/kt_resource_format_loader.cpp
```diff
@@ -30,7 +30,7 @@ Error read_all_file_utf8(const String &p_path, String &r_content) {
 RES KtResourceFormatLoader::load(const String& p_path, const String& p_original_path, Error* r_error) {
     KotlinScript* script = memnew(KotlinScript);
     Ref<KotlinScript> ref(script);
-    ref->set_path(p_original_path);
+    ref->set_path(p_original_path, false);
```
raniejade, line 33: what's the false for?

raniejade (changes requested):
My main concern is the caching of KtClass, please always delegate to GDKotlin

src/kt_resource_format_loader.cpp
piiertho: Was because over overrising method, but do not longer need it as I will not store ktClass anymore.

src/kotlin_script.h
CedNaru: It makes senses to me to have the KtClass inside KotlinScript. That way each script is responsible for its JVM class.

I don't like the idea of storing them in a singleton. If you want a central place to reload everything, it would make more sense to be to store the KotlinScripts themselves in the GDKotlin.

harness/tests/src/main/kotlin/godot/tests/Entry.kt
```diff
@@ -1,21 +1,121 @@
...
-                function(KtFunction1("string", Invocation::string, ::KtVariant, KtVariant::asString))
-            }
-        }
-    }
-}
\ No newline at end of file
+	override fun Context.init() {
+		with(registry) {
+			registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial") {
+				constructor(KtConstructor0(::Invocation))
```
raniejade, line 9: spaces instead of tabs :P Most likely IJ is picking up the `.editorconfig` of `godot`. You can turn that of in the settings

harness/tests/src/main/kotlin/godot/tests/Entry.kt
```diff
@@ -1,21 +1,121 @@
...
+				function(KtFunction1(KtFunctionInfo("int",
```
raniejade, line 10: for later, might be good to build a DSL here.

kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
```diff
@@ -2,19 +2,85 @@ package godot.core
 
 import godot.util.camelToSnakeCase
 
+enum class PropertyHint {
```
raniejade, line 5: I really hate this enum :P Can we drop the `PROPERTY_HINT` prefix?

kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
```diff
@@ -23,18 +89,18 @@ class KtFunction0<T: KtObject, R>(
...
 
 class KtFunction1<T: KtObject, P0, R>(
-    name: String,
-    private val function: (T, P0) -> R,
-    private val returnValueConverter: (R) -> KtVariant,
-    private val p0Converter: (KtVariant) -> P0
-) : KtFunction<T, R>(name, 1) {
+		functionInfo: KtFunctionInfo,
+		private val function: (T, P0) -> R,
+		val returnValueConverter: (R) -> KtVariant,
```
raniejade, line 94: why is this public?

src/java_instance_wrapper.cpp
```diff
@@ -4,7 +4,7 @@
 Map<String, JavaInstanceWrapper::ClassCache> JavaInstanceWrapper::CLASS_CACHE = Map<String, ClassCache>();
 
 JavaInstanceWrapper::JavaInstanceWrapper(const char* p_class_name, jni::JObject p_wrapped,
-                                         jni::JObject p_class_loader) : class_name(p_class_name) {
+                                         jni::JObject p_class_loader) : classname(p_class_name) {
```
raniejade, line 7: why change the field name? It doesn't follow the convention.

src/kt_class.h
```diff
@@ -5,20 +5,30 @@
...
     StringName super_class;
 
     KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader);
-    ~KtClass() = default;
+    ~KtClass();
 
     KtObject* create_instance(jni::Env& env, const Variant** p_args, int p_arg_count, Object* p_owner);
+    KtFunction* get_method(const StringName& methodName);
+    KtClass* get_parent_class() const;
+    KtClass* get_native_base() const;
```
raniejade, line 21: Please drop this for now, I assume you implemented it for the methods in `KotlinScript` but those methods don't map anything in the JVM side.

src/kt_function.cpp
```diff
@@ -0,0 +1,93 @@
...
+    jni::MethodId getClassNameMethod { get_method_id(env, "getClassName", "()Ljava/lang/String;") };
+    class_name = env.from_jstring(wrapped.call_object_method(env, getClassNameMethod));
+    jni::MethodId getPropertyHintMethod { get_method_id(env, "getHint", "()I") };
+    hint = static_cast<PropertyHint>(wrapped.call_int_method(env, getPropertyHintMethod));
+    jni::MethodId getHintStringMethod { get_method_id(env, "getHintString", "()Ljava/lang/String;") };
+    hint_string = env.from_jstring(wrapped.call_object_method(env, getHintStringMethod));
+}
+
+PropertyInfo KtPropertyInfo::toPropertyInfo() {
+    PropertyInfo info = PropertyInfo();
```
raniejade, line 48: `PropertyInfo info;` would suffice no?

raniejade (changes requested):

src/kotlin_script.h
raniejade: Having `KotlinScript` own the `KtClass` instance just complicates how we need to do reloading which is a very important feature. We have to keep track of all `KotlinScript` instances as well, additionally there is no guarantee that Godot will only create a single `KotlinScript` instance for a specific script file. For example user can call `load("res://src/main/kotlin/MyScript.kt") with caching disabled which means you can have two `KotlinScript` instance for `src/main/kotlin/MyScript.kt` which is only represented by a single `KtClass` instance.

src/java_instance_wrapper.cpp
piiertho: it is because `KtPropertyInfo` has a variable class_name. As it inherits from JavaInstanceWrapper, they should have differents name. I forgot to set a better one. Also I don't have strong opinion about changing name of `JavaInstanceWrapper` member or `KtPropertyInfo`.

kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
piiertho: Because I forgot to set it internal :P

kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
piiertho: I don't have strong opinion about it, so yes.

harness/tests/src/main/kotlin/godot/tests/Entry.kt
piiertho: Did not catch this, thx

src/kt_function.cpp
piiertho: oh yep

src/java_instance_wrapper.cpp
raniejade: Ahhh, didn't realize that. Maybe you can rename the one in `JavaInstanceWrapper` to `java_class_name`?

piiertho: yep that's sound better

src/kotlin_script.h
CedNaru: Yeah I guess we can add a preprocessor to check if we are in editor or not. That's way we can change the behavior from fetching to caching easily but it's not necessary for now.

src/gd_kotlin.cpp
```diff
@@ -67,7 +67,7 @@ void GDKotlin::init() {
     jni::Jvm::init(args);
     print_line("Starting JVM ...");
     auto project_settings = ProjectSettings::get_singleton();
-    scripts_root = project_settings->globalize_path("res://src/main/kotlin");
+    scripts_root = "res://src/main/kotlin/";
```
raniejade, line 70: I don't mind this change, but there are places that `scripts_root` is used that expects a global path.

raniejade (approved):

PR #6, opened 2020-09-05
Feature/entry file dsl

piiertho:
This adds entry file DSL to register methods and properties.
This also create `KtProperty` class, which for now only wrap `KtPropertyInfo` class, but should be used to query properties in future.


piiertho:
I merge this because I need it. If there is any issue, please comment so that I can fix it.

PR #8, opened 2020-09-08
Enable call methods from engine to JVM

piiertho:
This enables calls from godot engine to JVM code.
Please review it carefully, I think we can optimize things, like the way to check if class has method.
It adds an `_enter_tree` method to `Invocation.kt`, printing when node is entering tree.


src/gd_kotlin.cpp
```diff
@@ -131,6 +131,11 @@ KtClass* GDKotlin::find_class(const String& p_script_path) {
     return classes[class_name];
 }
 
+KtClass* GDKotlin::find_class_by_class_name(const String& class_name) {
```
raniejade, line 134: nit: `find_class_by_name` should be enough :)

src/kotlin_instance.cpp
```diff
@@ -47,7 +48,7 @@ void KotlinInstance::get_method_list(List<MethodInfo>* p_list) const {
 }
 
 bool KotlinInstance::has_method(const StringName& p_method) const {
-    return false;
+    return GDKotlin::get_instance().find_class_by_class_name(wrappedObject->get_class_name())->get_method(p_method) != nullptr;
```
raniejade, line 51: maybe `wrappedObject->get_class()` is better? It can call `find_class_by_class_name` under the hood.

raniejade (commented):
A few comments, but this looks good! Can you implement the GUT tests I added in the prototype as well? That should give us a good indication that serializing/deserializing works.

src/kotlin_instance.cpp
piiertho: I can't, this method would return a KtClass, and kt_class import kt_object, so this would introduce a cyclic dependency.

raniejade (approved):
:tada: Would be great to have GUT tests after this :)

PR #11, opened 2020-09-11
Feature/godot unit test

piiertho:
This adds GUT tests on jvm calls.
This fixes method argument passing and return value conversion.


harness/tests/src/main/kotlin/godot/tests/Entry.kt
```diff
@@ -73,7 +73,7 @@ class Entry : godot.runtime.Entry() {
                             className = "Double"
                         }
                 )
-                function("boolean", Invocation::boolean, ::KtVariant, KtVariant::asBoolean,
+                function("boolean_value", Invocation::booleanValue, ::KtVariant, KtVariant::asBoolean,
```
raniejade, line 76: does this mean that the entry generator will transform the name to snake case?

raniejade (approved):
One question, but LGTM!

harness/tests/src/main/kotlin/godot/tests/Entry.kt
chippmann: We did that in the K/N binding. I think we need it here as well? Because one reason was the correct calling of overridden functions IIRC. And this "problem" we also have as a module i suppose.

piiertho: Yep we have to. But we can maybe use method name to create the right snake case name when registering.

PR #12, opened 2020-09-13
Feature/get set properties

piiertho:
This implement get and set property calls.

### DIFF
--- a/harness/tests/src/main/kotlin/godot/tests/Entry.kt
+++ b/harness/tests/src/main/kotlin/godot/tests/Entry.kt
@@ -8,22 +8,22 @@ class Entry : godot.runtime.Entry() {
             registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial") {
                 constructor(KtConstructor0(::Invocation))
 
-                property {
+                property(Invocation::x::get, Invocation::x::set, ::KtVariant, KtVariant::asInt) {
                     name = "x"
                     type = KtVariant.Type.LONG
                     className = "Int"
                 }
-                property {
+                property(Invocation::y::get, Invocation::y::set, ::KtVariant, KtVariant::asDouble) {
                     name = "y"
                     type = KtVariant.Type.DOUBLE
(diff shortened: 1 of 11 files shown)


harness/tests/src/main/kotlin/godot/tests/Entry.kt
```diff
@@ -8,22 +8,22 @@ class Entry : godot.runtime.Entry() {
             registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial") {
                 constructor(KtConstructor0(::Invocation))
 
-                property {
+                property(Invocation::x::get, Invocation::x::set, ::KtVariant, KtVariant::asInt) {
```
raniejade, line 11: You can pass in a `KMutableProperty` IIRC, so you only need to do `Invocation::x` instead of passing the getter and setter.

harness/tests/test/unit/test_calls.gd
```diff
@@ -11,6 +11,49 @@ func test_godot_to_jvm_call_using_new() -> void:
 	assert_eq(invocation_script.int_addition(2, 7), 9, "call to addition jvm method should return 2 + 7")
 	invocation_script.free()
 
+func test_get_int_call_using_scene() -> void:
+	var invocation_scene = load("res://Spatial.tscn").instance()
```
raniejade, line 15: maybe use `_new()` instead of loading the scene?

raniejade (changes requested):

harness/tests/test/unit/test_calls.gd
piiertho: I have two test, one with scene, the other with new :P

raniejade: I think the scene one is redundant, it's not really testing our code.

piiertho: yep that's true, new already test instance_create

raniejade (approved):

PR #13, opened 2020-09-13
Enhancement/use property and function name to register

piiertho:
This use property and function name as default registration name.

### DIFF
--- a/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
@@ -50,14 +50,13 @@ class ClassBuilderDsl<T : KtObject>(
     fun <R> function(
             func: KFunction1<T, R>,
             returnValueConverter: (R) -> KtVariant,
-            funcName: String = func.name.camelToSnakeCase(),
             returns: KtPropertyInfoBuilderDsl.() -> Unit
     ) {
         val returnBuilder = KtPropertyInfoBuilderDsl()
         returnBuilder.returns()
         appendFunction(
                 KtFunction0(
-                        KtFunctionInfo(funcName, listOf(), returnBuilder.build()),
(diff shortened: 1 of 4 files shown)


kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -64,12 +64,14 @@ class ClassBuilderDsl<T : KtObject>(
...
-                          func: (T, P0) -> R,
-                          returnValueConverter: (R) -> KtVariant,
-                          p0Converter: (KtVariant) -> P0,
-                          arg: KtPropertyInfoBuilderDsl.() -> Unit,
-                          returns: KtPropertyInfoBuilderDsl.() -> Unit) {
+    fun <P0, R> function(
+            func: KFunction2<T, P0, R>,
+            returnValueConverter: (R) -> KtVariant,
+            p0Converter: (KtVariant) -> P0,
+            funcName: String = func.name.camelToSnakeCase(),
```
raniejade, line 71: do we need it as a parameter? I don't we will ever need to pass in a different name

raniejade (changes requested):

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: I wonder if someone wants to register it with a different name, if its method name conflicts with another from godot.
In this case I would say user have to change method name. But as I don't know all use cases, I just let the possibility to override it.

raniejade: They have to change the name of their method then :P I don't that we should be the one to adjust for that.

piiertho: I don't have strong opinion regarding this

raniejade (approved):

PR #14, opened 2020-09-15
Feature/handle icalls

piiertho:
This enables to call godot methods from JVM.

### DIFF
--- a/src/transfer_context.h
+++ b/src/transfer_context.h
@@ -20,14 +20,18 @@ public:
 
     void write_args(jni::Env& p_env, const Vector<KtVariant>& p_args);
     Vector<KtVariant> read_args(jni::Env& p_env, bool p_refresh_buffer);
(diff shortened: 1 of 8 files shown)


src/transfer_context.h
```diff
@@ -21,11 +23,20 @@ class TransferContext : public JavaInstanceWrapper {
...
+               jstring jClassName, jstring jMethod,
+               jint expectedReturnType, bool p_refresh_buffer);
+
 private:
     SharedBuffer* get_buffer(jni::Env& p_env, bool p_refresh_buffer);
     bool ensure_capacity(jni::Env& p_env, long p_capacity);
     SharedBuffer* shared_buffer;
-};
 
+    static Variant variantArgs[MAX_ARGS_SIZE];
```
raniejade, line 37: do we need to declare them here? Can it be an implementation detail of the cpp file?

PR #17, opened 2020-09-19
hotfix/wrong invoke constructor ptrconversion

piiertho:
This fixes crash when trying to use the ptr created from invoke_constructor stored in JVM instance.


raniejade (commented):
in java represent the ptr as a long, how will this affect it?

piiertho:
> in java represent the ptr as a long, how will this affect it?
It is still a long, but depending on platform it can be signed or unsigned

raniejade (approved):

PR #19, opened 2020-10-01
Feature/register engine types

piiertho:
This enable to create right type when getting object from C to Kotlin.


kt/godot-runtime/src/main/kotlin/godot/core/TransferContext.kt
```diff
@@ -21,7 +23,7 @@ object TransferContext {
 
     private var inputStream  = object : InputStream() {
         override fun read(): Int {
-            return buffer.get().toInt()
+            return buffer.get().toInt() and 0xFF
```
raniejade, line 26: hmmm, do we still need the `toInt`?

piiertho: Yep we still need it I think.
The problem was that `toInt` returns an int with same significant 8 first bit, so I think mask should be applied on them after int conversion.

raniejade: but we lose the MSB for the byte right?

piiertho: Maybe I'm blind on this problem but think this should not.
If I take -2 signed byte as an example:
Its unsigned equivalent is 254, which is :
1111 1110
When converting it to int, it will create -2 int which is : 
1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1110
When applying 255 mask, we only keep 1111 1110, which result in 254.
As values goes from -128 to 127 this should not cause problem. 

But anyway, maybe we should use UByte class as you suggested on discord, as it would clearly express intention.

harness/tests/addons/gut/plugin_control.gd
```diff
@@ -125,26 +125,26 @@ var _cancel_import = false
 var _placeholder = null
 
 func _init():
-    # This min size has to be what the min size of the GutScene's min size is
-    # but it has to be set here and not inferred i think.
-    rect_min_size = Vector2(740, 250)
+	# This min size has to be what the min size of the GutScene's min size is
```
chippmann, line 128: I think we should define what indentation we use for `gd` files. If i saw that correctly @raniejade uses spaces and @piiertho  and I use tabs.
What should we use?

harness/tests/src/main/kotlin/godot/tests/Entry.kt
```diff
@@ -129,7 +130,25 @@ class Entry : godot.runtime.Entry() {
                             className = "Unit"
                         }
                 )
+                function(Invocation::parentIsSpatial, ::KtVariant,
```
chippmann, line 133: Note to myself: Whoever merges last, needs to change this to annotation registration

kt/godot-runtime/src/main/kotlin/godot/core/TransferContext.kt
```diff
@@ -19,9 +19,10 @@ object TransferContext {
         }
     }
 
-    private var inputStream  = object : InputStream() {
+    @ExperimentalUnsignedTypes
```
chippmann, line 22: Shouldn't we just add this as a compiler arg instead of annotations everywhere?
```kotlin
tasks.withType<KotlinCompile>().all {
    kotlinOptions.freeCompilerArgs += "-Xopt-in=kotlin.ExperimentalUnsignedTypes"
}
```

harness/tests/addons/gut/plugin_control.gd
piiertho: I don't have strong opinion about this.

src/bootstrap.h
```diff
@@ -7,11 +7,13 @@ class Bootstrap : public JavaInstanceWrapper {
...
     typedef void (*LoadClassesHook)(JNIEnv* p_env, jobject p_this, jobjectArray classes);
     typedef void (*UnloadClassesHook)(JNIEnv* p_env, jobject p_this, jobjectArray classes);
+    typedef void (*RegisterManagedEngineTypes)(JNIEnv* p_env, jobject p_this, jobjectArray classes_names);
 
     Bootstrap(jni::JObject p_wrapped, jni::JObject p_class_loader);
     ~Bootstrap() = default;
 
-    void register_hooks(jni::Env& p_env, LoadClassesHook p_load_classes_hook, UnloadClassesHook p_unload_classes_hook);
+    void register_hooks(jni::Env& p_env, LoadClassesHook p_load_classes_hook, UnloadClassesHook p_unload_classes_hook,
+                        RegisterManagedEngineTypes p_register_managed_engine_types_hook);
```
chippmann, line 16: Is it not common for cpp to format functions like this similar to kotlin?:
```cpp
void register_hooks(
    jni::Env& p_env, LoadClassesHook p_load_classes_hook,
    UnloadClassesHook p_unload_classes_hook,
    RegisterManagedEngineTypes p_register_managed_engine_types_hook
);
```

chippmann (approved):
Some questions that are not affecting the PR.
Very nice work though. The entry generation for it should also work already. But we'll see that once this is merged.

harness/tests/addons/gut/plugin_control.gd
chippmann: Me neither. I guess let's stick to the default of Godot which is tabs IIRC. WDYT @raniejade ?

src/bootstrap.h
piiertho: I don't think there is "common practices" in cpp. I saw projects with methods and properties in snake case (like godot) and projects in camel case.

chippmann: Ah i don't mean the casing
I mean the formatting in terms of line breaks

kt/godot-runtime/src/main/kotlin/godot/core/TransferContext.kt
raniejade: I'm a way to late here, but here are my thoughts: This would enable the usage across the codebase, but ideally for experimental types we need to use them sparingly. Having to add annotation everytime is a good source of friction: `Do I really need this experimental feature?`

PR #21, opened 2020-10-12
Implement EntryGeneration

chippmann:
This adds https://github.com/utopia-rise/godot-kotlin-entry-generator as a submodule and implement entry generation for everything the registration dsl supports for now.

Draft PR as it should only be merged after https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/1 is merged and the submodule is updated in this PR.


kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/GodotPlugin.kt
```diff
@@ -36,10 +45,29 @@ class GodotPlugin : Plugin<Project> {
...
+
+            val mainCompilation = jvm.target.compilations.getByName("main").apply {
+                configureSourceSets(true)
+            }
+
+            val dummy = jvm.target.compilations.create("dummy") {
+                defaultSourceSet.dependencies {
+                    main.allDependencies.forEach {
+                        implementation(it)
+                    }
```
chippmann, line 68: Here I'm not so sure if this really covers all cases and does not imply some problems.
I tried it by adding an external dependency (see comments at the corresponding places) and it worked.
Maybe @piiertho or @raniejade have a better idea or can assure me that this will work?

piiertho (approved):

PR #22, opened 2020-10-17
Feature/signal registration

piiertho:
This implements signals and their registration.
Get back signals code from kotlin-native implem (maybe need to be centralized in a common artifact ?).
This fix mapping from JVM variant type to Wire type.
This adds `GodotError` class and fake `VariantArray`.
This does not implements signal binds, this needs real `VariantArray` implem.


chippmann (approved):
LGTM
I'm not so sure though if it makes sense to share the signal registration code. It's not much code and I think it might change a bit in the future and will diverge from the one we have in the kotlin native project

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
```diff
@@ -6,59 +6,112 @@ import godot.Spatial
...
+
+	override fun _onDestroy() {
+		invocation.free()
+	}
+
+	//Type cast checks
+	@RegisterFunction
+	fun parentIsSpatial() = getParent() is Spatial
+	@RegisterFunction
+	fun isObjectSpatial(obj: Object) = obj is Spatial
```
raniejade, line 116: tabs :P

kt/godot-library/src/main/kotlin/godot/Wrapping.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Used for variadic icalls.
+ * TODO: Make non generic emit and connect methods at code generation time to avoid pattern matching when calling signals
+ */
+fun wrap(obj: Any?): KtVariant {
+    if (obj == null) {
+        return KtVariant(Unit)
+    }
+    return when (obj) {
```
raniejade, line 13: I bet tristan is scratching his head when he sees this 😂

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
piiertho: mmmh weird, still got spaces in intellij.

kt/godot-library/src/main/kotlin/godot/Wrapping.kt
piiertho: yep but we don't have much choice regarding variadic godot methods

piiertho: Was not on all file.
Seems godot sets tabs each time I open project.

CedNaru (approved):
Okay for merge.
Note: Might be necessary to rework the string manipulation for the name properties in Signal.

piiertho:
> Okay for merge.
> Note: Might be necessary to rework the string manipulation for the name properties in Signal.

I agree with this. There is maybe a way we can improve this like what we did with constructors. We should open an improvement issue.

PR #23, opened 2020-10-21
Add support for user type in signal arguments

totalorder:
Hi! 

I tried to address the following comment in Invocation.kt:
```
//		This will fail with:
//		class godot.Spatial cannot be cast to class godot.tests.Invocation.
...
//	fun hookTwoParam(str: String, inv: Invocation) {
```

I don't know anything about this project or how things are supposed to work. So this PR might be total garbage. But it was a nice way of trying to get to know the code-base. There is probably a much better solution that I didn't think of. If you have time, please point me in the right direction. If not, feel free to send any other issues my way if there are some small things I can help out with.

Running this code produces the expected output of:
```
Hook was called with parameters: My Awesome param !, godot.Spatial@647e447
Hello Invocation!
Hook was called with parameters (invocation): My Awesome param (invocation)!, godot.tests.Invocation@41fbdac4
```

This PR depends on the following change to godot-kotlin-entry-generator: https://github.com/utopia-rise/godot-kotlin-entry-generator/compare/master...totalorder:argument-full-fqdn?expand=1

### DIFF
--- a/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
@@ -5,7 +5,7 @@ import godot.util.VoidPtr
 class KtClass<T : KtObject>(
         val name: String,
         val superClass: String,
-        private val constructors: Map<Int, KtConstructor<T>>,
+        val constructors: Map<Int, KtConstructor<T>>,
         private val _properties: Map<String, KtProperty<T, *>>,
         private val _functions: Map<String, KtFunction<T, *>>,
         private val _signalInfos: Map<String, KtSignalInfo>
--- a/kt/godot-runtime/src/main/kotlin/godot/core/TypeManager.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/core/TypeManager.kt
@@ -1,19 +1,25 @@
 package godot.core
 
+import godot.util.VoidPtr
+
 object TypeManager {
-    private val userTypes = HashSet<String>()
+    val userTypeNames = LinkedHashSet<String>()
+    val userTypeConstructors = mutableListOf<(VoidPtr) -> KtObject>()
 
     val engineTypeNames = LinkedHashSet<String>()
-    val engineTypesConstructors = mutableListOf<() -> KtObject>()
+    val engineTypeConstructors = mutableListOf<() -> KtObject>()
 
-    fun registerUserType(className: String) = userTypes.add(className)
+    fun <T: KtObject> registerUserType(className: String, invocator: (rawPtr: VoidPtr) -> T) {
+        userTypeConstructors.add(invocator)
+        userTypeNames.add(className)
+    }
 
     fun <T: KtObject> registerEngineType(className: String, invocator: () -> T) {
-        engineTypesConstructors.add(invocator)
+        engineTypeConstructors.add(invocator)
         engineTypeNames.add(className)
     }
 
-    fun isUserType(className: String) = userTypes.contains(className)
+    fun isUserType(className: String) = userTypeNames.contains(className)
 
     fun isEngineType(className: String) = engineTypeNames.contains(className)
 }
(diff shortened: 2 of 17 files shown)


totalorder:
@chippmann @piiertho Please take a look if you have time.

src/kt_function.cpp
```diff
@@ -1,88 +1,95 @@
-#include <utility>
```
totalorder, line 1: I'm not sure what happened to the diff here. Line ending weirdness? I only made changes to `Variant KtFunction::invoke(const KtObject* instance, const Variant** p_args, int args_count)`

src/kt_variant.cpp
```diff
@@ -152,6 +159,20 @@ KtVariant::KtVariant(const Variant& variant) {
     TO_KT_VARIANT_FROM[type](value, variant);
 }
 
+KtVariant::KtVariant(const Variant& variant, const KtPropertyInfo* argInfo) {
+    String class_name {argInfo->class_name};
+
+    if (JAVA_USER_TYPE_CONSTRUCTORS.has(class_name)) {
```
totalorder, line 165: The user types are registered using the full fqdn. That's why [this change in godot-kotlin-entry-generator](https://github.com/utopia-rise/godot-kotlin-entry-generator/compare/master...totalorder:argument-full-fqdn?expand=1) is needed. I don't know if that would break a lot of other stuff somewhere. Please let me know :)

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -657,8 +657,9 @@ class ClassRegistry {
     fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, cb: ClassBuilderDsl<T>.() -> Unit) {
         val builder = ClassBuilderDsl<T>(name, superClass)
         builder.cb()
-        TypeManager.registerUserType(name)
-        registerClass(builder.build())
+        val clazz = builder.build()
+        TypeManager.registerUserType(name) { voidPtr -> clazz.new(voidPtr, 0) }
```
totalorder, line 661: I'm assuming all user types has a zero-arg constructor here. I don't know if that's a reasonable assumption or what to do if it doesn't hold.

chippmann:
@totalorder First of all: thx a lot for your time investment!
Regarding the entry generator changes: Feel free to open a PR in that repo.
I already did that change in [this](https://github.com/utopia-rise/godot-kotlin-entry-generator/tree/feature/default-values) branch for the Property and the Function generators but forgot the Signal generators.
So it' doesn't hurt if you do a PR there and we merge that beforehand and update this PR so it points at the right commit from the entry generator. If we would merge it now it would not function correctly i suppose.

Regarding the rest of the PR: I'll look at that later today.

kt/godot-runtime/src/main/kotlin/godot/core/TypeManager.kt
```diff
@@ -1,19 +1,25 @@
...
+import godot.util.VoidPtr
+
 object TypeManager {
-    private val userTypes = HashSet<String>()
+    val userTypeNames = LinkedHashSet<String>()
+    val userTypeConstructors = mutableListOf<(VoidPtr) -> KtObject>()
 
     val engineTypeNames = LinkedHashSet<String>()
-    val engineTypesConstructors = mutableListOf<() -> KtObject>()
+    val engineTypeConstructors = mutableListOf<() -> KtObject>()
```
raniejade, line 10: Encapsulation is broken here, this should be all private. Why do we need to expose a mutable field to the public?.

kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
```diff
@@ -5,7 +5,7 @@ import godot.util.VoidPtr
 class KtClass<T : KtObject>(
         val name: String,
         val superClass: String,
-        private val constructors: Map<Int, KtConstructor<T>>,
```
raniejade, line 8: this needs to be private.

raniejade (commented):
Thanks for the pr! This will only work if the user type has a no-arg constructor. It's also a bit wastefull, constructing a new object to wrap a user type. I think it will good if we can track instances of user types indexed by their rawPtr.

totalorder (commented):
@raniejade @chippmann: Thank you for the feedback!

I discussed the "track instances of user types indexed by their rawPtr" with @piiertho on Discord. How that is solved has a big impact on how to approach the problem I'm trying to solve in this PR. Because of that I'm closing this PR until a decision has been made on how to approach that.

In the mean time I created a PR out of the generator-changes, please take a look: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/2

Thanks for pointing me in the right direction about this. I know nothing about the godot-jvm code base and this was a good way to figure stuff out along the way.

totalorder:
Closing this until decision on "track instances of user types indexed by their rawPtr" is made.

PR #25, opened 2020-10-25
Feature/register references

piiertho:
This enable use or default value in registration.
This also maps REF to KtObject life cycle, in wait of GC.


piiertho:
Closed in favor of #26

PR #26, opened 2020-11-01
Feature/begin garbage collector

piiertho:
I know this is a too big PR, which does not include GC alone.
It also include some of changes of #25 concerning default value registration.
@chippmann I made a branch on entry generation to tweak it to experiment. Can you clean it ?

This PR include:
- Registration for default value (need a check on entry generation by @chippmann )
- A "GarbageCollector" JVM thread that bind lifecycle of `REF` to JVM GC and check validity of `Object` instances (using ptr and instanceid) to avoid Instancing hell.
- Some engine command arguments to help debug jvm: `--jvm-debug-port` and `--jvm-debug-address` to enable connecting remote debugger. `--jvm-gc-thread-period-millis` to override default GC's frequency and `--jvm-force-gc` to force GC call JVM each time we check instance /!\ Use it only for debugging /!\

This PR also shows we have a lot of jni local ref leak we need to manage.


src/gd_kotlin.cpp
```diff
@@ -75,30 +124,77 @@ void GDKotlin::init() {
...
+    };
+    jni::JObject garbage_collector_instance{
+            garbage_collector_cls.get_static_object_field(env, garbage_collector_instance_field)
+    };
+    CRASH_COND_MSG(garbage_collector_instance.isNull(), "Failed to retrieve GarbageCollector instance")
+    jni::MethodId close_method_id{garbage_collector_cls.get_method_id(env, "close", "()V")};
+    garbage_collector_instance.call_void_method(env, close_method_id);
+    jni::MethodId has_closed_method_id{garbage_collector_cls.get_method_id(env, "isClosed", "()Z")};
+    while (!garbage_collector_instance.call_boolean_method(env, has_closed_method_id)) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(600));
```
piiertho, line 190: Should remove use of stl.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+        }
+        gcState = GCState.CLOSED
+    }
+
+    private fun checkAndClean() {
+        // A native reference cannot die while a jvm instance exists (because counter > 0). When we don't need the
+        // jvm instance anymore, we decrease the counter.
+        for (entry in refWrappedMap) {
+            if (entry.value.get() == null) {
+                if (MemoryBridge.unref(entry.key)) {
```
raniejade, line 95: something for later: we probably can batch the call here to lessen the amount of calls we make to cpp.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+        }
+        for (ptr in refWrappedSuppressBuffer) {
+            refWrappedMap.remove(ptr)
+        }
+        refWrappedSuppressBuffer.clear()
+
+        // Check validity of cpp pointer for classic godot Object, if not valid, then remove jvm instance.
+        // This binds jvm instance lifecycle to native object's one.
+        for (entry in wrappedMap) {
+            if (!MemoryBridge.checkInstance(entry.key, entry.value.godotInstanceId)) {
```
raniejade, line 108: same as above, maybe batch the call.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+    val refWrappedSuppressBuffer = mutableListOf<VoidPtr>()
+    val wrappedSuppressBuffer = mutableListOf<VoidPtr>()
+
+    private var forceJvmGarbageCollector = false
+    private var gcState = GCState.NONE
+    val isClosed: Boolean
+        get() = gcState == GCState.CLOSED
+
+    fun registerInstance(instance: KtObject) {
+        synchronized(this) {
```
raniejade, line 20: Point of contention, do we need to lock every time or only when we start the gc checks?

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -0,0 +1,143 @@
...
+            if (forceJvmGarbageCollector) {
+                forceJvmGc()
+            }
+            checkAndClean()
+            sleep(500)
+        }
+        gcState = GCState.CLOSED
+    }
+
+    private fun checkAndClean() {
```
raniejade, line 90: We probably need to synchronize this with `registerInstance`. What if we move the state of the GC locally, i.e:

```kt
fun checkAndClean() {
	lateinit var refWrappedMapCopy: ...;
    // ...
	synchronize(this) {
         val refWrappedMapCopy = refWrappedMap;
         // do the same as the others

         refWrappedMap = new ...;
    }

    // do gc checks with the copy
}
```

raniejade: ahh, we still need to support `getObjectInstance` while running the gc.

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
```diff
@@ -232,11 +233,19 @@ class KtVariant {
 
     inline fun <reified T : KtObject> asObject(): T {
         val objectValue = data.objectValue
+        val ptr = objectValue.ptr
         val constructorIndex = objectValue.engineConstructorIndex
-        return KtObject.instantiateWith(
-                objectValue.ptr,
+        val isRef = objectValue.isRef
+        return (if (isRef) {
+            GarbageCollector.refWrappedMap[ptr]?.get()
```
raniejade, line 240: we need to synchronize this right? `refWrappedMap` can be modified by the another thread at this point.

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
```diff
@@ -232,11 +233,19 @@ class KtVariant {
...
+        } else {
+            GarbageCollector.getObjectInstance(ptr)
```
raniejade, line 242: same here I guess.

raniejade (commented):
This really looks amazing, good work guys! I just have a few concerns regarding thread safety.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
piiertho: Yep, that was one of our consideration with @CedNaru

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
CedNaru: Well we have to lock the container each time we want to read/write. Otherwise we take the risk of deleting/adding stuff at the same time.
But I have plans to add queue for adding and removing stuff. That way we will only lock on the queue instead of the whole class.

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
CedNaru: Hum correct.  This current implementation is dangerous.

piiertho: Yep I agree. Maybe we should go for queue now @CedNaru

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
CedNaru: Making a copy of the container seems heavy for real life applications.
We should probably prevent direct access to the maps and create methods for external access.
Code outside of the GC only need to read and we only be cautious about thread safety when writing, the only place where it's the case is when registering/removing instead bur I'm going to add queues so they are done in the GC thread.
We don't need to lock when we do the checks themselves because in that location we know we are in the GC thread so writing is not possible at that moment (because only the GC thread will write).

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -0,0 +1,154 @@
...
+
+    }
+
+    override fun run() {
+        while (gcState == GCState.STARTED) {
+            if (forceJvmGarbageCollector) {
+                forceJvmGc()
+            }
+            checkAndClean()
+            sleep(500)
```
piiertho, line 92: We should get rid of this, I think this can really impact performances.

CedNaru (approved):
We got a lot of potential optimizations possible for a future update.
But for now this first GC version seems to be able to handle every kind of situation so let's merge that.

chippmann (approved):
Very nice work!
Had to add `#include <thread>` in `gd_kotlin.cpp` in order to be able to compile it on linux but as we want to remove stl usage anyways it's ok for me.
Before merge though, I would merge https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/3 and update this PR.

raniejade (approved):

PR #27, opened 2020-11-07
Optimization/rework java instance wrapper

piiertho:
This implements CRTP to get rid of `ClassCache` class using CRTP to get a lazy static `jclass` for each `JavaInstanceWrapper` inheritors. Also implements some macro to declare jni methods statically for each jni wrapper class, with lazy init.
This also get rids of `StringBuilder` for each jni call from cpp.

This also get rids of string use for icalls using registration time to map ids to Strings.

This PR improves icall benchmark between 20% and 25%.

https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/4 needs to be merged in the same time.


CedNaru (approved):

raniejade (approved):

harness/benchmarks/src/main/kotlin/godot/benchmark/Simple.kt
```diff
@@ -47,4 +53,21 @@ class Simple : Object() {
         return s
     }
 
+    @RegisterFunction
+    fun benchmarkICall(): Node {
+        val node = Node()
+        node.getInstanceId()
+        return node
+    }
+
```
chippmann, line 62: Annotation missing. But not really important as the entry generation is not yet added to the benchmarks.
Maybe we should add this next?

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -11,24 +11,24 @@ open class Object(isRef: Boolean = false) : KtObject(isRef) {
     constructor() : this(false)
 
     override fun __new(): VoidPtr {
-        return TransferContext.invokeConstructor("Object")
+        return TransferContext.invokeConstructor(0)
```
chippmann, line 14: Couldn't we use `const val`'s for all these constant integers? would improove readability in my opinion and  the compiler should optimize these constants anyways if they can't change.
even though we'll generate these classes later on, once one needs to debug them, having variables in the source code rather than plain numbers might be beneficial?

chippmann (approved):
One comment. But i don't have a strong opinion on this one. So if you don't agree you can resolve and merge.

harness/benchmarks/src/main/kotlin/godot/benchmark/Simple.kt
piiertho: Yep, purpose here, was not to convert benchmarks with annotations, only optimizing.

kt/godot-library/src/main/kotlin/godot/tests.kt
piiertho: I agree, let me change this, so that I remember for later entry generation.

PR #28, opened 2020-11-11
optimization/ remove list conversions in jni

piiertho:
This adds helper function to create array and get rid of static_initializer and conversions.


chippmann (approved):
Nice work!
Tested on linux.

CedNaru (approved):

raniejade (approved):

PR #29, opened 2020-11-18
Optimization/get rid protobuf

piiertho:
~~This PR is not ready. It needs more testing and adapted entry generation.~~
This gets rid of protobuf to use classic jvm buffer, shared with cpp side.
On cpp side we use godot's serialisation api, on jvm side, classic bytebuffer api.
This remove KtVariant class to get rid of instancing on jvm side using static functions and any conversions.

This multiply performances on icall benchmark by 3 on my side.

Edit from @chippmann:
~~Needs to be merged after https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/5 is merged and this PR is updated.~~ Done...


kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
```diff
@@ -1,337 +1,285 @@
...
+    ),
+    STRING(
+            { buffer: ByteBuffer ->
+                val stringSize = buffer.int
+                val charArray = kotlin.ByteArray(stringSize)
+                buffer.get(charArray, 0, stringSize)
+                kotlin.text.String(charArray, kotlin.text.Charsets.UTF_8)
+            },
+            { buffer: ByteBuffer, any: Any ->
+                any as String
```
raniejade, line 102: this is a bit weird, maybe use `require`?

raniejade (commented):
From what I understand the buffer size is fixed? What will happen when you pass in a very long string?

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
piiertho: Here, the check is not necessary, by design, we know the type we have as it is linked to the enum.

piiertho:
> From what I understand the buffer size is fixed? What will happen when you pass in a very long string?

I think a big buffer allocated one time would be better than a reallocating on the fly. I think what should be discussed is what buffer size we should have, so that we have a "big" memory space reserved at thread startup. 
I think very large string like npc text should go in localization resource file. 
WDYT ?

chippmann:
> > From what I understand the buffer size is fixed? What will happen when you pass in a very long string?
> 
> I think a big buffer allocated one time would be better than a reallocating on the fly. I think what should be discussed is what buffer size we should have, so that we have a "big" memory space reserved at thread startup.
> I think very large string like npc text should go in localization resource file.
> WDYT ?

Generally i agree. Constant size checking and reallocation hurts perf.
Still i see the point of ranie. While i agree that something like a NPC dialog should go in a translation file, we cannot be sure what a user does with strings. For example dynamic dialog's which are assembled based on user actions.

Could we not have a big fixed size that should cover most needs, and extend it dynamically if it's not enough and help the compiler with `unlikely` ?
In any case i would make that initial buffer size configurable. Maybe through a cmdline argument?

piiertho:
> > > From what I understand the buffer size is fixed? What will happen when you pass in a very long string?
> > 
> > 
> > I think a big buffer allocated one time would be better than a reallocating on the fly. I think what should be discussed is what buffer size we should have, so that we have a "big" memory space reserved at thread startup.
> > I think very large string like npc text should go in localization resource file.
> > WDYT ?
> 
> Generally i agree. Constant size checking and reallocation hurts perf.
> Still i see the point of ranie. While i agree that something like a NPC dialog should go in a translation file, we cannot be sure what a user does with strings. For example dynamic dialog's which are assembled based on user actions.
> 
> Could we not have a big fixed size that should cover most needs, and extend it dynamically if it's not enough and help the compiler with `unlikely` ?
> In any case i would make that initial buffer size configurable. Maybe through a cmdline argument?

The problem is that this buffer is allocated on jvm side. That's another reason I want it to be fix. I think we should ask which size is enough. Or maybe we should define max buffer size as a parameter, so that user can increase it, but he would be aware that what he is doing is not recommended for performances and design purpose.

chippmann:
Ah yeah. Didn't pay attention.
The size currently is 2 KB right?
Imagine one downloads a big json through the godot api's and sends it to kotlin-jvm for serialisation and further processing. There we already have a problem right? Of course he could save the json string to disk and send the path to the jvm but that hurts perf more than dynamically increasing the buffer size.

So what is enough really depends on the usage and I'm not really comfortable defining whats enough for most use cases.
As a middle way we could (as said) make the size configurable, but also if the size should be extended if needed. Of course write (maybe even through a log warning) that this is not recommended.
As we are on the jvm, jit should be able to then do the same as `unlikely`.

But another problem arises if we make it dynamic (as option or not): When do we reduce the size again? We can't just let it grow and stay big.

raniejade:
An alternative is we some kind of registry for strings, it can be some buffer that we manage which does some caching for commonly used strings. Then the main buffer can just be just an "index" into that registry, we probably need to add a length component as well.

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
raniejade: Yep, it is just an unusual way to do a smart cast, may be do 

```kotlin
val stringBytes = (any as String).encodeToByteArray()
```

raniejade: or if you must really do that - then I suggest doing a `require(any as String)` - it's not really about the check but more of the readability.

chippmann:
> An alternative is we have some kind of registry for strings, it can be some buffer that we manage which does some caching for commonly used strings. Then the main buffer can just be just an "index" into that registry, we probably need to add a length component so we know when to stop reading.

Hmm i think that is overkill and does not really help with performance. The main reason not to make the buffer size dynamic is to gain perf. I don't think managing a cache which in turn has a dynamics buffer will make things better. More i think the perf is unpredictable if you don't know the inner workings of this module. Furthermore it means something special to maintain just for strings.

kt/godot-runtime/src/main/kotlin/godot/core/Variants.kt
piiertho: If it is only readability purpose, why not only renaming any parameter ? I mean, yes it is unusual, but it is the most optimized way, and what we do here is somewhat unusual I think ^^

raniejade:
True, I guess we need a different way to pass in big strings - for small strings we can just use the buffer. One option is to use plain old java strings for big strings. So a string in the buffer can either be direct (in the buffer) or a java string (maybe an index to some data structure?)

piiertho:
> > An alternative is we have some kind of registry for strings, it can be some buffer that we manage which does some caching for commonly used strings. Then the main buffer can just be just an "index" into that registry, we probably need to add a length component so we know when to stop reading.
> 
> Hmm i think that is overkill and does not really help with performance. The main reason not to make the buffer size dynamic is to gain perf. I don't think managing a cache which in turn has a dynamics buffer will make things better. More i think the perf is unpredictable if you don't know the inner workings of this module. Furthermore it means something special to maintain just for strings.

I agree with cedric.
Maybe @CedNaru have an idea about how we shoule manage the buffer. I wonder what impact would have to only set a big size to the buffer.

piiertho:
> True, I guess we need a different way to pass in big strings - for small strings we can just use the buffer. One option is to use plain old java strings for big strings. So a string in the buffer can either be direct (in the buffer) or a java string (maybe an index to some data structure?)

Do you know how jvm manage big parameters ? I wonder if there's a max length and if not how they do.

raniejade:
> Do you know how jvm manage big parameters ? I wonder if there's a max length and if not how they do.
We just pass in a reference, the string itself is allocated in the heap. 

There are two scenarios that I can think of:

(1) Big string coming from Godot, so CPP -> JVM:
- we can allocate a jstring in the cpp side
- store that string in a jarray
- in the buffer, we encode the index of the string in the jarray.
- we call jvm including the jarray (note we only do this if we have big strings)

(2) Big string coming from JVM -> CPP:
- put string in jarray
- encode in buffer the index of the string in the jarray
- call cpp including the jarray (again we only do this if we have to pass in a big string)

For both cases we need to differentiate in the buffer if the string is "direct" (can be read from the buffer directly) or "indirect", we need to look it up in the string jarray. We just need to define what a big string means in terms of length. The number should ideally allow us lessen how many times we need to use indirection.

raniejade:
or instead of java strings, we can use plain buffers (void*) - we can read them in the java side using Unsafe. We can also allocate offheap memory using unsafe, so the reverse is true as well.

chippmann:
Hmm somehow how Variant wraps pointers rather than the object itself if the type is Object?
This doesn't sound too bad IMO.

still we would not have a fixed perf with this approach like we have atm. But if we document that additional cost of handling big strings, it sounds good to me.

But unsafe sounds even better. There the perf would also be fixed and generally better than with jstring i assume

piiertho:
IIRC unsafe is not accessible on android

chippmann:
Android should not be the limiter as godot launches a dedicated traditional jvm IIRC. And our plan was to use that right? On the dalvik vm your right, there we don't have access to unsafe

chippmann:
Discussion continued on discord...
Outcome: TBD

piiertho:
So to sum up discord's discussions.
We have ways to be able to signify that more than one buffer use is needed, but it still need some investigations.
This issue should be solved in a seperate PR.

For now I added the possibility to change shared buffer size.
I think this PR is ready for review.

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -89,18 +89,19 @@ class ClassBuilderDsl<T : KtObject>(
...
+                        propertyName,
+                        "Int",
+                        PropertyHint.ENUM,
+                        enumValues<P>().joinToString { it.name }
+                ),
+                kProperty,
+                { enum -> getVariantType(enum.ordinal) },
+                { any -> enumValues<P>()[(any as Long).toInt()] },
+                //TODO : Not sure here
+                enumValues<P>()[0],
```
chippmann, line 103: Still relevant?
I would use `null` as no value was set. Using the first enum value could lead to undesired behaviour no?
Also it would match with what we had with `KtVariant(Unit)`

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -172,16 +173,18 @@ class ClassBuilderDsl<T : KtObject>(
 
                 enums
             },
-            KtVariant(Unit),
+                //TODO : Not sure here
+            setOf(enumValues<P>()[0]),
```
chippmann, line 177: same here

src/gd_kotlin.cpp
```diff
@@ -4,6 +4,10 @@
 #include "core/os/os.h"
 #include "core/project_settings.h"
 #include "bootstrap.h"
+#include "type_manager.h"
+
+// If changed, remember to change also TransferContext::bufferCapacity on JVM side
+const int DEFAULT_SHARED_BUFFER_SIZE{20000000};
```
chippmann, line 10: I don't know if in cpp there is also another way of writing long numbers more readable like in kotlin with `_`. If so i would use that. Otherwise just mark as resolved ;-)

kt/godot-runtime/src/main/kotlin/godot/core/TransferContext.kt
```diff
@@ -1,94 +1,68 @@
...
+import godot.util.threadLocalLazy
 import java.nio.ByteBuffer
+import java.nio.ByteOrder
 
 object TransferContext {
-    private var threadLocalBuffer = ThreadLocal.withInitial { ByteBuffer.allocateDirect(100) }
-    var buffer: ByteBuffer
-        get() = threadLocalBuffer.get()
-        set(value) { threadLocalBuffer.set(value) }
+    var bufferSize = 20000000
```
chippmann, line 9: And maybe also add a comment like you did on the cpp side

src/gd_kotlin.cpp
```diff
@@ -69,9 +73,15 @@ void unload_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes)
 }
 
 void register_engine_types_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_engine_types, jobjectArray p_method_names) {
+    print_line("Starting to register managed engine types...");
```
chippmann, line 76: Shouldn't this be a verbose log?

src/gd_kotlin.cpp
```diff
@@ -81,12 +91,10 @@ void register_engine_types_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_en
     j_object.delete_local_ref(env);
     engine_types.delete_local_ref(env);
     method_names.delete_local_ref(env);
+    print_line("Done registering managed engine types...");
```
chippmann, line 94: Same here

src/gd_kotlin.cpp
```diff
@@ -129,6 +138,13 @@ void GDKotlin::init() {
             if (split_jvm_debug_argument(cmd_arg, result) == OK) {
                 gc_thread_period_interval = result.to_int64();
             }
+        } else if (cmd_arg.find("--jvm-to-engine-shared-buffer-size") >= 0) {
+            String result;
+            if (split_jvm_debug_argument(cmd_arg, result) == OK) {
+                jvm_to_engine_shared_buffer_size = result.to_int();
+                print_line(vformat("Warning ! Buffer capacity was changed to %s, this is not a recommended practice",
```
chippmann, line 145: Shouldn't this be a warning?

chippmann: Maybe also add a todo so the warning links to the documentation once we have one

src/kt_variant.h
```diff
@@ -2,28 +2,356 @@
...
+                }
+            }
+        }
+
+        des->increment_position(encode_uint32(TypeManager::get_instance().JAVA_ENGINE_TYPES_CONSTRUCTORS[class_name], des->get_cursor()));
+        des->increment_position(encode_uint32(src.is_ref(), des->get_cursor()));
+        des->increment_position(encode_uint64(ptr->get_instance_id(), des->get_cursor()));
+    }
+
+    static void init_to_kt_methods(void (* to_kt_array[27 /* Variant::Type count */])(SharedBuffer*, const Variant&)) {
```
chippmann, line 162: This 27 appears some times in this file, maybe a const would make sense here for readability?

chippmann: Then also the comment would be obsolete

chippmann (changes requested):
All my change request are pretty minor. All in all awesome work! Thx a lot for the work!

chippmann:
@piiertho also i get:
```
Exception in thread "main" java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
        at godot.Object.connect(tests.kt:26)
        at godot.signals.Signal.connect$godot_library(Signals.kt:23)
        at godot.Object.connect(tests.kt:55)
        at godot.Object.connect$default(tests.kt:52)
        at godot.tests.Invocation._ready(Invocation.kt:122)
        at godot.Entry$init$1$2$26.invoke(Entry.kt:43)
        at godot.Entry$init$1$2$26.invoke(Entry.kt:24)
        at godot.core.KtFunction0.invoke$godot_runtime(Functions.kt:78)
        at godot.core.KtFunction.invoke(Functions.kt:64)
terminate called after throwing an instance of 'jni::JniError'
```

when running the `tests` sample because `as Int` should be `as Long` apparently:
`return GodotError.values()[TransferContext.readReturnValue() as Int]`
```kotlin
    fun connect(name: String, target: Object, method: String, binds: VariantArray, flags: Long): GodotError {
        val refreshBuffer = TransferContext.writeArguments(
                VariantType.STRING to name,
                VariantType.OBJECT to target,
                VariantType.STRING to method,
                VariantType.ARRAY to binds,
                VariantType.LONG to flags
        )
        TransferContext.callMethod(rawPtr, OBJECT, OBJECT_CONNECT, VariantType.LONG)
        return GodotError.values()[TransferContext.readReturnValue() as Int]
    }
```

piiertho:
> @piiertho also i get:
> 
> ```
> Exception in thread "main" java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
>         at godot.Object.connect(tests.kt:26)
>         at godot.signals.Signal.connect$godot_library(Signals.kt:23)
>         at godot.Object.connect(tests.kt:55)
>         at godot.Object.connect$default(tests.kt:52)
>         at godot.tests.Invocation._ready(Invocation.kt:122)
>         at godot.Entry$init$1$2$26.invoke(Entry.kt:43)
>         at godot.Entry$init$1$2$26.invoke(Entry.kt:24)
>         at godot.core.KtFunction0.invoke$godot_runtime(Functions.kt:78)
>         at godot.core.KtFunction.invoke(Functions.kt:64)
> terminate called after throwing an instance of 'jni::JniError'
> ```
> 
> when running the `tests` sample because `as Int` should be `as Long` apparently:
> `return GodotError.values()[TransferContext.readReturnValue() as Int]`
> 
> ```kotlin
>     fun connect(name: String, target: Object, method: String, binds: VariantArray, flags: Long): GodotError {
>         val refreshBuffer = TransferContext.writeArguments(
>                 VariantType.STRING to name,
>                 VariantType.OBJECT to target,
>                 VariantType.STRING to method,
>                 VariantType.ARRAY to binds,
>                 VariantType.LONG to flags
>         )
>         TransferContext.callMethod(rawPtr, OBJECT, OBJECT_CONNECT, VariantType.LONG)
>         return GodotError.values()[TransferContext.readReturnValue() as Int]
>     }
> ```

Ah yep, that's a mistake of mine. As Godot returns long and not int it should be
```kotlin
return GodotError.values()[(TransferContext.readReturnValue() as Long).toInt()]
```

src/gd_kotlin.cpp
piiertho: IIRC it is `20'000'000`

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -15,28 +15,37 @@ open class Object(isRef: Boolean = false) : KtObject(isRef) {
     }
 
     fun connect(name: String, target: Object, method: String, binds: VariantArray, flags: Long): GodotError {
-        val refreshBuffer = TransferContext.writeArguments(KtVariant(name), KtVariant(target), KtVariant(method), KtVariant(binds), KtVariant(flags))
-        TransferContext.callMethod(rawPtr, OBJECT, OBJECT_CONNECT, KtVariant.Type.LONG, refreshBuffer)
-        return GodotError.values()[TransferContext.readReturnValue().asInt()]
+        val refreshBuffer = TransferContext.writeArguments(
```
chippmann, line 18: Oh and: Why do these local variables exist?
They are unused and of type `Unit`

piiertho: legacy, forgot to change this

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: `null` default value is not supported for now IIRC.
I don't think we can even register a nullable type for now.

chippmann: Hmm we definitely have to support that i think. But let's do it in a separate PR then. Your Todo is fine for me :-)

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -0,0 +1,284 @@
...
+                buffer.putLong(any.rawPtr)
+                buffer.bool = any.isRef
+            }
+    ),
+    DICTIONARY(
+            { buffer: ByteBuffer -> TODO() },
+            { buffer: ByteBuffer, any: Any -> TODO() }
+    ),
+    ARRAY(
+            { buffer: ByteBuffer -> VariantArray() },
```
chippmann, line 246: This impl isn't correct isn't it? It's just a placeholder right? If so maybe add a Todo (that does not crash with not yet implemented) so we don't forget to change it.

chippmann (approved):
One minor comment that does not prevent merging IMO

chippmann:
Updated entry generator ref.

CedNaru (approved):

raniejade (approved):

PR #30, opened 2020-11-26
Optimization/method bind optim and better bench

piiertho:
This improves icall performances caching `MethodBind` pointer.
This improves measures of benchmarks and add sufficient warmup iterations.
This improves benchmark results prints readbility.


harness/benchmarks/scripts/Main.gd
```diff
@@ -1,59 +1,108 @@
...
+			__do_run(warmup, benchmark, stats, true)
+	for iteration in range(iterations):
+		__do_run(iteration, benchmark, stats, false)
+	printt("%s:" % benchmark.lang.name, str(stats))
+	report.add(benchmark, stats.get_results())
+
+
+func __do_run(iteration: int, benchmark: Benchmark, stats: Stats, is_warmup: bool):
+	var start: float = OS.get_ticks_usec()
+	benchmark.exec()
```
raniejade, line 59: why not use iteration here as a loop?

harness/benchmarks/src/main/kotlin/godot/benchmark/Simple.kt
```diff
@@ -54,20 +54,25 @@ class Simple : Object() {
...
+    fun benchmarkIcall(): Int {
         val node = Node()
         node.getInstanceId()
-        return node
+        node.free()
+        return 0
     }
 
-    fun benchmarkICallWithLoop(): Node {
+    fun benchmarkIcallWithLoop(): Int {
```
raniejade, line 64: out of curiosity why the change to returning an int?

harness/benchmarks/scripts/Main.gd
piiertho: Because gdscript loops cost a lot as we can see in other bench. Si to avoid to mesure the cost of loop and so be more precise, we manually wrote calls.

harness/benchmarks/src/main/kotlin/godot/benchmark/Simple.kt
piiertho: Can be unit. Fix here is that we free node. Not freeing it was corrupting benchmark as GC was checking a list that was becoming bigger and bigger. That was causing cost augmentation in function of iterations number.

CedNaru (approved):

harness/benchmarks/gd/Simple.gd
```diff
@@ -1,46 +1,51 @@
 func benchmark_simple_add():
-    var a = 1
-    var b = 2
-    return a + b
+	var a = 1
```
chippmann, line 2: We need to setup our godot editors and IDE's to either use spaces or tabs for gdscript. Else we permanently have changes in those files.
Nothing to do with this particular PR but definitely something we have to look at sooner than later IMO

harness/benchmarks/gd/Simple.gd
```diff
@@ -1,46 +1,51 @@
 func benchmark_simple_add():
```
chippmann, line 1: Shouldn't we now use strong typing in gdscript as well? IIRC the perf benefits of typed gdscript have been backported to 3.2 and will be available with the next patch release. IMO it would be more fair for gdscript to benefit from that in the benchmark.
Best would be to have two gdscript benchmarks: one typed and one not.

harness/benchmarks/scripts/Benchmarks.gd
```diff
@@ -1,23 +1,53 @@
...
+		scripts.append([cls.new(), method_list, lang])
+		
+		for method_dict in method_dict_list:
+			var method_name = method_dict["name"]
+			
+			#check the method name stats with "benchmark_" so we don't test methods like "init"
+			if method_name.begins_with("benchmark_"):
+				method_list.append(method_name)
+	
+				#If not already in the global method last, we add it.
```
chippmann, line 33: typo: list not last

harness/benchmarks/scripts/Benchmarks.gd
```diff
@@ -1,23 +1,53 @@
...
+				if !methods.has(method_name):
+					methods.append(method_name)
+				
+	
+	methods.sort()
+	print("Tests to run:")
+	##Loop over the methods, then the scripts so they are all next toeach other for a same test.
```
chippmann, line 40: typo: toeach -> to each

harness/benchmarks/scripts/Main.gd
chippmann: Understandable. But maybe add a comment explaining this?

harness/benchmarks/scripts/Stats.gd
```diff
@@ -1,66 +1,69 @@
...
\ No newline at end of file
+	var results = get_results()
+	var _min = results.min
+	var _max = results.max
+	var avg = results.avg
+	var median = results.median
+	var pc05 = results.p05
+	var pc95 = results.p95
+	var args = [avg, _min, pc05, median, pc95, _max]
+	return "AVG=%d op/s, MIN=%d op/s, 05PC=%d op/s, MEDIAN=%d op/s, 95PC=%d op/s, MAX=%d op/s" % args
```
chippmann, line 69: could be use tabs between the results rather than spaces? so the numbers are evenly spaced for all benchmark methods? This would greatly improve readability IMO

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -19,6 +19,8 @@ object GarbageCollector {
 
     private var forceJvmGarbageCollector = false
 
+    var shouldDisplayLeakInstancesOneClose = true
```
chippmann, line 22: Typo: shouldDisplayLeakInstancesOneClose -> shouldDisplayLeakInstancesOnClose

src/gd_kotlin.cpp
```diff
@@ -154,6 +167,9 @@ void GDKotlin::init() {
             is_gc_activated = false;
             //TODO: Link to documentation
             WARN_PRINT("GC thread was disable. --jvm-disable-gc should only be used for debugging purpose")
+        } else if (cmd_arg == "--jvm-disable-closing-leaks-warning") {
```
chippmann, line 170: Could we add a little doc to the readme about all the cmdline arguments we can pass in?

chippmann: Because there are quite a few now

chippmann (changes requested):

chippmann:
@CedNaru & @piiertho can you please double check that the csharp benchmarks now really are similar to the other languages?

harness/benchmarks/gd/Simple.gd
piiertho: I agree with this, on my side I got conflicts between intellij and godot editor. Would be simpler to make a PR to solve this issue.

src/gd_kotlin.cpp
piiertho: If you're ok I'll do it in a seperate PR, in which I will create a conf object on cpp side so that we can move command args parsing from GdKotlin singleton. This would improve maintainability. Maybe write an issue about this ?

piiertho:
> @CedNaru & @piiertho can you please double check that the csharp benchmarks now really are similar to the other languages?

Ok on my side :)

harness/benchmarks/gd/Simple.gd
chippmann: Ok for me

src/gd_kotlin.cpp
chippmann: Ok I'll write one and assign it to you

harness/benchmarks/gd/Simple.gd
piiertho: @CedNaru ?

piiertho:
> @CedNaru & @piiertho can you please double check that the csharp benchmarks now really are similar to the other languages?

@CedNaru ?

harness/benchmarks/scripts/Main.gd
piiertho: I agree with this, @CedNaru can you fix benchmark, it was more your modifications IIRC.

CedNaru: Done

harness/benchmarks/scripts/Benchmarks.gd
CedNaru: done

harness/benchmarks/scripts/Stats.gd
CedNaru: done

harness/benchmarks/scripts/Benchmarks.gd
CedNaru: done

harness/benchmarks/scripts/Stats.gd
CedNaru: image
Using tabs gives us ugly results. As the tab push the next string to a multiple of 4 spaces so it doesn't align properly.

harness/benchmarks/gd/Simple.gd
CedNaru: Done. But there is no difference between Typed and Dynamic yet. Those changes are only for Godot 4.0

image

CedNaru:
image

Total is going to be inferred as an integer, so a integer division is going to be used at the end. It's unfair because Integer division is more expensive than a float divison.

CedNaru:
I changed the icall return type to void in gdscript, it should probably be done the same for the other languages.

harness/benchmarks/scripts/Stats.gd
chippmann: Ok let's leave it then

harness/benchmarks/typed_gd/Simple.gd
```diff
@@ -0,0 +1,51 @@
...
+		var v: Vector3 = Vector3(i, i, i)
+		v = b.xform(v)
+		s = s + v
+
+	return s
+
+func benchmark_vectors2_only() -> Vector2:
+	var s: Vector2 = Vector2()
+	for i in range(1000):
+		var v = Vector2(i, i)
```
chippmann, line 29: @CedNaru Not typed. Either type it explicitly or use `:=`

chippmann (changes requested):

chippmann:
> I changed the icall return type to void in gdscript, it should probably be done the same for the other languages.

Changed kotlin and csharp to do the same. @piiertho and @CedNaru , please look over that as well real quick :-)

harness/benchmarks/typed_gd/Simple.gd
CedNaru: done

harness/benchmarks/gd/Simple.gd
chippmann: Cool thx. IIRC they are in the process of being backported to 3.x (but i can't find the PR anymore). But maybe I'm just wrong. Anyways good to have it.

chippmann (approved):

piiertho:
> > I changed the icall return type to void in gdscript, it should probably be done the same for the other languages.
> 
> Changed kotlin and csharp to do the same. @piiertho and @CedNaru , please look over that as well real quick :-)

seems ok to me

CedNaru (approved):

PR #32, opened 2020-12-11
[Working Proposal] Entry Generator Extensions

chippmann:
# [Proposal] Entry generator extensions

Related: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/6

**Note:** this proposal already contains working code for such extensions and a corresponding example.
It's mainly a proposal to discuss the viability and usefulness of such extensions and if the proposed way of implementing
it could be improved ot changed.
The core idea was already proposed in the old project: https://github.com/utopia-rise/godot-kotlin/pull/164 but is
refined here and completed with a working example.

**Note2:** Base for this branch was core types so i set it as base here as well so just my changes are listed in the diff.

## General idea
One can extend the functionality of the entry generator to generate custom code and do various additional validation tasks.
To be able to do this, one can create an entry generator extension and add it as a dependency to his project. The entry
generator (in conjunction with our gradle plugin) automatically picks up the extensions applied to the project and
lets them intercept the entry generation through a predefined api.

## Affected components
- Entry generator
- Godot annotation processor
- Compiler plugin
- Gradle plugin

## General functionality
- Entry generator provides an interface and convenience functions for entry generator extensions.
A entry generator project adds the entry generator as dependency to make use of said interface and convenience functions.
A entry generator project defines a service file which describes the entry point (like we did for the entry file itself).

- The gradle plugin provides a configuration block where a user can add extension dependencies like so:
    ```kotlin
    entryGenerator {
        extensions {
            extension("com.utopia-rise:some-extension") //only applied to entry generation. No dependencies are added to the project itself
            extensionWithAnnotations("com.utopia-rise:godot-node-injection-extension") //also added as a dependency to the users project so he can use the additional annotations provided by this extension
        }
    }
    ```

- The gradle plugin places the built entry generation extensions in the `build/libs` folder, so the annotation processor has one
place where it can pick up extensions.

- During compilation the annotation processor looks into the libs folder and loads all extension through their provided
  service file
- The annotation processor calls `provideFullyQualifiedAnnotations` on each extension and not only collect all registration annotations during the compilation (like it does now), it also collects all custom defined annotations
- The annotation processor calls the entry generator with all found annotations (registration annotations and custom annotations) and with all extensions
- The entry generator generates the registration code like today but lets the extensions it got from the annotation processor inject certain points in the generation process and calls the extensions at certain points so they can generate custom logic

## Extension API
The api and thus capability of entry generator extensions is defined and limited by this abstract class each extension
has to implement: [EntryGeneratorExtension](https://github.com/utopia-rise/godot-kotlin-entry-generator/blob/6a9d85a17fe2ebce3939cea6dfb08f683bf550d2/src/main/kotlin/godot/entrygenerator/extension/EntryGeneratorExtension.kt)

---

# Example
In this proposal one working example of such an extension is included: [godot-node-injection-extension](https://github.com/utopia-rise/godot-jvm/tree/feature/entry-gen-extensions/kt/entry-generation/godot-node-injection-extension)
In "the real world" this would be a separate project, but it is included here for simplicity of this proposal.

### Use case of the example
If one does not want to hardcode NodePaths into scripts, one can `export` NodePath's to godot and populate them in the
inspector.
In gdscript one can leverage the `onready` keyword to retrieve the node with `get_node` and the exported NodePath:

```gdscript
export var nodePathForNode: NodePath
onready var nodeToInject: Spatial = get_node(nodePathForNode) as Spatial
```

Should the node path ever change (for example when rearranging the nodes in the hierarchy or rename them), godot
automatically updates the exported NodePath and thus no script code has to be changed just because the path's are different.

This approach works well for gdScript but introduces a lot of boilerplate code for other scripting languages like kotlin:
```kotlin
@RegisterProperty
lateinit var nodePathForNode: NodePath
lateinit var nodeToInject: Spatial

@RegisterFunction
override fun _ready() {
    nodeToInject = getNode(nodePathForNode) as Spatial
    //...
}
```

With this entry generation extension the kotlin code can be limited to:
```kotlin
@InjectNode
lateinit var nodeToInject: Spatial
```

### How does it work
The entry generation extension registers a new annotation `@InjectNode`.
If the annotation is detected it generates the following code in the entry file:
```kotlin
//...
/**
 * Generated by Godot Node Injection Extension
 */
private var Invocation.nodeToInjectNodePath: NodePath
  get() = Invocation_nodeToInject_nodePathBacking
  set(value) {
    Invocation_nodeToInject_nodePathBacking = value
  }

/**
 * Generated by Godot Node Injection Extension
 */
private object GodotNodeInjectionExtension {
  public var Invocation_nodeToInject_nodePathBacking: NodePath = NodePath()
}
```
and registeres a additional property of type node path for that property:
```kotlin
//...
registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial", false) {
    //...

    // ================EXTENSION BLOCK================
    // START: Generated by extension: Godot Node Injection Extension
    // Extension entry class: godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension
    property(Invocation::nodeToInjectNodePath, NODE_PATH, NODE_PATH, "godot.core.NodePath", NONE, "", NodePath())
    // END: Generated by extension: Godot Node Injection Extension
    // Extension entry class: godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension
    // ================EXTENSION BLOCK================
}
//...
```

and it lets the entry generator override the call to `_ready`:
```kotlin
//...
registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial", false) {
    //...
    function(ExtensionHelper::_ready, NIL, KtFunctionArgument(NIL, "kotlin.Unit"))
    //...
}
//...
private object ExtensionHelper {
  public fun _ready(instance: Invocation): Unit {
    // ==== START EXTENSION BLOCK generated by Godot Node Injection Extension with class godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension ====
    instance.nodeToInject = instance.getNode(instance.nodeToInjectNodePath) as Spatial?
    // ==== END EXTENSION BLOCK generated by Godot Node Injection Extension with class godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension ====
    val result = instance._ready()
    // ==== START EXTENSION BLOCK generated by Godot Node Injection Extension with class godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension ====
    // ==== END EXTENSION BLOCK generated by Godot Node Injection Extension with class godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension ====
    return result
  }
}
//...
```

---


chippmann:
To summarize some discussions on discord:
Generally the idea of allowing extensions to intercept code generation is good but the implementation details were not discussed in detail yet.
What primarily was discussed was the example i added i this Proposal.
The current problem with the code the example generates is that all script instances of one particular script class share the same node path property which could lead to some problems.
@CedNaru had some great ideas in how to solve that problem:
ced_naru_idea

I'll implement this now and add it to this proposal.
Some further discussions are needed and the outcome will be documented here as well.

chippmann:
Changed the extension so it generates the following code which stores a node path per instance and subscribes to newly added `onDestroy` hooks in `KtObject` to clear them if they are not needed anymore:

```kotlin
//...

      registerClass<Invocation>(Invocation::class.qualifiedName!!, "Spatial", false) {
        //...

        // ================EXTENSION BLOCK================
        // START: Generated by extension: Godot Node Injection Extension
        // Extension entry class: godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension
        subscribeToOnDestroy<Invocation> { injectionInstanceToNodePaths.remove(it) }
        property(Invocation::nodeToInjectNodePath, NODE_PATH, NODE_PATH, "godot.core.NodePath", NONE, "", NodePath())
        // END: Generated by extension: Godot Node Injection Extension
        // Extension entry class: godot.entrygenerator.extension.nodeinjection.GodotNodeInjectionExtension
        // ================EXTENSION BLOCK================
      }

//...

/**
 * Generated by Godot Node Injection Extension
 */
private var Invocation.nodeToInjectNodePath: NodePath
  get() = injectionInstanceToNodePaths[this] ?: NodePath()
  set(value) {
    injectionInstanceToNodePaths[this] = value
  }

/**
 * Generated by Godot Node Injection Extension
 */
private object GodotNodeInjectionExtension {
  public val injectionInstanceToNodePaths: MutableMap<KtObject, NodePath> = mutableMapOf()
}
```

The implementation of the newly added hooks in `KtObject`:
```kotlin
//...
    private fun __onInit() {
        onInitListener[this::class]?.forEach { it.invoke(this) }
        _onInit()
    }
    fun __onDestroy() { //not private, but final -> needs to be called through jni
        onDestroyListener[this::class]?.forEach { it.invoke(this) }
        _onDestroy()
    }
//...
    companion object {
        //...

        @PublishedApi
        internal val onInitListener: MutableMap<KClass<out KtObject>, MutableList<(KtObject) -> Unit>> = mutableMapOf()

        @PublishedApi
        internal val onDestroyListener: MutableMap<KClass<out KtObject>, MutableList<(KtObject) -> Unit>> = mutableMapOf()

        //...

        inline fun <reified T : KtObject> subscribeToOnInit(noinline listener: (T: KtObject) -> Unit) {
            onInitListener[T::class]?.add(listener) ?: run { onInitListener[T::class] = mutableListOf(listener) }
        }

        inline fun <reified T : KtObject> unSubscribeFromOnInit(noinline listener: (KtObject) -> Unit) {
            onInitListener[T::class]?.remove(listener)
            if (onInitListener[T::class]?.size ?: 0 == 0) {
                onInitListener.remove(T::class)
            }
        }

        inline fun <reified T : KtObject> subscribeToOnDestroy(noinline listener: (T: KtObject) -> Unit) {
            onDestroyListener[T::class]?.add(listener) ?: run { onDestroyListener[T::class] = mutableListOf(listener) }
        }

        inline fun <reified T : KtObject> unSubscribeFromOnDestroy(noinline listener: (KtObject) -> Unit) {
            onDestroyListener[T::class]?.remove(listener)
            if (onDestroyListener[T::class]?.size ?: 0 == 0) {
                onDestroyListener.remove(T::class)
            }
        }
//...
```

@CedNaru FYI

CedNaru:
I think it's good that the init and destroy lambda are globally stored, but I wouldn't directly use the `MutableMap<KClass<out KtObject>, MutableSet<(KtObject) -> Unit>>`
inside the _onInit and _onDestroy. 
Every Godot class should instead have their own MutableSet<(KtObject) -> Unit> accessible statically so we don't need to retrieve it from a HashMap each time a new instance is created/destroyed.

Also the current implementation doesn't work, because if someone inherits the Invocation Script, the new class won't inherit the lambda from its parent. 
We need _onInit and _onDestry to call their parent implementation as well.

chippmann:
Closing for now as it has no priority and the changes made to support script inheritance and incremental builds make this proposal's implementation incompatible.
Once we're at a more stable stage I reopen the PR with an updated impl.

PR #34, opened 2020-12-22
Fix and improve setup of additional compilation

chippmann:
This fixes our build setup for the formerly named `dummy` target.  
Before we added a `dummy` target **before** the `main` compilation to generate the entry file.  
To this dummy compilation we added all the dependencies the user added to his `main` compilation.  

This worked well for normal `implementation` dependencies, but did not work for `kapt` dependencies. Making the use of `kapt` impossible as no `kapt` sources were generated for the `dummy` target thus the `dummy` compilation could not compile while the `main` compilation would (an example of such a dependency would be [dagger](https://dagger.dev/)).  

Another problem was that we added all dependencies of the `main` compilation as `implementation` to the `dummy` target rather that as what they were added to the `main` compilation. For example if the user added a dependency transitive (using `api` for example) the dependency was not transitive for the `dummy` target, which could lead to some errors too.

This PR fixes that by restructuring our build setup.  
Rather than have a `dummy` target **before** the main compilation. We now add a sort of *finalisation* compilation (named `game` compilation) **after** the `main` compilation and just add all outcome of the `main` compilation to it.

This has several advantages:
Regardless of what the user is doing in the `main` compilation, we should have it as well in the `game` compilation as we add the output of that compilation to it.
We don't have to check what and which type of dependency a user has added to the code, as we don't need those. As said, we just use the outcome.
The user can use other code generation frameworks and other annotation processors as well, as (you guessed it) we just use the outcome. So we don't have to apply the same annotation processing and code generation logic to the `dummy` (now the `game` compilation) target.

The entry file is now not generated in the `game` compilation (formerly `dummy` target), it is generated during the `main` compilation and then added as additional sources to the `game` compilation.


kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * Leveraging the incremental build capabilities the additional time needed to compile the exact same code again with just one file added (the entry file), the additional time needed is negligible.
+ *
+ * Another way would be to directly generate Kotlin IR (intermediate representation). Then we would be able to add the generated entry sources to the current compilation process and would not need to do a second compilation.
+ *
+ * We opted to not do this for maintainability as well as easier debugging for the user as the user can look at the generated entry file, set breakpoints and get a meaningful stacktrace in case of an error.
+ *
+ * We also not used a normal annotation processor to be able to use the full information the compiler has at compile time for the entry generation process. Again increasing maintainability.
+ */
+fun Project.setupConfigurationsAndCompilations(jvm: KotlinJvmProjectExtension, project: Project) {
```
raniejade, line 36: `this` here is a `Project, why do we need to pass in a `project` instance?

raniejade (changes requested):

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
chippmann: Because i did not pay attention when i instructed intellij to extract that code into a separate function ^^ I'll fix that :-)

raniejade (approved):

CedNaru (approved):

piiertho (approved):

PR #35, opened 2020-12-23
Feature/core types

piiertho:
This implements core types.
This fixes a memory leak when passing GC handled types as icall param.
This introduce a bridge manager to handle bridges glue code in one place.

Edit from @chippmann: Corresponding Entry-Gen PR: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/7


piiertho:
This needs JVM_UBYTE registration before merge

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
```diff
@@ -139,8 +189,124 @@ class Invocation : Spatial() {
 
     override fun _onDestroy() {
         invocation.free()
+        navMeshes.clear()
+        navMeshesDictionary.clear()
+        anyToAnyDictionary.clear()
+        nullableArray.clear()
+        nullableDictionary.clear()
+        testArrayAny.clear()
```
chippmann, line 197: Why are these needed? Shouldn't they be cleared by the GC?

harness/tests/test/unit/test_node_path.gd
```diff
@@ -0,0 +1,12 @@
+extends "res://addons/gut/test.gd"
+
+
+func test_should_find_camera_node():
+	var invocation_scene = load("res://Spatial.tscn").instance()
+	assert_true(invocation_scene.has_camera_node(), "Should find Camera Node using not path")
```
chippmann, line 6: Typo: ... using **node** path

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -337,20 +335,30 @@ open class Node : Object() {
...
-        TransferContext.readReturnValue(VariantType.NIL)
     }
 
     open fun removeChild(node: Node) {
         TransferContext.writeArguments(VariantType.OBJECT to node)
         TransferContext.callMethod(rawPtr, NODE_REMOVE_CHILD, VariantType.NIL)
-        TransferContext.readReturnValue(VariantType.NIL)
+    }
+
+    open fun getNode(nodePath: NodePath): Node? {
```
chippmann, line 352: I think we should also provide the same method with a string as a param and internally wrap that string into a nodePath for convenience

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -337,20 +335,30 @@ open class Node : Object() {
...
+        TransferContext.writeArguments(VariantType.NODE_PATH to nodePath)
+        TransferContext.callMethod(rawPtr, NODE_GET_NODE, VariantType.OBJECT)
+        return TransferContext.readReturnValue(VariantType.OBJECT, true) as Node?
+    }
+
+    open fun getNodeOrNull(nodePath: NodePath): Node? {
```
chippmann, line 358: same here

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -337,20 +335,30 @@ open class Node : Object() {
...
     open fun removeChild(node: Node) {
         TransferContext.writeArguments(VariantType.OBJECT to node)
         TransferContext.callMethod(rawPtr, NODE_REMOVE_CHILD, VariantType.NIL)
-        TransferContext.readReturnValue(VariantType.NIL)
+    }
+
+    open fun getNode(nodePath: NodePath): Node? {
+        TransferContext.writeArguments(VariantType.NODE_PATH to nodePath)
+        TransferContext.callMethod(rawPtr, NODE_GET_NODE, VariantType.OBJECT)
+        return TransferContext.readReturnValue(VariantType.OBJECT, true) as Node?
```
chippmann, line 355: Why is the return type nullable here? Isn't `getNodeOrNull` for nullables and this one should crash if it's null?

kt/godot-runtime/src/main/kotlin/godot/core/Dictionary.kt
```diff
@@ -0,0 +1,335 @@
...
+    val valVariantType = variantMapper[V::class]
+    checkNotNull(valVariantType) {
+        "Can't create a Dictionary with generic value ${V::class}."
+    }
+    return Dictionary<K, V>(keyVariantType, valVariantType)
+}
+
+inline fun <reified K, reified V> dictionaryOf(vararg args: Pair<K, V>) = Dictionary<K, V>().also {
+    it.putAll(args)
+}
```
chippmann, line 335: Shouldn't we also have convenience functions to convert a Map into a dictionary other than the corresponding constructor?
Like extension functions for Map itself?

kt/godot-runtime/src/main/kotlin/godot/core/NativeCoreTypeWeakReference.kt
```diff
@@ -0,0 +1,8 @@
+package godot.core
+
+import java.lang.ref.WeakReference
+
+internal class NativeCoreTypeWeakReference(referent: NativeCoreType)
+    : WeakReference<NativeCoreType>(referent) {
+    val coreVariantType = referent.coreVariantType
```
chippmann, line 7: doesn't this create a normal reference to `coreVariantType` and thus the ref to `referent` isn't weak anymore?
Souldn't it be (pseudo code): 
```kotlin
val coreVariantType: VariantType
    get() = referent.coreVariantType
```

kt/godot-runtime/src/main/kotlin/godot/core/Properties.kt
```diff
@@ -62,10 +68,27 @@ class KtEnumProperty<T : KtObject, P : Any>(
...
-        require(argsSize == 1) { "Setter should be called with only one argument." }
-        val arg = TransferContext.readSingleArgument(VariantType.JVM_INT)
-        TransferContext.buffer.rewind()
-        kProperty.set(instance, setValueConverter(arg as Int))
+        val arg = extractSetterArgument<Int>()
+        kProperty.set(instance, setValueConverter(arg))
+    }
+}
+
+class KtArrayProperty<T: KtObject, P: Any?>(
```
chippmann, line 76: This is not used atm. is it needed?

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -1,9 +1,42 @@
 package godot.core
 
+import godot.core.VariantType.*
 import godot.util.toRealT
 import java.nio.ByteBuffer
 
 
+val variantMapper = mutableMapOf(
```
chippmann, line 8: Isn't the new `JVM_BYTE` missing?

kt/godot-runtime/src/main/kotlin/godot/core/VariantArray.kt
```diff
@@ -1,57 +1,472 @@
...
+            variantType
+    )
+}
+
+//HELPER
+inline fun <reified T> variantArrayOf(vararg args: T) = VariantArray<T>().also { it.addAll(args) }
+
+/**
+ * Convert an iterable into a GodotArray
+ * Warning: Might be slow if the iterable contains a lot of items because GDNative can only append items one by one
```
chippmann, line 466: Maybe add a Todo here as this is about to change (not sure if only 4.0 or if it will be backported to 3.2

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolRealArray.kt
```diff
@@ -0,0 +1,172 @@
...
+
+package godot.core
+
+import godot.util.IndexedIterator
+import godot.util.RealT
+import godot.util.VoidPtr
+import godot.util.realTVariantType
+
+@Suppress("MemberVisibilityCanBePrivate", "unused")
+class PoolRealArray : NativeCoreType, Iterable<RealT> {
```
chippmann, line 11: Why are the pool arrays not inheriting mutable collection? just because of the function naming for append and so on? or does it have another reason as well?

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -78,6 +78,32 @@ class ClassBuilderDsl<T : KtObject>(
         )
     }
 
+    fun <P: Any> arrayProperty(
```
chippmann, line 81: Not used at the moment. Is this needed? See also my comment on `KtArrayProperty`

chippmann (changes requested):
Awesome work! Some minor comments though

kt/godot-library/src/main/kotlin/godot/tests.kt
piiertho: This will be generated with json api, maybe we should do extention function after api generation.

kt/godot-library/src/main/kotlin/godot/tests.kt
piiertho: Nop, `get_node` will call `get_node_or_null` and add an addtional error print.

kt/godot-runtime/src/main/kotlin/godot/core/Dictionary.kt
piiertho: yep we can add some I think.

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
piiertho: Forgot to remove them, was during debugging

kt/godot-runtime/src/main/kotlin/godot/core/NativeCoreTypeWeakReference.kt
CedNaru: It's a normal reference to coreVariantType yes, but this has no effect on referent.

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
CedNaru: You are right, we forgot to add it to the allowed types.

kt/godot-runtime/src/main/kotlin/godot/core/VariantArray.kt
CedNaru: Actually, it can be changed right now.
GDNative didn't have this function, but now we are in a module so I can get the true equal operator from the C++

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolRealArray.kt
CedNaru: No particular reason, I just kinda implemented Itererable and didn't think about using a more specialized interface.

kt/godot-runtime/src/main/kotlin/godot/core/VariantArray.kt
piiertho: We will backport to 3.2 I think, 4.0 should not be so much changes.

kt/godot-runtime/src/main/kotlin/godot/core/Properties.kt
CedNaru: Was that to store the type of the Array? 
If that's the case, it's not necessary anymore, we already have the type information stored in VariantArray

piiertho: was to register array, but is deprecated, I removed it.

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
piiertho: done

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: removed

kt/godot-runtime/src/main/kotlin/godot/core/Dictionary.kt
piiertho: done

kt/godot-library/src/main/kotlin/godot/tests.kt
chippmann: Ok for me if we do it there

kt/godot-library/src/main/kotlin/godot/tests.kt
chippmann: I see. Thx for the clarification

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolRealArray.kt
chippmann: shouldn't we then not use mutable collection?
Then one can use all of the kotlin collection function

chippmann (approved):

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolRealArray.kt
piiertho: Maybe we can do it after merging this as an enhancement ? This would allow us to split last tasks for alpha.

chippmann: Ok for me

CedNaru (approved):

raniejade (approved):
Sorry for the delay!

PR #36, opened 2020-12-25
Initial Godot Jvm Idea Plugin implementation

chippmann:
This adds an initial implementation of a Idea plugin for Godot Jvm.

This plugin adds some additional code inspections for stuff that would lead to compilation errors we throw during entry generation but would not regularily be detected as an error by the IDE because it's correct kotlin syntax. 

Currently implemented inspections:
- If a class has registered members but is itself not registered -> shows error at class in IDE and supplies quick fix action to register it
- If a constructor of a class has more than 5 params -> shows error at corresponding constructors in IDE
- If a property has an invalid default value -> shows error at initializer in IDE
- If property is not mutable -> shows error at `val` in IDE and supplies quick fix to change to `var`
- If a property has an unregistrable type -> shows error at property in IDE
- If a signal property is mutable -> shows warning at `var` in IDE and supplies quick fix to change to `val`
- If a signal property has a wrong type -> shows error at initializer in IDE and supplies quick fix to change to signal delegate
- If engine function (like `_ready`) is overriden but not registered -> shows warning at function in IDE and supplies quick fix to register it
- All type checks for property hint annotations

Immediate roadmap:
~~Type hint inspections are not implemented yet as the type hint annotations are not yet added to this project (will do that next).~~ <- included now
After the type hint inspections I'll add gutter icons for signals which are connected in the godot editor and mark those functions as used.

**Note:** I could not use the same error detections for the default values i used in the entry generator as even the psi types i get from the IDE are fundamentally different than those i get from the compiler. Even though it's the same psi. So i opted for a "duplication" of the checks where needed.  
I will investigate a possibility to share the errordetection between the entry generator and the idea plugin tough. But for this initial implementation i thought this is fine. If you don't agree and think otherwise, say so and I'll investigate it in this PR.


kt/godot-library/src/main/kotlin/godot/annotation/propertyHintAnnotations.kt
```diff
@@ -0,0 +1,265 @@
...
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ColorNoAlpha
+
+/**
+ * Can only be used on Image properties!
+ */
+//@Target(AnnotationTarget.PROPERTY)
+//@Retention(AnnotationRetention.RUNTIME)
+//annotation class ImageCompressLossy
```
piiertho, line 149: Is this for future or should it be removed ?

kt/godot-library/src/main/kotlin/godot/annotation/propertyHintAnnotations.kt
```diff
@@ -0,0 +1,265 @@
...
+
+The following annotations are implemented based on https://github.com/godotengine/godot/blob/3.2/core/object.h
+ */
+
+/**
+ * Can only be used on Int properties!
+ */
+@Target(AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class IntRange(val start: Int, val end: Int, val step: Int = -1, val or: Range = Range.NONE)
```
CedNaru, line 17: We have a Float and Double range. Shouldn't we have LongRange then ? We only got a IntRange.

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterFunctionInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+                    engineFunctionNotRegisteredQuickFix
+                )
+            }
+        }
+    }
+
+    private val engineFunctions = listOf(
+        //TODO: add all engine functions or find a better way of checking all -> maybe add godot-library as a dependency and query engine functions through reflection or generate during compilation of the plugin
+        "_ready",
+        "_process"
```
piiertho, line 38: Here is the list:
https://docs.godotengine.org/fr/stable/getting_started/workflow/best_practices/godot_notifications.html

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterFunctionInspection.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun buildVisitor(holder: ProblemsHolder, isOnTheFly: Boolean): PsiElementVisitor {
+        return namedFunctionVisitor { ktNamedFunction ->
+            if (
+                ktNamedFunction.containingClass()?.findAnnotation(FqName(RegisterClass::class.java.canonicalName)) != null
+                && engineFunctions.contains(ktNamedFunction.name)
+                && ktNamedFunction.findAnnotation(FqName(RegisterFunction::class.java.canonicalName)) == null
+            ) {
+                holder.registerProblem(
+                    ktNamedFunction.navigationElement,
+                    "Overridden engine function which is not registered will not be called by Godot. Using engine functions for other purposes than to be called from Godot is considered a bad practise. Either register it or move your logic to a custom function you defined",
```
piiertho, line 27: "Engine Function" is not appropriate IMO.
I think "notification function" would be better.

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterPropertiesInspection.kt
```diff
@@ -0,0 +1,278 @@
...
+    private fun checkMutability(ktProperty: KtProperty, holder: ProblemsHolder) {
+        if (!ktProperty.isVar) {
+            holder.registerProblem(ktProperty.valOrVarKeyword, "Registered properties have to be mutable", mutabilityQuickFix)
+        }
+    }
+
+    private fun checkRegisteredType(ktProperty: KtProperty, holder: ProblemsHolder) {
+        val type = ktProperty.type() ?: return
+        if (type.getJetTypeFqName(false).startsWith("kotlin.collections") && ktProperty.findAnnotation(FqName("godot.annotation.EnumFlag")) == null) { //enum flag is the only case where registering a kotlin collection is allowed
+            holder.registerProblem(getInitializerProblemLocation(ktProperty), "Kotlin collections cannot be registered as default values. Consider using one of the godot collections. There are handy conversion functions available")
```
piiertho, line 54: Can't we add a quick fix using collection fqdn to provide the conversion ?

piiertho (changes requested):
Really awesome work (did not tested it in editor for now, will try in next few days.
I only had few questions while reading the code.

kt/plugins/godot-intellij-plugin/build.gradle.kts
```diff
@@ -0,0 +1,98 @@
...
+version = "0.0.1"
+
+dependencies {
+    implementation(project(":godot-library"))
+    detektPlugins("io.gitlab.arturbosch.detekt:detekt-formatting:1.14.2")
+}
+
+// Configure gradle-intellij-plugin plugin.
+// Read more: https://github.com/JetBrains/gradle-intellij-plugin
+intellij {
```
raniejade, line 31: If we are planning to allow this plugin to be used in android studio at some point, then I would suggest to have a release artifact per minor version of IJ. See https://github.com/spekframework/spek/blob/2.x/spek-ide-plugin-intellij-idea/build.gradle.kts#L42.

kt/plugins/godot-intellij-plugin/build.gradle.kts
```diff
@@ -0,0 +1,98 @@
...
+        xml.enabled = false
+        txt.enabled = false
+    }
+}
+
+tasks {
+    patchPluginXml {
+        version(version)
+        sinceBuild("201")
+        untilBuild("203.*")
```
raniejade, line 61: This will eventually bite you, significant code changes can happen between those versions - then you will eventually need to provide to artifacts one before the significant change and the other after. As I mentioned above, I would suggest to have a release per minor version, so for this part: sinceBuild = 201, untilBuild = 201.* (this means any 201 version)

raniejade (commented):
You probably don't need the actual annotation classes here, you just need the fully qualified name. Anyway, up to you if you want to investigate it here.

kt/godot-library/src/main/kotlin/godot/annotation/propertyHintAnnotations.kt
chippmann: Hmm i guess you're right.
I'll add it

kt/godot-library/src/main/kotlin/godot/annotation/propertyHintAnnotations.kt
chippmann: These are for future implementation. We did not add those to the kotlin-native project because we were not sure how they work and though we should only add them once we know how they work. So i did the same here.
Once we have spare time we can add these. I thought they just are not that important atm.

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterFunctionInspection.kt
chippmann: Cool thx. I'll add these.
But i leave the todo there so i remember to find a better way to add those rather than hardcoding them

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterFunctionInspection.kt
chippmann: You are right. They are even named `notification functions` in the godot documentation.
I'll change that

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterPropertiesInspection.kt
chippmann: Sure we can. But for this case it might not be that trivial. If it's ok for you I'll just add a todo for now and add them later once the whole plugin is a bit more mature

chippmann:
> You probably don't need the actual annotation classes here, you just need the fully qualified name. Anyway, up to you if you want to investigate it here.

@raniejade I actually had that initially. Though it's cleaner this way. But looking at it again, you're probably right. It's overkill to add the whole `godot-library` as a dependency just for the annotation fqNames

kt/plugins/godot-intellij-plugin/build.gradle.kts
chippmann: Hmm didn't think of that. I just used the official jetbrains template and this is configured by it like this. I'll take a look at your spek framework plugin and change this here.

chippmann: @raniejade i changed it now to what i think matches what you did in the spek framework plugin. Could you check my impl if this holds up?

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/inspector/RegisterPropertiesInspection.kt
piiertho: I'm fine with it :-)

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/quickfix/EngineFunctionNotRegisteredQuickFix.kt
```diff
@@ -0,0 +1,19 @@
+package godot.intellij.plugin.inspections.registration.quickfix
```
piiertho, line 1: Name and file and class should not change to notification ?

piiertho (changes requested):
Only one small change for me.

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/inspections/registration/quickfix/EngineFunctionNotRegisteredQuickFix.kt
chippmann: Ups. Cood catch. I'll change it.

piiertho (approved):
Thx !

CedNaru (approved):

raniejade (approved):
Looks good! You probably want to add tests for the plugin :)

piiertho (changes requested):
I think this will need #38 to be merge before and then rebase and add PR check for this plugin in github actions.
WDYT ?

chippmann:
@piiertho Yeah indeed. Thats necessary now. I'll add the yaml config

chippmann:
@raniejade Indeed the tests are missing for now. If you are ok with it i'd rather first fix #39 as it's kind of a blocker and add the tests in a separate PR.

chippmann:
> @raniejade Indeed the tests are missing for now. If you are ok with it i'd rather first fix #39 as it's kind of a blocker and add the tests in a separate PR.

@raniejade I leave the tests for a later date. I spent almost an entire day to figure out how one can test kotlin code inspections and still have no clue. Maybe i invest some time again but until we have an useful alpha, I'd rather spend my time on other things that are more fun to do.

raniejade (approved):

CedNaru (approved):

piiertho (approved):

PR #37, opened 2020-12-29
Enhancement/easy to maintain command line parsing

### LINKED ISSUE #31: Improve handling of commandline args
We now have quite a few command line arguments that one can supply to change how the jvm/cpp interaction works.
It would simplify things a lot if we had a table in the readme that lists all available arguments and their functionality.
If written well, this can be later copied one to one to the documentation.

piiertho:
This adds JvmConfig class to extract command line arguments parsing.  
This also resolves #31 by adding documentation in readme and code.


README.md
```diff
@@ -57,6 +57,21 @@ Each Issue has a Maintainer that is the "supervisor" for the general topic the i
 7. In order to debug your jvm code, you should start godot with command line `--jvm-debug-port=XXXX`, where `XXXX`
 stands for the jmx port of you choice. You can then setup remote debug configuration in Intellij' Idea.
 
+## Command line arguments
+
+In order to debug, or override behaviour of this module, there are some command lines arguments you can add when
+starting godot:  
+| Command line argument               | Description                                                                                                                | Usage                                                                                        |
```
chippmann, line 64: Personally i find the usage section obsolete. IMO it just duplicates whats i  the first two sections

src/gd_kotlin.cpp
```diff
@@ -232,18 +166,19 @@ void GDKotlin::init() {
 
     BridgesManager::get_instance().initialize_bridges(env, class_loader);
 
-    if (is_gc_activated) {
+    if (jvm_config.is_gc_activated) {
+        bool is_gc_force_mode{jvm_config.is_gc_force_mode};
```
chippmann, line 170: Couldn't this be inlined below?

src/jvm_config.cpp
```diff
@@ -0,0 +1,104 @@
+#include <core/os/os.h>
+#include "jvm_config.h"
+
+// If changed, remember to change also TransferContext::bufferCapacity on JVM side
+const int DEFAULT_SHARED_BUFFER_SIZE{20'000'000};
+const int DEFAULT_JVM_GARBAGE_COLLECTOR_THREAD_PERIOD{500};
+
+JvmConfig::JvmConfig(const String& p_jvm_debug_port, const String& p_jvm_debug_address, const String& p_jvm_jmx_port,
```
chippmann, line 8: Is it really common to format constructor calls like this? I find this incredibly hard to read.

chippmann (changes requested):
Some minor comments. Good work though.

src/gd_kotlin.cpp
piiertho: We use it twice, here and 5 lines below.

src/jvm_config.cpp
piiertho: It is, but we can also format it a bit better, that's now the case

README.md
piiertho: I find that it does not permit any ambiguity about the usage.  
We have commands with `=` and others without.  
Usage section is here so that user know the different ways to use the command.

chippmann: ok then

src/jvm_config.cpp
chippmann: Thx

chippmann (approved):

CedNaru (approved):

src/jvm_config.h
```diff
@@ -0,0 +1,89 @@
+#ifndef GODOT_JVM_JVM_CONFIG_H
+#define GODOT_JVM_JVM_CONFIG_H
+
+
+#include <core/ustring.h>
+#include <core/list.h>
+#include <modules/kotlin_jvm/src/jni/init_args.h>
```
piiertho, line 7: Oops, seems I did not clean imports

raniejade (commented):
I think we should add a parameter to pass in raw JVM args.

piiertho:
> I think we should add a parameter to pass in raw JVM args.

I added it, with arguments seperated by a `;`

chippmann:
> > I think we should add a parameter to pass in raw JVM args.
> 
> I added it, with arguments seperated by a `;`

I would not use `;` as a separator. On unix you normally use them to issue multiple different commands on one line. Why not use normal `,`? IIRC no jvm arg uses commas internally?

piiertho:
> > > I think we should add a parameter to pass in raw JVM args.
> > 
> > 
> > I added it, with arguments seperated by a `;`
> 
> I would not use `;` as a separator. On unix you normally use them to issue multiple different commands on one line. Why not use normal `,`? IIRC no jvm arg uses commas internally?

Here is one argument with `,`.
Do we have a character that is unused ?

chippmann:
> > > > I think we should add a parameter to pass in raw JVM args.
> > > 
> > > 
> > > I added it, with arguments seperated by a `;`
> > 
> > 
> > I would not use `;` as a separator. On unix you normally use them to issue multiple different commands on one line. Why not use normal `,`? IIRC no jvm arg uses commas internally?
> 
> Here is one argument with `,`:
> 
> https://github.com/utopia-rise/godot-jvm/blob/815d13af8b7f10c03cad861ba12ac00d1005b15e/src/gd_kotlin.cpp#L118
> 
> Do we have a character that is unused ?

`·` (note: not a fullstop)

raniejade (approved):

PR #38, opened 2021-01-02
feature/check pr github actions

piiertho:
This adds check PR CI.
This validates debug and release build of editor and exports templates.
This runs test on debug build and check tests failure and jvm is shutting down correctly.

This also fixes compilation errors on linux and OSX.


.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
```diff
@@ -0,0 +1,90 @@
...
+          scons platform=${{ matrix.platform }}
+      - name: Upload OSX binary
+        if: matrix.platform == 'osx'
+        uses: actions/upload-artifact@v1
+        with:
+          name: osx-editor-debug-app
+          path: bin/godot.osx.tools.64
+  test:
+    needs: build
+    runs-on: macos-latest
```
chippmann, line 63: Why do we run the tests only on osx?

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
```diff
@@ -0,0 +1,90 @@
...
+          build-root-directory: harness/tests/
+          arguments: build
+      - name: Download OSX Editor Debug App
+        uses: actions/download-artifact@v1
+        with:
+          name: osx-editor-debug-app
+      - name: Run Tests
+        run: |
+          cd harness/tests/
+          wget https://gist.githubusercontent.com/piiertho/fb25f9326189d00608ffeb7702b91395/raw/dddef0b20a61062bb4d091bf2f7ada6b7b25b7c7/run_godot_kotlin_tests.sh
```
chippmann, line 86: Why a gist and not a script inside the repo?

src/kt_variant.h
```diff
@@ -137,66 +137,89 @@ namespace ktvariant {
...
-    static void to_kvariant_fromNATIVECORETYPE(SharedBuffer* des, const Variant& src) {
-        des->increment_position(
-                encode_uint64(reinterpret_cast<uintptr_t>(memnew(T(src.operator T()))), des->get_cursor())
-        );
-    }
-
     static void to_kvariant_fromDICTIONARY(SharedBuffer* des, const Variant& src) {
         set_variant_type(des, Variant::Type::DICTIONARY);
-        to_kvariant_fromNATIVECORETYPE<Dictionary>(des, src);
+        des->increment_position(
```
chippmann, line 142: Why is this needed? Did the template we had earlier not work?

chippmann (changes requested):

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
piiertho: Because I plan to use it in other projects :p

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
piiertho: IIRC gut cannot run from command line on windows. I do not remember for Windows. I can try it.

src/kt_variant.h
piiertho: Not on Linux, this was failing on `operator T()`

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
piiertho: IIRC Linux cannot run because we have no graphics on CI

chippmann: Hmm we could use a headless build i guess. But yeah overkill for now and something to maybe lookt at at a later stage

harness/tests/run_godot_kotlin_tests.sh
```diff
@@ -0,0 +1 @@
...
    echo "ERROR: Some assertions failed !"
    exit 1
elif [[ "$jvm_closed" == "false" ]]; then
    echo "ERROR: JVM has not closed properly !"
    exit 1
else
    echo "ERROR: Some assertions failed !"
    exit 0
fi
); exit $?
```
chippmann, line 1: Hmm somethings off here... We have many printed carriage returns in this script ( `^M`) . Did you copy it from windows or something?
Maybe this helps?: https://stackoverflow.com/questions/15020883/unix-script-appends-m-at-end-of-each-line

chippmann (changes requested):

harness/tests/run_godot_kotlin_tests.sh
piiertho: yep ! that was the problem :p

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+    runs-on: macos-latest
+    steps:
+      - name: Clone Godot JVM module.
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
```
raniejade, line 72: I think we should make this a matrix property. It will be good if we can test against different JVM versions.

.github/workflows/check-pr-jvm-godot-library.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Godot JVM module.
+        uses: actions/checkout@v2
+      - name: Checkout submodules
+        uses: snickerbockers/submodules-init@v4
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
```
raniejade, line 22: same here, make it a matrix property so we can easily add new JVM targets to test against.

raniejade (changes requested):
A few comments, but overall looking good!

.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
piiertho: Good idea !

CedNaru (approved):

chippmann (approved):

raniejade (approved):

Issue #39, opened 2021-01-08
Incremental build breaks entry file generation


chippmann:
## Problem
@piiertho noticed this bug. Quote from discord:
```
@Cedric I'm facing a strange behaviour of entry generator.
I have two kotlin files in src. If I modify only one and rebuild, only the modified file is added to entry.
For now I have to clean each time
```

## Source
As we use a compiler plugin, in hindsight it's logical why it happens. When an incremental build is happening, only the affected files are recompiled, thus we override the already existing entry file with an incomplete entry file as we only get the changed classes in the entry generator

Sadly the solution is most likely not as trivial as i hoped it would be...

### Disable Incremental build for user module
The simplest solution would be to disable incremental build support for the users module for which we generate the entry file. But of course this is neither wanted nor a feasible solution IMO.

### Generate separate entry files
Another option would be to generate a separate registration file for each class we process and drop the call to those individual registration files into needle comments on a main entry file that is not overriden on each build, rather it's modified:

```kotlin
// individual entry file per class
// overriden on each build

fun registerOtherScript(registry: ClassRegistry) = with(registry) {
      registerClass<OtherScript>(OtherScript::class.qualifiedName!!, "Node", false) {
            //...
      }
}
```

```kotlin
// main entry file per project
// not overriden on each build, but modified

class Entry : Entry() {
  override fun Context.init() {
    registerOtherScript(registry)
    registerAnotherClass(registry)
    /* <--- needle insert registry functions */                     // <- here at this needle the calls for class registrations would be inserted
  }
}
```

The problem with this approach is: how do we handle deleted or renamed classes? For this I do not have a solution yet...

### Rewrite to AnnotationProcessor
Again picking up this topic. It would result in a significant amount of work especially regarding extraction of property default values as we cannot parse the PSI tree that easily in an annotation processor like we can in a compiler plugin.
That said it is possible, but the default value extraction code would most likely be even complexer than it is now making it harder to maintain and extend.

### Write an `AnalysisHandlerExtension`
As far as i understand, an `AnalysisHandlerExtension` runs for all classes even on an incremental build. Maybe we can hook our entry generation in there and use `RetryWithAdditionalRoots` as a result to retrigger the build. This eventually would even allow us to get rid of the dummy target we setup in the gradle plugin. But i need to investigate that more.

## Issue placement
So why is this issue placed in this repo if it "only" touches the entry generator?
[...]

PR #40, opened 2021-01-11
Feature/first script inheritance

piiertho:
I make this PR as a draft as this needs entry generation debug of #39 and modifications.
Here is sum up of discords discussion about registration design:
If we have
```kotlin
@RegisterClass
open class ClassInheritanceParent : Node() {

    //---------------- Here to check ------------------

    @RegisterProperty
    var closedFunctionHasBeenCalled = false

    @RegisterProperty
    var parentOpenFunctionHasBeenCalled = false

    //-------------------------------------------------

    @RegisterProperty
    var closedVar = 0

    @RegisterProperty
    open var openVar = 0

    @RegisterFunction
    fun closedFunction() {
        closedFunctionHasBeenCalled = true
    }

    @RegisterFunction
    open fun openFunction() {
        parentOpenFunctionHasBeenCalled = true
    }
}
```
And:
```kotlin
@RegisterClass
class ClassInheritanceChild : ClassInheritanceParent() {

    //---------------- Here to check ------------------

    @RegisterProperty
    var childOpenFunctionHasBeenCalled = false

    //-------------------------------------------------

    override var openVar: Int = 100

    override fun openFunction() {
        childOpenFunctionHasBeenCalled = true
    }
}
```

Overriden values should be automatically registered, but intellij plugin would show a warning and provide a quick fix to add annotation, for code readability.

This is currently the first version of script inheritance, you cannot inherit from abstract classes for now (Don't know if we should be able).
Instead of searching methods in inherited classes, we provide inheritance by registering methods and properties.
This duplicates registered methods, but avoid CPU time.

Discussions are open !

### DIFF
--- a/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
@@ -8,7 +8,8 @@ class KtClass<T : KtObject>(
         private val constructors: Map<Int, KtConstructor<T>>,
         private val _properties: Map<String, KtProperty<T, *>>,
         private val _functions: Map<String, KtFunction<T, *>>,
-        private val _signalInfos: Map<String, KtSignalInfo>
+        private val _signalInfos: Map<String, KtSignalInfo>,
+        val baseGodotClass: String
 ) {
     val functions: Array<KtFunction<T, *>>
         get() = _functions.values.toTypedArray()
--- a/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
@@ -2,6 +2,7 @@ package godot.core
 
 import godot.util.VoidPtr
 
+@Suppress("unused")
 class KtClass<T : KtObject>(
         val name: String,
         val superClass: String,
--- a/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
@@ -666,7 +666,7 @@ class ClassBuilderDsl<T : KtObject>(
 class ClassRegistry {
     val classes = mutableListOf<KtClass<*>>()
 
-    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, baseGodotClass: String = superClass, cb: ClassBuilderDsl<T>.() -> Unit) {
+    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, baseGodotClass: String, cb: ClassBuilderDsl<T>.() -> Unit) {
         val builder = ClassBuilderDsl<T>(name, superClass, baseGodotClass)
         builder.cb()
         TypeManager.registerUserType(name)
--- a/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
@@ -31,7 +31,8 @@ data class KtFunctionArgument(
 
 class ClassBuilderDsl<T : KtObject>(
     @PublishedApi internal val name: String,
-    private val superClass: String
+    private val superClass: String,
+    private val baseGodotClass: String
 ) {
     private val constructors = mutableMapOf<Int, KtConstructor<T>>()
 
@@ -642,7 +643,7 @@ class ClassBuilderDsl<T : KtObject>(
 
     internal fun build(): KtClass<T> {
         check(constructors.isNotEmpty()) { "Please provide at least one constructor." }
-        return KtClass(name, superClass, constructors, properties, functions, signals)
+        return KtClass(name, superClass, constructors, properties, functions, signals, baseGodotClass)
     }
 
     @PublishedApi
@@ -665,8 +666,8 @@ class ClassBuilderDsl<T : KtObject>(
 class ClassRegistry {
     val classes = mutableListOf<KtClass<*>>()
 
-    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, cb: ClassBuilderDsl<T>.() -> Unit) {
-        val builder = ClassBuilderDsl<T>(name, superClass)
+    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, baseGodotClass: String = superClass, cb: ClassBuilderDsl<T>.() -> Unit) {
+        val builder = ClassBuilderDsl<T>(name, superClass, baseGodotClass)
         builder.cb()
         TypeManager.registerUserType(name)
         registerClass(builder.build())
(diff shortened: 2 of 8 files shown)


kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -31,7 +31,8 @@ data class KtFunctionArgument(
 
 class ClassBuilderDsl<T : KtObject>(
     @PublishedApi internal val name: String,
-    private val superClass: String
+    private val superClass: String,
```
piiertho, line 34: Don't know if we should keep it. It is used only for logging.

chippmann: I would remove it if it's not needed.
The entry generation really is only working for kotlin, so if one wants to use another language, he has to write the entry file manually.  So the lesser one has to write, the better IMO

chippmann:
I added the changes in the entry generator and updated the ref in this PR. See: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/9

As a result this PR  is dependent on https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/9

chippmann:
@piiertho IMO you can remove the draft thingy and we can review it (depending on what you want to do with `private val superClass: String`).

kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
```diff
@@ -8,7 +8,8 @@ class KtClass<T : KtObject>(
         private val constructors: Map<Int, KtConstructor<T>>,
         private val _properties: Map<String, KtProperty<T, *>>,
         private val _functions: Map<String, KtFunction<T, *>>,
-        private val _signalInfos: Map<String, KtSignalInfo>
+        private val _signalInfos: Map<String, KtSignalInfo>,
+        val baseGodotClass: String
```
chippmann, line 12: I think we should add `@Suppress("UNUSED_VARIABLE") ` to each variable accessed via jni from cpp.

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
chippmann: Should we now leave it or remove it?

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -665,8 +666,8 @@ class ClassBuilderDsl<T : KtObject>(
 class ClassRegistry {
     val classes = mutableListOf<KtClass<*>>()
 
-    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, cb: ClassBuilderDsl<T>.() -> Unit) {
-        val builder = ClassBuilderDsl<T>(name, superClass)
+    fun <T : KtObject> registerClass(name: String, superClass: String, isTool: Boolean = false, baseGodotClass: String = superClass, cb: ClassBuilderDsl<T>.() -> Unit) {
```
chippmann, line 669: I wouldn't set the default value to the superclass. we cannot check that. better to be explicit in this case imo.

src/kt_class.cpp
```diff
@@ -8,6 +8,7 @@ KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) :
     jni::Env env { jni::Jvm::current_env() };
     name = get_name(env);
     super_class = get_super_class(env);
+    base_godot_class = get_base_godot_class(env);
```
chippmann, line 11: I'm just thinking: couldn't we combine those calls into one? because as of now we make three subsequent calls from cpp to jvm for each class right?

chippmann (changes requested):

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: I'll remove it today :)

chippmann: But the entry generator will also need updating then.
Do you do it or should i?

src/kt_class.cpp
raniejade: This happens only once per class during registration, probably not worth optimizing.

src/kotlin_script.cpp
```diff
@@ -128,15 +128,15 @@ KtClass* KotlinScript::get_kotlin_class() const {
 Variant KotlinScript::_new(const Variant **p_args, int p_argcount, Variant::CallError &r_error) {
     r_error.error = Variant::CallError::CALL_OK;
 
-    Object *owner = ClassDB::instance(get_kotlin_class()->super_class);
+    Object* owner{ClassDB::instance(get_kotlin_class()->base_godot_class)};
```
raniejade, line 131: From what I understand, we just need to pass in the user defined inherited class?

raniejade: Also, does the order of registration matter? i.e can `ClassInheritanceChild` be registered before `ClassInheritanceParent`?

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: you can do entry gen, I'll adapt cpp code after if you want :)

src/kotlin_script.cpp
piiertho: Nop, thing is with registration trick, godot does not even have to know the user defined super class.  
All inherited methods are also registered in children class.  
This makes more functions registered by classes, but avoid cpu time to retrieve methods throught inheritance.  
With this, registration order does not matter at all. Godot just have to know the base godot class.  
Basically we take advantage of JVM vtable.

src/kt_class.cpp
piiertho: This can be optimized later if we see we have long engine boot times.

chippmann: ok for me

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: To sum up discord discussion, we'll keep `KtClass::super_class` for now and see what we do with `KotlinScript::get_base_script`.

kt/godot-runtime/src/main/kotlin/godot/core/KtClass.kt
chippmann: I know it's more work but i my idea was to just add it to the variables that are used from cpp. Not just everything. The reason being: to be able to see in the ide if it really is used or not and kinda as documentation as what variables are accessed through jni.
If you don't agree I'm also fine with just leaving it like this.

chippmann (approved):

CedNaru (approved):

raniejade (approved):

PR #44, opened 2021-01-17
Feature/handling singletons

piiertho:
This adds a method to retrieve singleton pointers from engine.
This adds private to `KtObject` field `isSingleton` to avoid singleton registering in GC.

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/tests.kt
+++ b/kt/godot-library/src/main/kotlin/godot/tests.kt
@@ -6,9 +6,9 @@ import godot.util.VoidPtr
 import godot.util.camelToSnakeCase
 import kotlin.reflect.KCallable
 
-open class Object(isRef: Boolean = false) : KtObject(isRef) {
+open class Object internal constructor(isRef: Boolean = false, isSingleton: Boolean = false) : KtObject(isRef, isSingleton) {
 
-    constructor() : this(false)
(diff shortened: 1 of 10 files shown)


chippmann (approved):

kt/godot-library/src/main/kotlin/godot/tests.kt
```diff
@@ -6,9 +6,9 @@ import godot.util.VoidPtr
 import godot.util.camelToSnakeCase
 import kotlin.reflect.KCallable
 
-open class Object(isRef: Boolean = false) : KtObject(isRef) {
+open class Object internal constructor(isRef: Boolean = false, isSingleton: Boolean = false) : KtObject(isRef, isSingleton) {
```
raniejade, line 9: `isSingleton` (and `isRef`) is a property of the class not the instance, so it is quite weird to have it as instance fields. Is there a way we can use the types for this checks?

piiertho: No way to use the type and get performances :/
I think I'll wait for merge of other PR and implement the check with internal method with unifying of godot-library and godot-runtime modules

CedNaru (approved):

raniejade (approved):
LGTM, if you add a todo to rework it later.

chippmann (approved):

PR #55, opened 2021-01-25
Call kotlin script by name

chippmann:
This implements the ability to call kotlin scripts by name from GDScript.

By default, kotlin classes are registered with their fully qualified names where `.` are replaced by `_` as GDScript cannot handle `.` in type names.

Example:
`godot.tests.Invocation` will be registered as `godot_tests_Invocation`.
As this can get very unreadable and long to type in GDScript, one can now register a class with a custom name:
```kotlin
package very.long.package.name.hehe

@RegisterClass(customClassName = "VeryCoolClassName")
class MyRealNameDoesntMatterAnymore: Node
```
This can then be used as `VeryCoolClassName` in GDScript instead of `very_long_package_name_hehe_MyRealNameDoesntMatterAnymore`.

Other examples can be found in the unit tests where the Invocation script is now called by name rather than with the `load` function.

Of course this would lead to errors if the same class name is registered twice. Hence https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/11 implements checks in the EntryGenerator which lets a compilation fail if two or more classes are registered with the same name.

To provide some hints and tools before compilation, the IDE plugin now also implements those checks and provides errors in the editor and a `GoTo` option to quickly navigate between the classes which are registered with the same name:
screenshot1
screenshot2
screenshot5

## OtherFixes:
- Use jetbrains `Annotator` classes for code inspection rather than `AbstractKotlinInspection`. This change allows the IDE to show errors on file level in the project panel.
- Use bundle messages rather than hardcoded strings for ui texts and id's

This PR depends on https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/11


src/kotlin_language.cpp
```diff
@@ -377,9 +379,12 @@ void KotlinLanguage::frame() {
 }
 
 bool KotlinLanguage::handles_global_class_type(const String& p_type) const {
-    return ScriptLanguage::handles_global_class_type(p_type);
+    return p_type == "KotlinScript";
 }
 
 String KotlinLanguage::get_global_class_name(const String& p_path, String* r_base_type, String* r_icon_path) const {
-    return ScriptLanguage::get_global_class_name(p_path, r_base_type, r_icon_path);
+    if (p_path.begins_with(GODOT_ENTRY_PATH)) {
```
piiertho, line 386: Implementing this methods make engine slow, we need to investigate this.  
Note to myself: If only used from editor, not include it in export.

kt/godot-library/src/main/kotlin/godot/annotation/RegisterClass.kt
```diff
@@ -4,7 +4,8 @@ package godot.annotation
  * Registeres a class in godot so it can be used from another language or from godot
  *
  * @param isTool Same as the `tool` keyword for GDScript. Makes the class usable in the editor
+ * @param customClassName Register the class name with this string rather than with `fully_qualified_ClassName`. **Note:** If this class name is already registered, a compilation error will be thrown! So make sure your custom class name is unique!
  */
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
-annotation class RegisterClass(val isTool: Boolean = false)
+annotation class RegisterClass(val isTool: Boolean = false, val customClassName: String = "")
```
raniejade, line 11: Why not use `className` instead? The custom prefix is a bit redundant here. Also, you might want to move it as the first parameter so users can just do: `@RegisterClass("OtherClassName")`.

raniejade (changes requested):

src/kotlin_language.cpp
chippmann: Do you want to have a look in this PR or do you want to tackle that in a separate PR?

piiertho: I think we should have a look here, as this introduce an unexpected behaviour

piiertho: not anymore slow engine closing, don't know what happened.

piiertho (approved):

raniejade (approved):

CedNaru (approved):

CedNaru (approved):

Issue #47, opened 2021-01-22
Implement Api Generation


chippmann:
Implement the api generation to generate Kotlin wrappers for all Godot classes.
Similar to the K/N project.

chippmann:
Implemented by #57

PR #57, opened 2021-01-31
Feature/api generation

piiertho:
This adds API generation as gradle plugin.  
This also fixes method returns of type reference ref count.
This solves #47


chippmann:
To document some debugging and discussions from discod:
With this current impl most likly `496` methods are not working.
Prime example `ResourceLoader::load`

The reason being: they are not in the `ClassDB` which we currently use in `gd_kotlin.cpp` to get the `MethodBind` of engine methods. Hence the methods cannot be found and called during runtime.
All those methoda have in common that they are `virtual` functions.

But we cannot just ditch them as we need them to work. C# is generating it's own `api.json`. Most likely because of this reason.
We need to investicate the C# code to see how they have managed to make functions like `ResourceLoader::load` work.

chippmann:
Some further investigations on the cSharp code lead to this:
```csharp
        private const string nativeName = "_ResourceLoader";

        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
        private static IntPtr method_bind_1 = Object.godot_icall_Object_ClassDB_get_method(nativeName, "load");

        [GodotMethod("load")]
        public static Resource Load(string path, string typeHint = "", bool noCache = false)
        {
            return NativeCalls.godot_icall_3_1107(method_bind_1, ptr, path, typeHint, noCache);
        }
```

The class name in the ClassDB seems to be `_ResourceLoader` and not `ResourceLoader`.

The following example also supports this:
```gdscript
func _ready():
    print(ClassDB.class_has_method("_ResourceLoader", "load"))
    print(ClassDB.class_has_method("ResourceLoader", "load"))
    print(ClassDB.class_has_method("Node", "get_children"))
```
Prints:
`True`
`False`
`True`

kt/godot-library/build.gradle.kts
```diff
@@ -1,5 +1,18 @@
...
+apply<godot.gradle.ApiGeneratorPlugin>()
+
+configure<godot.gradle.ApiGeneratorPluginExtension> {
+    outputDir.set(project.file("$rootDir/godot-library/src/main/kotlin/godot/gen"))
+    sourceJson.set(project.file("$rootDir/api-generator/src/main/resources/api.json"))
+    isNative = false
+}
+
+tasks.compileKotlin {
+    dependsOn(tasks.generateAPI)
```
chippmann, line 15: Setting the comment here as the api-generator plugin has no PR:
I would put the task `generateApi` into the group `godot-jvm` ike we do in the main gradle plugin

chippmann (changes requested):

chippmann (approved):

CedNaru (approved):

raniejade (approved):

PR #58, opened 2021-02-01
Implement support for custom srcDirs

chippmann:
This implements the support to define other source dirs than `src/main/kotlin` and write and use scripts from them.
The source dirs are now also written to the entry file and retreived by cpp when searching a class (similar like before but before, the source dir was hardcoded in cpp)

Entry file:
```kotlin
class Entry : Entry() {
  override fun Context.init() {
    TestScriptRegistrar().register(registry)
    InvocationRegistrar().register(registry)
    ClassInheritanceChildRegistrar().register(registry)
    ClassInheritanceParentRegistrar().register(registry)
    OtherScriptRegistrar().register(registry)
  }

  override fun Context.initEngineTypes() {
    registerVariantMapping()
    registerEngineTypes()
    registerEngineTypeMethods()
  }

  override fun provideSrcDirs() = listOf<String>("src/main/kotlin/", "src/main/java/", "scripts/")
}
```

Please look carefully at the cpp code. As you know i'm by no means a cpp developer.

Depends on: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/12

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -64,6 +64,20 @@ void load_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes) {
     classes.delete_local_ref(env);
 }
 
+void provide_src_dirs_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_src_dirs) {
+    jni::Env env(p_env);
+    jni::JObjectArray src_dirs_raw{jni::JObjectArray(p_src_dirs)};
+    Vector<String> src_dirs{};
+    for (int i = 0; i < src_dirs_raw.length(env); ++i) {
+        String src_dir = env.from_jstring(src_dirs_raw.get(env, i));
+        src_dirs.push_back(src_dir);
+    }
+    GDKotlin::get_instance().scripts_root = src_dirs;
+    jni::JObject j_object{p_this};
+    j_object.delete_local_ref(env);
+    src_dirs_raw.delete_local_ref(env);
+}
+
 void unload_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes) {
     jni::Env env(p_env);
     jni::JObjectArray classes{jni::JObjectArray(p_classes)};
@@ -201,7 +215,6 @@ void GDKotlin::init() {
     jni::Jvm::init(args);
     print_line("Starting JVM ...");
     auto project_settings = ProjectSettings::get_singleton();
-    scripts_root = "res://src/main/kotlin/";
     String bootstrap_jar = project_settings->globalize_path("res://build/libs/godot-bootstrap.jar");
     print_line(vformat("Loading bootstrap jar: %s", bootstrap_jar));
     jni::Env env{jni::Jvm::current_env()};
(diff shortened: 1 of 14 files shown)


raniejade (approved):

src/gd_kotlin.cpp
```diff
@@ -64,6 +64,20 @@ void load_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes) {
     classes.delete_local_ref(env);
 }
 
+void register_src_dirs_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_src_dirs) {
+    jni::Env env(p_env);
+    jni::JObjectArray src_dirs_raw{jni::JObjectArray(p_src_dirs)};
+    Vector<String> src_dirs{};
```
piiertho, line 70: This can be written this way to get code easier to read:
```cpp
Vector<String> src_dirs;
```
This will call default constructor.

src/gd_kotlin.cpp
```diff
@@ -64,6 +64,20 @@ void load_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes) {
...
+    for (int i = 0; i < src_dirs_raw.length(env); ++i) {
+        String src_dir = env.from_jstring(src_dirs_raw.get(env, i));
```
piiertho, line 72: cpp 11 init ?

src/gd_kotlin.cpp
```diff
@@ -64,6 +64,20 @@ void load_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes) {
...
+        src_dirs.push_back(src_dir);
+    }
+    GDKotlin::get_instance().scripts_root = src_dirs;
+    jni::JObject j_object{p_this};
+    j_object.delete_local_ref(env);
```
piiertho, line 77: IIRC it is a JVM to CPP call so there is no need to delete_local_ref.
Or am I missing something ?

src/gd_kotlin.cpp
```diff
@@ -330,7 +344,12 @@ void GDKotlin::unregister_classes(jni::Env& p_env, jni::JObjectArray p_classes)
 }
 
 KtClass* GDKotlin::find_class(const String& p_script_path) {
-    StringName class_name = p_script_path.trim_prefix(scripts_root).trim_suffix(".kt").replace("/", ".");
+    String class_name_as_string = p_script_path.trim_prefix("res://").trim_suffix(".kt");
+    for (int i = 0; i < scripts_root.size(); ++i) {
+        String script_root = scripts_root[i].strip_edges();
```
piiertho, line 349: cpp 11 init

src/gd_kotlin.cpp
```diff
@@ -330,7 +344,12 @@ void GDKotlin::unregister_classes(jni::Env& p_env, jni::JObjectArray p_classes)
 }
 
 KtClass* GDKotlin::find_class(const String& p_script_path) {
-    StringName class_name = p_script_path.trim_prefix(scripts_root).trim_suffix(".kt").replace("/", ".");
+    String class_name_as_string = p_script_path.trim_prefix("res://").trim_suffix(".kt");
```
piiertho, line 347: This should be merged after `KtClasse` are cached in `KotlinScript`.
If we cannot cache them we'll have to find another way to retreive `KtClass`
Those lines will cost too much.

src/gd_kotlin.cpp
```diff
@@ -330,7 +344,12 @@ void GDKotlin::unregister_classes(jni::Env& p_env, jni::JObjectArray p_classes)
...
+    for (int i = 0; i < scripts_root.size(); ++i) {
+        String script_root = scripts_root[i].strip_edges();
+        class_name_as_string = class_name_as_string.trim_prefix(script_root);
+    }
+    StringName class_name = class_name_as_string.replace("/", ".");
```
piiertho, line 352: cpp 11 init

piiertho (changes requested):

src/gd_kotlin.cpp
chippmann: Indeed. I'll fix the rest as well once we have the caching merged

CedNaru (approved):

chippmann:
Replaced by https://github.com/utopia-rise/godot-jvm/pull/75

PR #63, opened 2021-02-09
Support nullables and fix reloading

### LINKED ISSUE #48: Bugfix: Start editor from anywhere
Currently the editor can only be started from within a godot project that was already built.
The reason being that the `bootstrap.jar` cannot be found: `<pwd>/build/libs/godot-bootstrap.jar`

### LINKED ISSUE #49: Implement nullable support for registered members
Currently no nullable types can be registered. Neither as function params or return types, nor as property types.

### LINKED ISSUE #53: Bugfix: Fix editor crash on project rebuild
Currently the engine crashes when the loaded project is rebuilt.
```
Done registering managed engine types...
Loading resource: res://src/main/kotlin/godot/tests/Invocation.kt
ERROR: set_path: Another resource is loaded from path 'res://src/main/kotlin/godot/tests/Invocation.kt' (possible cyclic resource inclusion).
   At: core/resource.cpp:82.
ERROR: find_class: Failed to find class godot.tests.Invocation for path: res://src/main/kotlin/godot/tests/Invocation.kt
   At: modules/kotlin_jvm/src/gd_kotlin.cpp:325.
handle_crash: Program crashed with signal 11
```

TBD: add more stacktrace (not done yet as first testing is very inconsistent)

chippmann:
This resolves #48, resolves #49 and resolves #53. 

This PR implements the support for nullables for registered members.
It also fixes editor crashes on project rebuilds while the editor is open and it allows to run the editor from anywhere.

It introduces one important change:
The bootstrap jar will no longer be loaded from the users build directory. It will be shipped together with the engine. Thus the bootstrap jar is loaded from the same directory the engine executable is in.
This fixes class loader problems when we try to reload the project and it also ensures version compatibility between the bootsrap jar and the engine.

To eliminate the need of manually copying the bootstrap jar for local development after yeach change top it, a scons command was added that builds the bootstrap jar and copies it whenever the engine is built. And for when it is not needed to rebuild the engine but only the bootstrap jar, a gradle task was added that automatically copies the new bootstrap jar to the engine bin dir whenever the bootstrap jar is built.

Also all log messages from our module are now prefixed with `Godot-Jvm: ` to match the way the mono module is doing it and to make it clear for the user from where a log message is coming from.
For future log needs, macros where added that prefix your messages for you, also for the kotlin side, logging functions were added.


piiertho (changes requested):
I'm havving errors on signals during tests:
```
* test_nullables_are_correctly_set_from_function_lika_a_ready_call
Godot-JVM: Try to create godot.tests.Invocation instance.
Hello Invocation!
Godot-JVM: Instantiated an object of type godot.tests.Invocation
Enter tree !
Instance id: 2175
CustomName is Idonthaveanyidea
Name is: @@108
Name is: TestName
ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'no_param' to method 'Node.hook_no_param'.
   At: core\object.cpp:1503
ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'one_param' to method 'Node.hook_one_param'.
   At: core\object.cpp:1503
ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'two_param' to method 'Node.hook_two_param'.
   At: core\object.cpp:1503
```

src/gd_kotlin.cpp
```diff
@@ -74,16 +76,16 @@ void unload_classes_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_classes)
 }
 
 void register_engine_types_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_engine_types, jobjectArray p_singleton_names,
-        jobjectArray p_method_names, jobjectArray p_types_of_methods) {
-    print_verbose("Starting to register managed engine types...");
+                                jobjectArray p_method_names, jobjectArray p_types_of_methods) {
+    LOG_VERBOSE("Starting to register managed engine types...")
```
piiertho, line 80: Maybe we can add verbose logs only for DEBUG builds, so that we don't get extra calls in release ?

chippmann:
> I'm havving errors on signals during tests:
> 
> ```
> * test_nullables_are_correctly_set_from_function_lika_a_ready_call
> Godot-JVM: Try to create godot.tests.Invocation instance.
> Hello Invocation!
> Godot-JVM: Instantiated an object of type godot.tests.Invocation
> Enter tree !
> Instance id: 2175
> CustomName is Idonthaveanyidea
> Name is: @@108
> Name is: TestName
> ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'no_param' to method 'Node.hook_no_param'.
>    At: core\object.cpp:1503
> ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'one_param' to method 'Node.hook_one_param'.
>    At: core\object.cpp:1503
> ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'two_param' to method 'Node.hook_two_param'.
>    At: core\object.cpp:1503
> ```

@piiertho 
I debugged this a bit and it seems we have a bug in the signal registration. No user signals are registered at runtime it seems. I think we just never noticed it the tests because this test that i added is the first one to add an invocation scene to the active scene tree and is the only one the `_ready` function that emits the signals is actually called.
If you are ok with it i would rather fix this in a separate PR as this bug seems to have nothing to do with this PR.

piiertho:
> > I'm havving errors on signals during tests:
> > ```
> > * test_nullables_are_correctly_set_from_function_lika_a_ready_call
> > Godot-JVM: Try to create godot.tests.Invocation instance.
> > Hello Invocation!
> > Godot-JVM: Instantiated an object of type godot.tests.Invocation
> > Enter tree !
> > Instance id: 2175
> > CustomName is Idonthaveanyidea
> > Name is: @@108
> > Name is: TestName
> > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'no_param' to method 'Node.hook_no_param'.
> >    At: core\object.cpp:1503
> > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'one_param' to method 'Node.hook_one_param'.
> >    At: core\object.cpp:1503
> > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'two_param' to method 'Node.hook_two_param'.
> >    At: core\object.cpp:1503
> > ```
> 
> @piiertho
> I debugged this a bit and it seems we have a bug in the signal registration. No user signals are registered at runtime it seems. I think we just never noticed it the tests because this test that i added is the first one to add an invocation scene to the active scene tree and is the only one the `_ready` function that emits the signals is actually called.
> If you are ok with it i would rather fix this in a separate PR as this bug seems to have nothing to do with this PR.

Do we have same bug when running the scene from editor ?

chippmann:
> > > I'm havving errors on signals during tests:
> > > ```
> > > * test_nullables_are_correctly_set_from_function_lika_a_ready_call
> > > Godot-JVM: Try to create godot.tests.Invocation instance.
> > > Hello Invocation!
> > > Godot-JVM: Instantiated an object of type godot.tests.Invocation
> > > Enter tree !
> > > Instance id: 2175
> > > CustomName is Idonthaveanyidea
> > > Name is: @@108
> > > Name is: TestName
> > > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'no_param' to method 'Node.hook_no_param'.
> > >    At: core\object.cpp:1503
> > > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'one_param' to method 'Node.hook_one_param'.
> > >    At: core\object.cpp:1503
> > > ERROR: In Object of type 'Spatial': Attempt to connect nonexistent signal 'two_param' to method 'Node.hook_two_param'.
> > >    At: core\object.cpp:1503
> > > ```
> > 
> > 
> > @piiertho
> > I debugged this a bit and it seems we have a bug in the signal registration. No user signals are registered at runtime it seems. I think we just never noticed it the tests because this test that i added is the first one to add an invocation scene to the active scene tree and is the only one the `_ready` function that emits the signals is actually called.
> > If you are ok with it i would rather fix this in a separate PR as this bug seems to have nothing to do with this PR.
> 
> Do we have same bug when running the scene from editor ?

Yes but only for the gut tests. When i run the invocation scene standalone (from editor or command line)  the signal registration works.
Hmm maybe i really did something wrong in the tests.
Does godot have a limitation on what one can add to the root node of the active scene? `get_tree().root.add_child(spatial)`

harness/tests/src/main/kotlin/godot/tests/inheritance/ClassInheritanceParent.kt
```diff
@@ -1,44 +1,44 @@
...
-    fun closedFunction() {
-        closedFunctionHasBeenCalled = true
-    }
-
-    @RegisterFunction
-    open fun openFunction() {
-        parentOpenFunctionHasBeenCalled = true
-    }
-}
+package godot.tests.inheritance
```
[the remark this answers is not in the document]
chippmann: yeah it had tab  based indentation

src/gd_kotlin.cpp
chippmann: done

chippmann:
Ok i debugged the signal issue and discussed it with @piiertho on discord:
The problem currently only happens when instancing our scripts with:
`var invocation_script = godot_tests_Invocation.new()`
when instancing a scene that uses our script like:
`var invocation_script = load("res://Spatial.tscn").instance()`
then everything works fine.

The problem seems to be that we're not setting the script on the resulting node in the first example.
@piiertho will fix this in another branch and this a different PR as it has nothing to do with this PR and we have to think a bit more about the cases where such scripts are instanced from the JVM side.

piiertho (approved):

CedNaru (approved):

PR #66, opened 2021-02-09
Add GD singleton and extensions

chippmann:
This adds the GD singleton and extensions @CedNaru implemented for the `godot-kotlin` project.

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/global/GD.kt
+++ b/kt/godot-library/src/main/kotlin/godot/global/GD.kt
@@ -5,6 +5,7 @@ import godot.RandomNumberGenerator
 import godot.Resource
 import godot.ResourceLoader
 import godot.core.Dictionary
+import godot.core.GarbageCollector
 import godot.core.NodePath
 import godot.core.PoolByteArray
 import godot.core.PoolColorArray
@@ -27,7 +28,7 @@ object GD : GDMath, GDCore, GDRandom, GDPrint {
 
     /** Returns whether instance is a valid object (e.g. has not been deleted from memory).*/
     fun isInstanceValid(instance: Object): Boolean {
-        return instance.rawPtr != nullptr
+        return instance.rawPtr != nullptr && GarbageCollector.isInstanceValid(instance)
     }
 
     /** Returns length of Variant var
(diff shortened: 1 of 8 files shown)


kt/godot-library/src/main/kotlin/godot/global/GD.kt
```diff
@@ -0,0 +1,68 @@
...
+    override val rng = RandomNumberGenerator()
+
+    /** Asserts that the condition is true.
+    If the condition is false, an error is generated and the program is halted until you resume it.
+    Only executes in debug builds. Use it for debugging purposes, to make sure a statement is true during development. */
+    fun assert(condition: Boolean, message: String = "") = assert(condition) { message }
+
+    /** Returns whether instance is a valid object (e.g. has not been deleted from memory).*/
+    fun isInstanceValid(instance: Object): Boolean {
+        return instance.rawPtr != nullptr
```
chippmann, line 30: Here I'm unsure if this really is enough. If the pointer on the cpp side is gone this value is not updated as far as i can see.
Maybe some of you cpp devs have an idea how we can properly implement that?

piiertho (approved):

kt/godot-library/src/main/kotlin/godot/global/GD.kt
CedNaru: For this one, you should probably use this function from the GarbageCollector:
https://github.com/utopia-rise/godot-jvm/blob/bf09d77505502fedacdd2ec0b3290d2adeb23509/kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt#L188

Just checking if the ptr is null is not enough. Freeing the memory doesn't make the pointers to its location null.

piiertho: Good catch

piiertho (changes requested):
Need fix of @CedNaru catch

CedNaru (changes requested):
Not sure if he can actually use it because we need the id.

kt/godot-library/src/main/kotlin/godot/global/GD.kt
chippmann: Added the suggested check. The `rawPointer` we store is the right one to use for the `checkInstance` check i assume?

CedNaru (approved):

piiertho (approved):

PR #71, opened 2021-02-17
Bunnymark

chippmann:
This adds the bunnymark to the samples with some scores.
I included the `.import` files so one does not have to open the benchmarks in the editor first to be able to run them from commandline.


src/kt_variant.h
```diff
@@ -236,12 +236,12 @@ namespace ktvariant {
         //TODO: Manage 32 bits systems.
         des->increment_position(encode_uint64(reinterpret_cast<uintptr_t>(ptr), des->get_cursor()));
 
-        String class_name {ptr->get_class()};
+        StringName class_name{ptr->get_class()};
```
chippmann, line 239: @piiertho Is this still needed?

src/kt_variant.h
```diff
@@ -236,12 +236,12 @@ namespace ktvariant {
...
 
         if (!TypeManager::get_instance().JAVA_ENGINE_TYPES_CONSTRUCTORS.has(class_name)) {
             class_name = ClassDB::get_parent_class(class_name);
 
-            while (class_name.empty()) {
+            while (class_name == String()) {
```
chippmann, line 244: @piiertho Is this still needed?

src/transfer_context.cpp
```diff
@@ -141,16 +141,11 @@ void TransferContext::icall(JNIEnv* rawEnv, jobject instance, jlong jPtr, jint p
 #endif
 
     write_return_value(buffer, ret_value);
-    jni::JObject local_ref{instance};
-    local_ref.delete_local_ref(env);
```
chippmann, line 145: @piiertho could you also look closely here and the other removals in this file?
Is this change needed or irrelevant after the rebase?

src/kt_variant.h
piiertho: Yep, it is an optimization. We were creating the `StringName` from the `String` twice.

src/kt_variant.h
piiertho: As we now directly use a `StringName` yes.

src/transfer_context.cpp
piiertho: All removals are optimization. There is no need to clean local ref as we're in jvm -> cpp case.

piiertho (approved):

CedNaru (approved):

PR #74, opened 2021-02-22
Support nullables as function params and return types

chippmann:
We forgot to support nullables as function params and return types. This implements it together with some tests that test this.

Depends on: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/18

### DIFF
--- a/kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
+++ b/kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
@@ -58,7 +58,7 @@ abstract class KtFunction<T : KtObject, R : Any?>(
         val functionInfo: KtFunctionInfo,
         val parameterCount: Int,
         val variantType: VariantType,
-        private vararg val parameterTypes: VariantType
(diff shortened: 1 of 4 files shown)


CedNaru (approved):

kt/godot-runtime/src/main/kotlin/godot/core/Functions.kt
```diff
@@ -58,7 +58,7 @@ abstract class KtFunction<T : KtObject, R : Any?>(
         val functionInfo: KtFunctionInfo,
         val parameterCount: Int,
         val variantType: VariantType,
-        private vararg val parameterTypes: VariantType
+        private vararg val parameterTypes: Pair<VariantType, Boolean>
```
piiertho, line 61: I would rather use it only as a constructor argument to initialize two array of primitives. To avoid an array of complex object in buffer context.

piiertho (changes requested):

piiertho (approved):

Issue #69, opened 2021-02-13
Implement multi args constructors


piiertho:
For now only JVM no args constructors can be called from cpp.  
We'd like to provide a way to call multi-args constructors.  
A first approach would be to allow one constructor per argument count.  
This means if you register this constructor for a class:  
```kotlin
constructor(i: Int, o: Object)
```
You would not be able to register another constructor with two arguments for this class (as c# implem IIRC).

This would need few modifications of engine code in `KotlinScript` and `KtClass`.  
This would need JVM glue code modification on `KtClass`.  
This would also need some registration and idea plugin modifications I think.

For performance purpose, we should use arrays to store constructors and indices should be argument count.

chippmann:
Implemented with #80

PR #80, opened 2021-02-26
Feature/multi args constructor

piiertho:
This solves #69  
This implements multi args constructors, without overloading. This means that registering two constructors with same arguments count is not possible.  
Calling constructor with arguments from `gscript` using registered class name is not possible, `load` method should be called instead.  
`KtConstructor` are now registered and exposed to cpp, as `KtFunction` and `KtProperty`.
This needs https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/20

Edit from @chippmann :
This now also needs https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/21
And it also includes now an additional annotation `RegisterConstructor` with corresponding documentation and IDE plugin checks.
It also adds a previously missing check that always a default constructor is present.


kt/godot-runtime/src/main/kotlin/godot/core/Constructors.kt
```diff
@@ -1,26 +1,121 @@
...
+            paramsArray[i] = TransferContext.readSingleArgument(parameterTypes[i], parameterNullables[i])
+        }
+        TransferContext.buffer.rewind()
+        val instance = invoke()
+        resetParamsArray()
+        instance
+    }
+
+    companion object {
+        val paramsArray by threadLocal { arrayOf<Any?>(null, null, null, null, null) }
```
CedNaru, line 32: I would make use of the constant `CONSTRUCTOR_MAX_ARGS` to init the array so we don't need to modify that part if we change its value.

kt/entry-generation/godot-annotation-processor/src/main/kotlin/godot/annotation/processor/GodotAnnotationProcessor.kt
```diff
@@ -95,11 +95,6 @@ class GodotAnnotationProcessor(
     }
 
     private fun performSanityChecks() {
-        classes.forEach {
```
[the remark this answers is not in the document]
chippmann: This check is in the entry generator: https://github.com/utopia-rise/godot-kotlin-entry-generator/blob/f76b32eb264898bc6bf8805557b582b8bfca90d0/src/main/kotlin/godot/entrygenerator/generator/clazz/JvmClassRegistrationGenerator.kt#L55.
I yet need to combine all sanity checks so they are located in the same place.

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
```diff
@@ -48,6 +48,9 @@ class ClassBuilderDsl<T : KtObject>(
         require(!constructors.containsKey(constructor.parameterCount)) {
             "A constructor with ${constructor.parameterCount} argument(s) already exists."
         }
+        require(constructor.parameterCount <= CONSTRUCTOR_MAX_ARGS) {
+            "Cannot register a constructor with ${constructor.parameterCount} arguments, max argument count is $CONSTRUCTOR_MAX_ARGS"
+        }
```
chippmann, line 53: I don't remember anymore: why do we have a limit of 5 args per constructor?

src/kt_class.h
```diff
@@ -7,6 +7,10 @@
 #include "kt_function.h"
 #include "kt_signal_info.h"
 #include "java_instance_wrapper.h"
+#include "kt_constructor.h"
+
+const int CONSTRUCTOR_MAX_ARGS_SIZE = 5;
+const int MAX_CONSTRUCTOR_SIZE = CONSTRUCTOR_MAX_ARGS_SIZE + 1;
```
chippmann, line 13: Maybe add a comment on where to change the same constant on the jvm so it doesn't get forgotten if changed at either the cpp or the jvm side

chippmann (changes requested):
~~Please also update the docs with the limitation you descriped that one currently can only use the `load` function and that the arg count is limited..~~ done

kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt
piiertho: Constructor call is pretty much the same as method call. IIRC we can have more than 5 params, but it becomes less performant because there is unpacking.

chippmann: Ok I will add that as an explanation to the docs

src/kt_class.h
piiertho: done

chippmann (approved):

raniejade (approved):

PR #82, opened 2021-02-27
Fix editor crashes

### LINKED ISSUE #76: Creating a kotlin script in the editor crashes the editor
Currently the editor crashes when we create a kotlin script in it.
Cause: We added a failing condition for debug builds if a class could not be found:
```cpp
KtClass* GDKotlin::find_class(const StringName& p_script_path) {
#ifdef DEBUG_ENABLED
    //FIXME: this crashes the editor when creating a kotlin script through the godot editor. Cause: this function gets called for the newly created script but of course it's not compiled yet as it was just added now...
    JVM_ERR_FAIL_COND_V_MSG(!classes.has(p_script_path), nullptr,
                   vformat("Failed to find class for path: %s", p_script_path));
#endif
    return classes[p_script_path];
}
```

### LINKED ISSUE #81: Non project kotlin files crashes editor
If you have kotlin files (`kt`extension) which do not belong to the project itself in the project dirs, the editor crashes.
For example if you have a custom gradle plugin inside the main project dir, which is of course not built into the final  `main.jar` the editor crashes as it cannot find those classes in the `main.jar`.
I suspect the issue here is mainly that godot sees the kotlin files and tries to get script specific informations from them as it sees them as script resources.

chippmann:
This resolves #76 and resolves #81

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -393,9 +393,9 @@ void GDKotlin::unregister_classes(jni::Env& p_env, jni::JObjectArray p_classes) 
 
 KtClass* GDKotlin::find_class(const StringName& p_script_path) {
 #ifdef DEBUG_ENABLED
-    //FIXME: https://github.com/utopia-rise/godot-jvm/issues/76
-    JVM_ERR_FAIL_COND_V_MSG(!classes.has(p_script_path), nullptr,
-                   vformat("Failed to find class for path: %s", p_script_path));
+    if (!classes.has(p_script_path)) {
+        return nullptr;
+    }
 #endif
     return classes[p_script_path];
 }
--- a/src/kotlin_language.cpp
+++ b/src/kotlin_language.cpp
@@ -383,11 +383,14 @@ String KotlinLanguage::get_global_class_name(const String& p_path, String* r_bas
     if (p_path.begins_with(GODOT_ENTRY_PATH)) {
         return String();
     }
-    KtClass* clazz{GDKotlin::get_instance().find_class(p_path)};
 
-    if (r_base_type) {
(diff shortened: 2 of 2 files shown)


src/gd_kotlin.cpp
```diff
@@ -393,9 +393,9 @@ void GDKotlin::unregister_classes(jni::Env& p_env, jni::JObjectArray p_classes)
 
 KtClass* GDKotlin::find_class(const StringName& p_script_path) {
 #ifdef DEBUG_ENABLED
-    //FIXME: https://github.com/utopia-rise/godot-jvm/issues/76
-    JVM_ERR_FAIL_COND_V_MSG(!classes.has(p_script_path), nullptr,
-                   vformat("Failed to find class for path: %s", p_script_path));
+    if (!classes.has(p_script_path)) {
```
piiertho, line 396: I would at least make a print verbose here or even a print WARN.

chippmann: Not sure about that. Because this would print that message or warning for each kt class which is not registered in godot.
Which leads to hundret's of prints for bigger projects.

chippmann: This would clutter up the output in such a great way that it renders it unusable (if we use warnings) and mostly unusable for verbose logs in general

piiertho: True, anyway I don't have strong opinion, we'll see if we need it later for debugging.

piiertho (approved):

raniejade (approved):

src/gd_kotlin.cpp
chippmann: :+1:

PR #87, opened 2021-03-01
Add GD print functions

### LINKED ISSUE #65: ApiGeneration: Generate print functions
Currently the api generator does not generate the godot print functions as they are not present in the api.json

chippmann:
This resolves #65 and implements GD print functions. It also improves the error output in the garbage collector by printing the actually leaked objects rather than the weak ref's.

An [issue](https://github.com/utopia-rise/godot-jvm/issues/86) was encountered along the way where we leak a `RandomNumberGenerator` when the `GD` object is used. This is not fixed in this PR as it has nothing to do with this addition.


chippmann:
The tests are now failing because of the issue i created.
Will comment the printing functions for now. Just uncomment them to see that they are working

CedNaru (approved):

src/gd_kotlin.h
```diff
@@ -18,6 +19,7 @@ class GDKotlin {
 
     jni::JObject class_loader;
     Bootstrap* bootstrap;
+    GDPrintGlue* gd_print_glue;
```
piiertho, line 22: I would prefer a Bridge in BridgeManager

src/gd_print_glue.cpp
```diff
@@ -0,0 +1,54 @@
...
+    Vector<jni::JNativeMethod> methods;
+    methods.push_back(print_method);
+    methods.push_back(print_err_method);
+    methods.push_back(print_raw_method);
+    jni::Env env{jni::Jvm::current_env()};
+    j_class.register_natives(env, methods);
+}
+
+void GDPrintGlue::print(JNIEnv *p_env, jobject p_this, jstring p_message) {
+    jni::Env env(p_env);
```
piiertho, line 36: don't send string through jni, use buffer instead

src/gd_print_glue.cpp
```diff
@@ -0,0 +1,54 @@
...
+    print_line(env.from_jstring(jni::JString{p_message}));
+    jni::JObject j_object{p_this};
+    j_object.delete_local_ref(env);
+}
+
+void GDPrintGlue::print_err(JNIEnv *p_env, jobject p_this, jstring p_message) {
+    jni::Env env(p_env);
```
piiertho, line 43: same.

src/gd_print_glue.cpp
```diff
@@ -0,0 +1,54 @@
...
+    print_error(env.from_jstring(jni::JString{p_message}));
+    jni::JObject j_object{p_this};
+    j_object.delete_local_ref(env);
+}
+
+void GDPrintGlue::print_raw(JNIEnv *p_env, jobject p_this, jstring p_message) {
+    jni::Env env(p_env);
```
piiertho, line 50: same.

piiertho (changes requested):

piiertho (approved):
Thx !

PR #89, opened 2021-03-02
Remove root project requirement

chippmann:
This removes the requirement from our gradle plugin to be applied to the root project.
I don't think this is necessary and with this change it enables one to build multiple godotJvm projects in one project (say server and client, or multiple clients, or a plain java or kotlin module and a godotJvm module).


raniejade:
I'm trying to remember why this was introduce - there was definitely a reason for it.

chippmann:
> I'm trying to remember why this was introduce - there was definitely a reason for it.

IIRC it was for kotlin/native as we required some of it's configuration inside the plugin

piiertho (approved):

CedNaru (approved):

raniejade (approved):

chippmann:
As @raniejade correctly pointed out on discord, with this change the signal resolving would not take into account from which module a connection is coming from, resulting in false positives when having multiple modules with `tscn` files in them and they contain the same node structure, signal names and method names in some cases.

Hence i added support for that in the IDE plugin by introducing a `GodotRootsCache` which resolves all checks by godot root. A godot root being a folder which contains  a `project.godot` file. I also added checks for the code insight to only check code if the file is belonging to a godot root.

Because of these changes i would appreciate a review again.

piiertho (approved):

raniejade:
@chippmann I found what I was talking about before https://plugins.jetbrains.com/docs/intellij/facet.html. Also, since we are always working with gradle - we can probably get the gradle project from a given module.

chippmann:
> @chippmann I found what I was talking about before https://plugins.jetbrains.com/docs/intellij/facet.html. Also, since we are always working with gradle - we can probably get the gradle project from a given module.

As discussed on Discord; I cannot use facets this way as intellij modules != gradle modules and we get a intellij module per source set and configuration. Thus the `tscn` files are not in the same module for intellij as the source files. Hence the cache wouldn't work anymore as we retrieve a different cache instance for source files and `tscn` files.

But i did manage to get gradle module data with the gradle plugin as a dependency. With this i could get rid of the `GodotRootsCache` and replace it with some extension functions which do a lookup if a given element is in a godot root. This should be fast enough as i use virtual files which are cached internally and I'm sure the gradle plugin does as well (but i didn't check that).

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/extension/godotRootExt.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.jetbrains.kotlin.idea.core.util.toVirtualFile
+import org.jetbrains.kotlin.idea.framework.isGradleModule
+import org.jetbrains.kotlin.idea.util.projectStructure.getModule
+import org.jetbrains.kotlin.idea.util.projectStructure.module
+import org.jetbrains.plugins.gradle.util.GradleUtil
+import java.io.File
+
+typealias GodotRoot = String
+
+fun Module?.getGodotRoot(): GodotRoot? {
```
raniejade, line 17: 🎉

raniejade (approved):

Issue #50, opened 2021-01-22
Embedd JRE in pkg and accessible through res://


chippmann:
The JRE should be embedded inside the `pkg` file and from there be accessible through the `res://` path. This makes it so that shipping the JRE separate is obsolete.

piiertho:
I'm not sure anymore about this one.
On iOS jvm would be embedded in exe and on desktop it makes sense to have it in a subfolder. On Android this is already here.
I think we should better copy jre folder in Export plugin we're making for embedded jar.

PR #91, opened 2021-03-03
Feature/desktop exports

piiertho:
This implements an export plugin to add mandatory jars to resource export and copy jre side to exe on desktop Platforms.  
Jars are then copied on startup in user path if they are not present or different.

This get rid of #50


chippmann (changes requested):

chippmann (approved):

chippmann:
@piiertho What just came to my mind: It would probably make sense to document that export copies the jars to the user dir on each update or for first game start.
This is quite an important info. Especially for writing an uninstaller.
But I'm fine with adding that in a different PR if you prefer. But i would add it.

CedNaru (approved):

docs/src/doc/user-guide/exporting.md
```diff
@@ -0,0 +1,25 @@
...
+this will do the job for you.
+
+Then you can export your game as usual, your game `jar` will be included in `pck`.  
+On desktop platforms, this also copies the jre folder of your project in the exported game folder.
+
+## Particularities
+
+`godot-bootstrap.jar` and `main.jar` are set into `pck` during the export process. Also, as a real file path is needed to
+handle them, they are copied on the first game version start from `res://` to `user://` (We check if exists and also md5)
+to only update when needed.
```
chippmann, line 25: I would also add something along the lines of: "Don't forget to remove them when you write an uninstaller."

docs/src/doc/user-guide/exporting.md
```diff
@@ -0,0 +1,25 @@
...
+- Windows: `windows_64_release.exe`
+- Linux: `linux_x11_64_release`
+
+And `release_debug` to:
+- Windows: `windows_64_debug.exe`
+- Linux: `linux_x11_64_debug`
+
+On osx use the script mentionned in [this issue](https://github.com/godotengine/godot-docs/issues/3194#issuecomment-588862977),
+this will do the job for you.
+
```
chippmann, line 17: Is this also needed when we ship the engine?

piiertho: yep, that would be cool to rename them during release CI

raniejade (commented):
One question, how will export work on CI? From what I can see you need to click something to do the export.

piiertho:
> One question, how will export work on CI? From what I can see you need to click something to do the export.

Exports works the same as classic godot exports.  
So this is possible to export using `--export` on engine (compiled with tools).

chippmann (approved):

raniejade (approved):

docs/src/doc/user-guide/exporting.md
chippmann: I would add a new issue to do this as it's a quality of life improvement

PR #96, opened 2021-03-05
New interface to fix Singleton leaks

### LINKED ISSUE #86: `RandomNumberGenerator`leaked upon use of `GD` object
As we store a `RandomNumberGenerator` inside the `GD` object, it is in fact "leaked" and will trigger the leaked exception in our GC when closing the application.
In fact the "leakage" is not a problem as the application closes anyways but we need to either not leak it or fix the error message about it.

Leaking object:
```kotlin
object GD : GDMath, GDCore, GDRandom, GDPrint {
    override val rng = RandomNumberGenerator() //<- this one leaks if this object is used

    //...
}
```

CedNaru:
Fix #86 

This adds a new public Interface you can implement to make some cleaning before the JVM shut down.
The singleton will register itself to the GarbageCollector at init time. When the GC is about to clean up everything, the `collect` method will be called.

I had to change the rng property in the GD singleton to nullable because I can't just call `free()` because it's a Reference, I have to set it to null.


kt/godot-library/src/main/kotlin/godot/global/GDRandom.kt
```diff
@@ -4,43 +4,43 @@ import godot.RandomNumberGenerator
 
 internal interface GDRandom {
     /** Global RandomNumberGenerator for all the rand functions, it should have the same behaviour than GdScript.*/
-    val rng: RandomNumberGenerator
+    var rng: RandomNumberGenerator?
```
raniejade, line 7: It might be good to leave a comment why this is nullable.

raniejade (approved):

piiertho (approved):

chippmann (changes requested):
Good work 👍 
Apart from what @raniejade mentioned, please also add documentation about this new interface. As you mentioned on discord, this interface is also usable and somewhat necessary for users. So it makes sense to document that it exists and how to use it properly and for what.
I would add this to the advanced section as a new page but feel free to put it somewhere else

chippmann (approved):

PR #97, opened 2021-03-06
Add generated docs for api

chippmann:
This generates kDoc for the generated api classes.
Depends on https://github.com/utopia-rise/godot-kotlin-api-generator/pull/5


kt/godot-library/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ apiGenerator {
     outputDir.set(project.file("$rootDir/godot-library/src/main/kotlin/godot/gen"))
     sourceJson.set(project.file("$rootDir/api-generator/src/main/resources/api.json"))
     isNative.set(false)
+    docsDir.set(project.file("$rootDir/../../../doc/classes"))
```
chippmann, line 13: Currently some CI runs fail because in some runs we don't check out godot.
I somehow don't want to include the whole godot class reference in our project.
Should i change those failing CI runs to also check out godot and just not build it, or should i really include the class reference files in our project?

raniejade: I actually prefer to just export the docs from Godot (I assume we can export it to xml/json?) and it to the project. Then generating the kotlin docs can be based on that.

raniejade: Not too keen in building godot so that we can generate docs. Alternatively we can just make building the docs optional on some of the CI builds. i.e just build the docs when we are publishing an artifact.

chippmann: > I actually prefer to just export the docs from Godot (I assume we can export it to xml/json?) and it to the project. Then generating the kotlin docs can be based on that.

The docs are already in xml. Thats what I'm parsing. I would just need to copy them to our project.
But i dont like to do that. We would constantly need to update those.

piiertho: I prefer to not include them in project.

chippmann: @raniejade If you look at my commit https://github.com/utopia-rise/godot-kotlin-jvm/pull/97/commits/848bed14281e0ae1776befe13622a2146a131b4b you might see what i mean.

CedNaru (approved):

raniejade (approved):

PR #99, opened 2021-03-07
Small GC refactor

### LINKED ISSUE #72: Bad frametimes caused by our GC
## The problem
Currently we're having some severe frame drops in certain cases from time to time.
The issue can be observed in the `BunnymarkV2` benchmark.

On my machine (Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz, AMD Radeon RX 5700 XT, 64GB DDR4 @ 2133 MT/s) the framedrops happen each ~2sec. At an average frametime of 16ms and a dropped frametime of ~70ms.

Normal frametime after the benchmark settles at ~60FPS:
normal_frametime

Increased frametime for one frame each ~2sec:
frame_drop_frametime

This microstuttering is noticeable by eye.

## The Source
The reason for this is the way our GC is implemented currently and how it works together with the JVM GC.

In the `V2` benchmark a new `VariantArray` is created each frame (inside `_process` with `get_children`).
Our own GC runs each 500ms.
But this does not mean that the JVM GC runs this often.
So we pile up hundrets of `VariantArray`'s and our GC will not clean them as the `WeakReference` for them is not null, as the JVM GC has not run yet.
But once the JVM GC kicks in and deletes them, all of a sudden our GC has to get rid of hundrets of `VariantArray`'s at once, causing a one frame hickup.

We can verify this by forcing the JVM GC to run together with our own GC.
If I do this I have a smaller hickup each 500ms (the interval at which the GC's then do their work) instead of a bigger hickup each ~2s.

The following timeline from hotspot shows this nicely:

Without forced JVM GC -> hickup each ~2s:
screenshot6

With forced JVM GC -> smaller hickup each 500ms:
screenshot5

And if we look at the flamegraph we see that the cause is the deletion of the `VariantArray`'s:
screenshot4

The reason why this only showed up on my machine in this scale and not so much on other machine's (including my laptop) is most likely because of the amount of RAM in the system and the amount i reserve for each JVM process by default (64GB and 6GB respectively) and thus the JVM GC runs less often, hence piling up more `VariantArray`'s to clean up once the JVM GC has done it's work.

## Needed fixes
So one thing that we need to look at is what breaks the escape analysis. Because the `VariantArray`'s are created in the function local scope and thus should be allocated on the stack and thus no GarbageCollection should be necessary.
Probably because we register a WeakRef to the VariantArrays in our GC is enough to break the escape analysis and thus the VariantArrays are actually allocated on the heap rather than on the stack.
We should investigate this and maybe find a solution for it.

A second problem is that we call CPP code from jvm for each object that needs to be deleted rather than batching all rawPointers into one call. Which creates unnecessary pressure on the JVM/CPP interop.

Lastly we should find a good rule on when to start the call to delete those objects so we mitigate the problem of locking the main thread up from our GC thread, and not just blindly run our GC at an fixed interval.

## Additional Data
No forced JVM GC ptrace -> hickup ~2s: [perf.data.zip](https://github.com/utopia-rise/godot-jvm/files/6008309/perf.data.zip)
Forced JVM GC ptrace -> hickup 500ms: [perf.data_1.zip](https://github.com/utopia-rise/godot-jvm/files/6008329/perf.data_1.zip)

CedNaru:
Fix #72 

This PR slightly changes the GC with the following:
-For Reference and NativeCoreType instances, we now use ReferenceQueue. The JVM GC automatically fills this Queue when an instance is freed. We don't need to manually check every entry in the HashMap.
-The rate of the Thread is not fixed anymore and can dynamically change during runtime between 0ms and 2s. The rate is adjusted for each iteration (+- 100ms)
-In order to not release too much memory at once, a maximum number of References can be free every iteration (256 for now). Don't be afraid of having objects added to the queue faster than we can process them. If it comes to this, the rate is quickly going to be set to 0.
-Same for Objects, we check at most 20% of the objects (minimum of 256 as well in case there is a small number of them). Simple for now, but it's going to be replaced at some point by a young/old generation design.
-VariantType has been removed from NativeCoreType and put in NativeCoreWeakReference. This was the only place it was used anyway.

All the numbers are constants in the GC for now. We can easily tweak the numbers if we wish.
Maybe it will be a good idea to be able to configure that when launching the JVM instead of having constants.

image


chippmann (changes requested):
Really nice work!
Few minor comments.

chippmann:
No hickups anymore on my machine as well.
Hotspot looks great as well!

kt/godot-runtime/src/main/kotlin/godot/core/memory/ReferenceWeakReference.kt
```diff
@@ -0,0 +1,13 @@
+package godot.core.memory
+
+import godot.core.KtObject
+import godot.core.NativeCoreType
+import java.lang.ref.ReferenceQueue
+import java.lang.ref.WeakReference
+
+internal class ReferenceWeakReference(
```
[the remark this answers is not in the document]
CedNaru: I don't have use of it as a data class. It's just a child of WeakReference.

piiertho:
Found the problem.
GC is just not started, this explains why this performs really better on cpu :)

raniejade (approved):

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -1,224 +1,280 @@
...
+        nativeCoreTypeMap[rawPtr] = NativeCoreWeakReference(nativeCoreType, nativeReferenceQueue, variantType)
+    }
+
+    fun registerStatic(instance: GodotStatic) {
+        staticInstances.add(instance)
+    }
+
+    fun getObjectInstance(ptr: VoidPtr): KtObject? {
+        val ktObject = wrappedMap[ptr]
+        if (ktObject != null) {
```
piiertho, line 86: Here there should be a warning about inverting if and return.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -1,224 +1,280 @@
...
-
-interface GodotStatic{
-
-    fun registerAsSingleton(){
-        GarbageCollector.registerStatic(this)
-    }
-
-    fun collect();
-}
+package godot.core
```
piiertho, line 1: I suggest to add suppress warning for methods used from CPP.

piiertho (approved):
Few remarks, can be done in the new branch, not blocking for merging

chippmann (approved):

PR #103, opened 2021-03-15
References are now handled in the CPP code.

CedNaru:
Following the very poor performance  when using References (Because of multiples C++ call)
This PR totally changes the way is done. The hashmap is now in C++. The Kotlin side only has a list of weak references now.
When Godot is about to send a Reference to the JVM, we check if the pointer is in the hashmap. 
If yes, we send the index of the already existing JVM instance.
If no, we add the pointer to the hashmap and generate a new index for the JVM Instance.

It also adds a new benchmark to measure the efficiency of the GC and how fast we can convert C++ pointer to JVM instances. 

Also contains a few fixes related to the buffer and instantiation of Objects.


harness/tests/src/main/kotlin/godot/tests/Invocation.kt
```diff
@@ -299,7 +299,7 @@ class Invocation : Spatial() {
         println("NavMesh instance id after re-re-assign: ${resourceTest.getInstanceId()}")
     }
 
-    override fun _onInit() {
+    init{
```
raniejade, line 302:
```suggestion
    init {
```

kt/godot-runtime/src/main/kotlin/godot/core/KtObject.kt
```diff
@@ -5,6 +5,21 @@ import godot.util.nullptr
...
+     * Instead we use the values set in that class  */
+    private class InitConfiguration {
+        var shouldOverride = false
+        var ptr: VoidPtr = 0
+        var id: Long = 0L
+
+        fun reset() {
+            shouldOverride = false
+            ptr = 0
+            id = 0
```
raniejade, line 19: so ids always start at 1?

kt/godot-library/src/main/kotlin/godot/gen/godot/YSort.kt
```diff
@@ -26,5 +25,11 @@ open class YSort : Node2D() {
       TransferContext.callMethod(rawPtr, ENGINEMETHOD_ENGINECLASS_YSORT_SET_SORT_ENABLED, NIL)
     }
 
-  override fun __new(): VoidPtr = TransferContext.invokeConstructor(ENGINECLASS_YSORT)
+  override fun __new() {
```
piiertho, line 28: I think we should have method body in an inline method.

src/gd_kotlin.cpp
```diff
@@ -430,12 +445,14 @@ void GDKotlin::register_classes(jni::Env& p_env, jni::JObjectArray p_classes) {
     LOG_INFO("Loading classes ...")
 #endif
     for (auto i = 0; i < p_classes.length(p_env); i++) {
-        auto kt_class = new KtClass(p_classes.get(p_env, i), class_loader);
+        jni::JObject clazz = p_classes.get(p_env, i);
+        auto kt_class = new KtClass(clazz, class_loader);
```
piiertho, line 449: This was something we did not catch. Can you add pointer to auto, so that is it explicit.

src/jni/env.cpp
```diff
@@ -32,7 +32,7 @@ namespace jni {
             loadClassMethodId = cls.get_method_id(*this, "loadClass", "(Ljava/lang/String;)Ljava/lang/Class;");
         }
         jvalue args[1] = {static_cast<JValue>(new_string(name)).value};
-        auto ret = class_loader.call_object_method(*this, loadClassMethodId, args);
+        jni::JObject ret = class_loader.call_object_method(*this, loadClassMethodId, args);
```
piiertho, line 35: <3

src/kt_variant.h
```diff
@@ -25,110 +27,110 @@ namespace ktvariant {
         des->increment_position(encode_uint32(variant_type, des->get_cursor()));
     }
 
-    static void to_kvariant_fromNIL(SharedBuffer* des, const Variant& src, bool is_engine_return) {
+    static void to_kvariant_fromNIL(SharedBuffer* des, const Variant &src) {
```
piiertho, line 30: codestyle problem: `Variant&` and not `Variant &`

kt/godot-runtime/src/main/kotlin/godot/core/KtObject.kt
CedNaru: No, it starts at 0. I guess I should init it with -1

src/ref_db.h
```diff
@@ -0,0 +1,31 @@
+#ifndef GODOT_JVM_REF_DB_H
+#define GODOT_JVM_REF_DB_H
+
+#include <cstdint>
+#include "core/hash_map.h"
+#include "core/reference.h"
+#include "core/list.h"
+#include "core/os/mutex.h"
+
+class RefDB {
```
piiertho, line 10: Missing delete of copy constructor and copy assignement operator on this meyer singleton.

src/transfer_context.cpp
```diff
@@ -49,17 +49,17 @@ TransferContext::TransferContext(jni::JObject p_wrapped, jni::JObject p_class_lo
...
     methods.push_back(set_script_method);
     methods.push_back(free_object_method);
-    jni::Env env {jni::Jvm::current_env()};
+    jni::Env env{jni::Jvm::current_env()};
     j_class.register_natives(env, methods);
 }
 
 TransferContext::~TransferContext() {
-    for (auto& variant_arg : variant_args) {
+    for (auto &variant_arg : variant_args) {
```
piiertho, line 57: code style

src/transfer_context.cpp
```diff
@@ -49,17 +49,17 @@ TransferContext::TransferContext(jni::JObject p_wrapped, jni::JObject p_class_lo
...
         variant_arg = Variant();
     }
 }
 
-SharedBuffer* TransferContext::get_buffer(jni::Env& p_env) {
+SharedBuffer* TransferContext::get_buffer(jni::Env &p_env) {
```
piiertho, line 62: code style

src/transfer_context.cpp
```diff
@@ -78,23 +78,23 @@ SharedBuffer* TransferContext::get_buffer(jni::Env& p_env) {
     return &shared_buffer;
 }
 
-void TransferContext::read_return_value(jni::Env& p_env, Variant& r_ret) {
+void TransferContext::read_return_value(jni::Env &p_env, Variant &r_ret) {
```
piiertho, line 81: code style

src/transfer_context.cpp
```diff
@@ -78,23 +78,23 @@ SharedBuffer* TransferContext::get_buffer(jni::Env& p_env) {
...
     SharedBuffer* buffer{get_buffer(p_env)};
     ktvariant::get_variant_from_buffer(buffer, r_ret);
     buffer->rewind();
 }
 
-void TransferContext::write_args(jni::Env& p_env, const Variant** p_args, int args_size) {
-    SharedBuffer* buffer {get_buffer(p_env)};
+void TransferContext::write_args(jni::Env &p_env, const Variant** p_args, int args_size) {
```
piiertho, line 87: code style

src/transfer_context.cpp
```diff
@@ -133,24 +135,40 @@ void TransferContext::icall(JNIEnv* rawEnv, jobject instance, jlong jPtr, jint p
 #endif
 
     Variant::CallError r_error{Variant::CallError::CALL_OK};
-    const Variant& ret_value{methodBind->call(ptr, variant_args_ptr, args_size, r_error)};
+    const Variant &ret_value{methodBind->call(ptr, variant_args_ptr, args_size, r_error)};
```
piiertho, line 138: code style

src/transfer_context.cpp
```diff
@@ -133,24 +135,40 @@ void TransferContext::icall(JNIEnv* rawEnv, jobject instance, jlong jPtr, jint p
...
+                       vformat("Call to method with id %s failed.", method_index))
 #endif
 
     write_return_value(buffer, ret_value);
 }
 
-jlong TransferContext::invoke_constructor(JNIEnv *p_raw_env, jobject p_instance, jint p_class_index) {
-    const StringName& class_name{GDKotlin::get_instance().engine_type_names[static_cast<int>(p_class_index)]};
+void TransferContext::invoke_constructor(JNIEnv* p_raw_env, jobject p_instance, jint p_class_index) {
+    const StringName &class_name{GDKotlin::get_instance().engine_type_names[static_cast<int>(p_class_index)]};
```
piiertho, line 149: code style

src/transfer_context.cpp
```diff
@@ -181,6 +199,6 @@ void TransferContext::free_object(JNIEnv *p_raw_env, jobject p_instance, jlong p
     memdelete(owner);
 }
 
-void TransferContext::write_return_value(jni::Env& p_env, Variant& variant) {
+void TransferContext::write_return_value(jni::Env &p_env, Variant &variant) {
```
piiertho, line 202: code style

piiertho (changes requested):
Not finished the review, will have to open it in IDEs.
Still I have some little changes request.
Also we should cpp uniform initialization which is now the new standard.

src/ref_db.h
```diff
@@ -0,0 +1,33 @@
...
+    RefDB& operator=(const RefDB&)= delete;
+    ~RefDB() = default;
+
+    struct RefIndex{
+        uint64_t index = 0;
+        uint32_t counter = 0;
+    };
+
+    Mutex* mut;
+    HashMap<uint64_t, RefIndex> ref_map;
```
piiertho, line 23: If I understand correctly, it is a map of pointer to RefIndex, so key type should be `uintptr_t`.

src/ref_db.cpp
```diff
@@ -0,0 +1,69 @@
+#include "ref_db.h"
+#include "logging.h"
+
+uint64_t RefDB::get_ref_id(Reference* ref) {
+    static uint64_t id = 0;
+    RefIndex index;
+
+    {
+        auto ptr = reinterpret_cast<uint64_t>(ref);
```
piiertho, line 9: `uint64_t` => `uintptr_t`

src/ref_db.cpp
```diff
@@ -0,0 +1,69 @@
...
+        ref->reference();
+    } else {
+        ref->init_ref();
+    }
+    return index.index;
+}
+
+void RefDB::remove_ref(Reference* ref, uint32_t counter) {
+    {
+        auto ptr = reinterpret_cast<uint64_t>(ref);
```
piiertho, line 38: same here.

piiertho (changes requested):

src/ref_db.h
CedNaru: I did before, and it worked on Windows but seems that MacOs doesn't like this and the CI fails. So I changed it back to uint64-t.
Godot is supposed to handle those types as key:
https://github.com/godotengine/godot/blob/e57f6e2cca90d4c9e956fe1a7ddb3ca086521b42/core/templates/hashfuncs.h#L151

src/ref_db.h
```diff
@@ -0,0 +1,33 @@
...
+#include "core/hash_map.h"
+#include "core/reference.h"
+#include "core/list.h"
+#include "core/os/mutex.h"
+
+class RefDB {
+private:
+    RefDB();
+    RefDB(RefDB const &other) = delete;
+    RefDB& operator=(const RefDB&)= delete;
```
piiertho, line 14: Deleted memebers should be public.

piiertho (changes requested):

chippmann (changes requested):
Very nice work!
Some minor comments of which are many optional
Flamegraph looks great as well for me.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -212,28 +258,34 @@ object GarbageCollector {
...
             forceJvmGc()
-            checkAndClean()
+            if (checkAndClean()) {
+                begin = Instant.now()
+            }
+
             val finish = Instant.now()
-            if (Duration.between(begin, finish).toMillis() > 5000) {
-                throw GCEndException(
+            if (Duration.between(begin, finish).toMillis() > 10000) {
```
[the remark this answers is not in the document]
CedNaru: It's when I was testing my changes in the benchmark. I was creating over 30 million instances and 5 were not enough to clean everything. 
I just forgot to set it back to 5.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -81,37 +106,42 @@ object GarbageCollector {
...
             }
-        } else {
-            return null
         }
+        return null
     }
 
-    fun getRefInstance(ptr: VoidPtr): KtObject? {
-        return refWrappedMap[ptr]?.get()
+    fun getRefInstance(index: Int): KtObject? {
```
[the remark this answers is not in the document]
CedNaru: That actually something I was wondering. Int (32bits) can go up to around 2 billion, so yeah it's possible to reach that limit.
I don't think widening it to Long (64 bits) should have any measurable cost. And that would make it impossible to reach the limit this way.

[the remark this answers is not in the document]
CedNaru: Well after checking it's actually not possible.
Godot holds a list of all Objects currently alive and it can't be bigger than 24 bits, so we will never go above the limit of a int32

src/transfer_context.cpp
```diff
@@ -136,21 +138,37 @@ void TransferContext::icall(JNIEnv* rawEnv, jobject instance, jlong jPtr, jint p
...
+void TransferContext::invoke_constructor(JNIEnv* p_raw_env, jobject p_instance, jint p_class_index) {
     const StringName& class_name{GDKotlin::get_instance().engine_type_names[static_cast<int>(p_class_index)]};
     Object* ptr = ClassDB::instance(class_name);
 
+    auto raw_ptr = reinterpret_cast<uintptr_t>(ptr);
+    int id;
+
 #ifdef DEBUG_ENABLED
-    JVM_ERR_FAIL_COND_V_MSG(!ptr, 0, vformat("Failed to instantiate class %s", class_name))
+    JVM_ERR_FAIL_COND_MSG(!ptr, vformat("Failed to instantiate class %s", class_name))
```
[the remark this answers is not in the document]
CedNaru: Godot would print an error in release builds anyway a bit before in that block.
This one is only to print that the error started from a JVM call.

CedNaru:
About the use of uint64 instead of uintptr in the ref_db.

I first tried to use uintptr because it gives the clear hint that it means the keys of the HashMap are pointers to References.
It was working well except in MacOs, because the way uintptr is implement doesn't give an implicit conversion to uint64.
Godot's Hashmap accepts uint64 as a possible key but not uintptr. Without implicit conversion to uintptr, the compiler doesn't know which Hash functions to use so it gives us an error.

So I chose to use uint64 directly. That shouldn't give any issue. We are only supporting 64 bits platforms and we don't use the reinterpreted pointer to manipulate memory, just as a key for the HashMap. As long as we receive the same pointer, we will always get the same uint64 when we reinterpret (as the 2 have the same size).

chippmann (approved):

piiertho (approved):

piiertho:
Does not prevent to merge, but still we should use uniform initialization.

PR #104, opened 2021-03-18
feature/use scons cach in ci

piiertho:
This enables scons cache to fasten CI builds. Unfortunately this does not seems to fasten on windows build, maybe someone will have an idea.

### DIFF
--- a/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
+++ b/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
@@ -1,6 +1,9 @@
 name: Check PR - Engine Editor Debug And Tests
 on: [pull_request]
 
+env:
+  GODOT_BASE_VERSION: 3.2.3
+
 jobs:
   build-godot-library:
     strategy:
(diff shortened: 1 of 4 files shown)


raniejade (commented):
How big of an improvement are we talking about? Also, how is the cache cleaned up? Does scons take care of that? If not, the cache may continue to grow and the benefits we get from caching is lost because it will take a while to download the cache.

piiertho:
> How big of an improvement are we talking about? Also, how is the cache cleaned up? Does scons take care of that? If not, the cache may continue to grow and the benefits we get from caching is lost because it will take a while to download the cache.

Cache is per branch.
Files would be overriden if there's any change (we're talking about `.o` / `.obj` files).  
It reduces a lot compilation times if a build has already been done in a PR, as it will use former binairies.

image

raniejade:
> Cache is per branch.

What happens to older cache branches? i.e after being merged.

piiertho:
> > Cache is per branch.
> 
> What happens to older cache branches? i.e after being merged.

From cache action readme:

A repository can have up to 5GB of caches. Once the 5GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

So it seems to be already managed.

raniejade (approved):
LGTM - cache is localized to a PR so that's good.

chippmann (approved):

PR #112, opened 2021-03-21
Implement project wizard

### LINKED ISSUE #110: Add intellij project template
Add a intellij project template to start a new godot project with godot kotlin jvm all pre setup so the user does not have to setup our plugin, the godot project and the jvm project by himselve.

chippmann:
Resolves #110.
From my point of view this could even be merged for alpha. But in any case it would require that our artifacts are published. Otherwise the newly created project cannot resolve any of the dependencies.


kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/module/GodotModuleBuilder.kt
```diff
@@ -0,0 +1,184 @@
...
+    override fun getGroupName() = groupName
+
+    override fun getCustomOptionsStep(context: WizardContext, parentDisposable: Disposable): ModuleWizardStep {
+        wizardContext = context
+        return object : ModuleWizardStep() {
+            val groupIdTextField = JTextField("com.example")
+            val artifactIdTextField = JTextField("game")
+            val versionTextField = JTextField("0.0.1-SNAPSHOT")
+            val kotlinVersionTextField = JTextField("1.4.31")
+            val godotKotlinJvmVersionTextField = JTextField("0.1.0-3.2")
```
piiertho, line 60: I think we should rely on official tags like 3.2.3.

piiertho (changes requested):

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/module/GodotModuleBuilder.kt
chippmann: Indeed. I'll update all of the versions in #109 as well

CedNaru: The plugin is built at the same time as the rest of the bindings, so I think the choice of the bindings version shouldn't exist and be hardcoded in the plugin.
Nothing guarantee that the project template will stay stable in the future (because of this project and because of Godot as well).
Let's say we later update to Godot 4.0, if the user doesn't update the plugin to the new version, there is a good chance nothing will work if they target the 4.0 version of our bindings with a old plugin.

chippmann: Still everything will work. Just the plugin may act crazy. But that's the case for all Intellij plugin's that add framework support (for example spring boot, or sqldelight).
You always have to make sure you're using the correct version of the plugin if you use it together with a specific framework version.

chippmann: But yeah i can of course remove this option alltogether

CedNaru (approved):

piiertho (approved):

kt/plugins/godot-intellij-plugin/src/main/resources/template/.gitignore.template
```diff
@@ -0,0 +1,138 @@
...
+.idea/**/dataSources.ids
+.idea/**/dataSources.local.xml
+.idea/**/sqlDataSources.xml
+.idea/**/dynamic.xml
+.idea/**/uiDesigner.xml
+.idea/**/dbnavigator.xml
+
+# Gradle
+.idea/**/gradle.xml
+.idea/**/libraries
```
raniejade, line 30: Ignore the full `.idea` directory instead of specifying things one by one. You can also use `!` as the prefix - which means don't ignore this <pattern>

kt/plugins/godot-intellij-plugin/src/main/resources/template/.gitignore.template
```diff
@@ -0,0 +1,138 @@
+
+# Created by https://www.toptal.com/developers/gitignore/api/intellij+all,kotlin,gradle
```
raniejade, line 2: there's too much verbosity in this file.

raniejade (approved):

PR #114, opened 2021-03-24
Implement godot plugin

### LINKED ISSUE #111: Add editor build command
Add a button and dialogs to the godot editor to be able to build the project from within the editor. Similarly like we had for K/N or like C# is doing it.

chippmann:
Resolves #111 

A little video about the features: https://drive.google.com/file/d/122pM1qdwi4JXAXcmuzny2_0Pyoo1mWGY/view?usp=sharing

Summary:
- Build
- Show build logs
- Build before scene or project run
- Warnings when kt files are moved, deleted within the editor


src/editor/build/build_manager.h
```diff
@@ -0,0 +1,45 @@
...
+
+    void build_project_non_blocking();
+
+    bool can_build_project();
+
+    bool is_build_finished();
+
+    void update_build_state();
+
+    String get_log();
```
piiertho, line 25: I suggest to return `const String&` as it is a getter, this gives indication to other devs.

src/editor/build/build_manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    EditorNode::progress_task_step("build_godot_kotlin_jvm", "Done", 2); //dummy to not start at 0% or 100%
+    EditorNode::progress_end_task("build_godot_kotlin_jvm");
+
+    GodotKotlinJvmEditor::get_instance()->on_build_check_timeout(); //manually call what timer would call to update ui
+    return result == Error::OK && last_build_successful();
+}
+
+void BuildManager::build_project_non_blocking() {
+    if (build_thread == nullptr) {
```
piiertho, line 38: `if (build_thread)` is enough here.

src/editor/build/build_manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return result;
+}
+
+// convenience function for better readability in some places
+bool BuildManager::is_build_finished() {
+    return can_build_project();
+}
+
+void BuildManager::update_build_state() {
+    if (build_thread == nullptr) {
```
piiertho, line 58: same here

src/editor/build/build_manager.h
```diff
@@ -0,0 +1,45 @@
...
+
+#ifndef GODOT_JVM_BUILD_MANAGER_H
+#define GODOT_JVM_BUILD_MANAGER_H
+
+#include <core/error_list.h>
+#include <core/os/os.h>
+#include <core/os/thread.h>
+
+class BuildManager {
+public:
```
piiertho, line 12: It misses deletion of assignement operator and copy constructor on this meyer singleton.

src/editor/dialog/about_dialog.cpp
```diff
@@ -0,0 +1,99 @@
...
+    ClassDB::bind_method(D_METHOD("on_about_to_show"), &AboutDialog::on_about_to_show);
+    ClassDB::bind_method(D_METHOD("on_checkbox_toggled"), &AboutDialog::on_checkbox_toggled);
+    ClassDB::bind_method(D_METHOD("on_url_clicked"), &AboutDialog::on_url_clicked);
+
+    set_title("About Godot Kotlin JVM");
+
+    connect("about_to_show", this, "on_about_to_show");
+
+    // Main VBoxContainer (icon + label on top, checkbox at bottom)
+    VBoxContainer* about_vbox = memnew(VBoxContainer);
```
piiertho, line 23: Cpp uniform init :)

src/editor/dialog/build_dialog.cpp
```diff
@@ -0,0 +1,38 @@
+
+#ifdef TOOLS_ENABLED
+
+#include "build_dialog.h"
+
+#include <editor/editor_scale.h>
+#include <modules/kotlin_jvm/src/editor/build/build_manager.h>
+
+BuildDialog::BuildDialog() {
```
piiertho, line 9: You should add initialization of `scroll_container` and `log_label` to constructor this way:
```cpp
BuildDialog::BuildDialog() : scroll_container(memnew(ScrollContainer)), log_label(memnew(LogLabel)) {
    ...
}
```

src/editor/godot_kotlin_jvm_editor.h
```diff
@@ -0,0 +1,47 @@
...
+
+
+#include <editor/editor_plugin.h>
+#include "dialog/about_dialog.h"
+#include "dialog/build_dialog.h"
+#include "menu_option.h"
+#include "panel/bottom_panel.h"
+
+class GodotKotlinJvmEditor : public EditorPlugin {
+public:
```
piiertho, line 15: It misses copy constructor and assignement operator deletion.

src/editor/godot_kotlin_jvm_editor.h
```diff
@@ -0,0 +1,47 @@
...
+    Timer* build_check_timer;
+
+    void on_build_check_timeout();
+
+protected:
+    void _notificationv(int p_notification, bool p_reversed) override;
+
+    bool build() override;
+
+private:
```
piiertho, line 27: Default constructor should be private.

piiertho (changes requested):

src/editor/build/build_manager.cpp
```diff
@@ -0,0 +1,133 @@
...
+    build_finished = true;
+    build_mutex->unlock();
+
+    return result;
+}
+
+bool BuildManager::last_build_successful() {
+    build_mutex->lock();
+    //TODO: find a better way. We cannot use the exit code of the task as it's successful also for failed builds because of the way we have to execute composite commands in godot. See: https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-execute
+    bool successful = build_log.find("BUILD FAILED") < 0;
```
raniejade, line 123: weird, gradle should return a non-zero exit code if any of its tasks fail.

raniejade: note CI wouldn't work if that's not the case :) i.e we won't be notified for build failures.

chippmann: Ah damn. That's a leftover from a platform specific implementation before i was using a background thread with godots `Thread` class.
I'll change it. This is not necessary anymore.

chippmann: The reason for this was: before i was piping all output to a file. And because of the way godots `execute` works i needed to call `bash` and `CMD.exe` directly which always returned exitcode 0 regardless of gradles output because all of that was piped to the file.

chippmann: The reason for this was: before i was piping all output to a file. And because of the way godots execute works i needed to call bash and CMD.exe directly which always returned exitcode 0 regardless of gradles output because all of that was piped to the file.

src/editor/build/build_manager.h
chippmann: As discussed on discord: this is not possible. I need to return a copy of the string because `build_log` is written from a background thread and needs a lock to read from it.

piiertho (approved):
Great job !

CedNaru (approved):

Issue #123, opened 2021-04-01
PoolVariantArrays are missing specific functions


chippmann:
Example: `get_string_from_ascii` in `PoolByteArray` is missing.
Also conversion functions from kotlin collections to PoolArrays are missing.

PR #128, opened 2021-04-07
Fix/poolarray

### LINKED ISSUE #124: PoolArray hash functions are recursive
```kotlin
override fun hashCode(): Int {
    return hashCode()
}
```
is a recursive loop and will crash if invoked.

CedNaru:
Fix #124  #123 

The missing functions were not part of the PoolArray class, I had to retrieve them from the VariantCall structure.
No other choice than to call the function with a StringName.


kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolByteArray.kt
```diff
@@ -8,12 +8,35 @@ import godot.util.VoidPtr
...
+         */
+        COMPRESSION_ZSTD,
+
+        /**
+         * Uses the gzip compression method.
+         */
+        COMPRESSION_GZIP
+    }
+
+
```
chippmann, line 33: Hmm isn't the idea that the compression modes in `File` class are used? At least thats to where the link [CompressionMode](https://docs.godotengine.org/en/stable/classes/class_file.html#enum-file-compressionmode) in https://docs.godotengine.org/en/stable/classes/class_poolbytearray.html#class-poolbytearray-method-compress is linking.

chippmann (changes requested):
I'd also add some gut tests for it? I think the ascii functions are a good candidate for it

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolByteArray.kt
CedNaru: Yeah, it is in File. 
But PoolByteArray is a different module than the GenAPI. I can't access it from there.

chippmann (approved):

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolByteArray.kt
piiertho: So maybe a comment here would help ?

src/bridges/pool_byte_array_bridge.cpp
```diff
@@ -104,6 +148,78 @@ void PoolByteArrayBridge::engine_call_appendArray(JNIEnv* p_raw_env, jobject p_i
...
+}
+
+void PoolByteArrayBridge::engine_call_decompress(JNIEnv* p_raw_env, jobject p_instance, jlong p_raw_ptr) {
+    jni::Env env{p_raw_env};
+    Variant args[2] = {};
+    TransferContext* transfer_context{GDKotlin::get_instance().transfer_context};
+    transfer_context->read_args(env, args);
+
+    Variant pool{*from_uint_to_ptr<PoolByteArray>(p_raw_ptr)};
+    static StringName func_name{_scs_create("decompress")};
```
piiertho, line 173: Not needed anymore ?

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
```diff
@@ -1,538 +1,547 @@
 package godot.tests
 
-import godot.ARVRServer
-import godot.Button
-import godot.NavigationMesh
-import godot.Object
-import godot.Spatial
+import godot.*
```
piiertho, line 3: Godot replaced spaces

kt/godot-runtime/src/main/kotlin/godot/core/pool/PoolByteArray.kt
```diff
@@ -8,12 +8,35 @@ import godot.util.VoidPtr
...
     //PROPERTIES
     val size: Int
         get() {
-		    Bridge.engine_call_size(_handle)
-			return TransferContext.readReturnValue(VariantType.JVM_INT) as Int
-		}
+            Bridge.engine_call_size(_handle)
```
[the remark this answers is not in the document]
CedNaru: It's just the result of auto reformat.

piiertho (approved):

raniejade (approved):

PR #137, opened 2021-04-16
Fix default values of user script types

chippmann:
This fixes a bug where, when a class has properties with default values which are other user scripts, and this class is registered after the other class, the editor/game would crash at startup:
```
Godot-JVM: Registered _VisualScriptEditor engine type with index 621.
Godot-JVM: Done registering managed engine types...
ERROR: get: FATAL: Index p_index = 6 is out of bounds (size() = 0).
   At: ./core/cowdata.h:152.
handle_crash: Program crashed with signal 4
Dumping the backtrace. Please include this when reporting the bug on https://github.com/godotengine/godot/issues
[1] /home/cedric/projects/04_godot/modules/kotlin_jvm/harness/tests/jre/lib/server/libjvm.so(+0xbe1f66) [0x7f96159eff66] (??:0)
[2] /home/cedric/projects/04_godot/modules/kotlin_jvm/harness/tests/jre/lib/server/libjvm.so(JVM_handle_linux_signal+0x2cb) [0x7f96159f5d1b] (??:0)
[3] /home/cedric/projects/04_godot/modules/kotlin_jvm/harness/tests/jre/lib/server/libjvm.so(+0xbda8fa) [0x7f96159e88fa] (??:0)
[4] /usr/lib/libpthread.so.0(+0x13960) [0x7f96cf76f960] (??:0)
[5] CowData<Ref<KotlinScript> >::get(int) const (/home/cedric/projects/04_godot/./core/cowdata.h:152 (discriminator 7))
[6] Vector<Ref<KotlinScript> >::operator[](int) const (/home/cedric/projects/04_godot/./core/vector.h:85)
[7] TransferContext::set_script(JNIEnv_*, _jobject*, long, int, _jobject*, _jobject*) (/home/cedric/projects/04_godot/modules/kotlin_jvm/src/transfer_context.cpp:195)
[8] [0x7f95d900f710] (??:0)
-- END OF BACKTRACE --
```

## Problem:
Example is our tests project:
In `Invocation` we have this property: `var registerObjectNullablePreInit: OtherScript? = OtherScript()` 
in the registrar there is this default value provider: `open val registerObjectNullablePreInitDefaultValue: OtherScript = OtherScript()` 
when this executes it tries to set the script in the `init` block of `KtObject` with:
```kotlin
//...
            // inheritance in Godot is faked, a script is attached to an Object allow
            // the script to see all methods of the owning Object.
            // For user types, we need to make sure to attach this script to the Object
            // rawPtr is pointing to.
            val classIndex = TypeManager.userTypeToId[this::class]
            // If user type
            if (classIndex != null) {
                println("called with $this")
                TransferContext.setScript(rawPtr, classIndex, this, this::class.java.classLoader)
            }
//...
```
Which returned `null` for  `val classIndex = TypeManager.userTypeToId[this::class]` for clean builds as `OtherScript` was registered *after* `Invocation` but on an incremental build it is registered *before* `Invocation` which leads to the problem that cpp does not yet know the class as we pass all of them at once. So kotlin knows about our classes before cpp does. Which is fine normally but not in this case.

## Solution:
To fix this, this PR registers all classes first in cpp and fetches all members *after* all classes are registered.
It also fetches the default values lazily.

Depends on: https://github.com/utopia-rise/godot-kotlin-entry-generator/pull/27

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -546,7 +546,7 @@ GDKotlin::GDKotlin() :
 }
 
 void GDKotlin::register_members(jni::Env& p_env) {
-    auto map_entry = classes.front();
+    auto* map_entry{classes.front()};
     while (map_entry) {
         map_entry->get()->fetch_members();
         map_entry = map_entry->next();
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -188,6 +188,11 @@ void register_user_types_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_type
     LOG_VERBOSE("Done registering user types.");
 }
 
+void register_user_types_members_hook(JNIEnv* p_env, jobject p_this) {
+    jni::Env env(p_env);
+    GDKotlin::get_instance().register_members(env);
+}
+
 void GDKotlin::init() {
     if (Main::is_project_manager()) {
 #ifdef DEBUG_ENABLED
@@ -377,7 +382,7 @@ void GDKotlin::init() {
     bootstrap = new Bootstrap(instance, class_loader);
 
     bootstrap->register_hooks(env, load_classes_hook, unload_classes_hook, register_engine_types_hook,
-                              register_user_types_hook);
+                              register_user_types_hook, register_user_types_members_hook);
     bool is_editor = Engine::get_singleton()->is_editor_hint();
 
 #ifdef TOOLS_ENABLED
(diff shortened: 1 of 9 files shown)


src/gd_kotlin.cpp
```diff
@@ -539,3 +544,11 @@ GDKotlin::GDKotlin() :
         is_gc_started(false),
         transfer_context(nullptr) {
 }
+
+void GDKotlin::register_members(jni::Env& p_env) {
+    auto map_entry = classes.front();
```
piiertho, line 549: I dont really like auto when no cast or explicit constructor call. Can we at least write ptr so that's explicit
```cpp
auto* map_entry{classes.front()};
```

CedNaru (approved):

raniejade (approved):

piiertho (approved):

Issue #133, opened 2021-04-16
Editor crashes when opening a project if JAVA_HOME is not defined


raniejade:
This happens even if there is an embedded `jre` and even with non kotlin projects.

mrParabol:
this problem still persists.
godot : 3.5.1- 0.5.0
godot-JVM intellijidea addons :0.5.0

PR #141, opened 2021-04-16
Fix: Use OS singleton instead of stdlib to get JAVA_HOME env variable

piiertho:
We were using `getenv` method from `stdlib` to get `JAVA_HOME` environment variable.  
This method is deprecated.  
To avoid problems and potential bugs on finding `JAVA_HOME`, this makes use of `OS` singleton.

### DIFF
--- a/src/jni/jvm_loader.cpp
+++ b/src/jni/jvm_loader.cpp
@@ -1,9 +1,7 @@
 #include "jvm_loader.h"
 
-#include <core/engine.h>
 #include <core/os/os.h>
 #include <core/project_settings.h>
-#include <modules/kotlin_jvm/src/logging.h>
 
 #ifndef __ANDROID__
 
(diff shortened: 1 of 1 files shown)


chippmann (approved):

CedNaru (approved):

raniejade (approved):

src/jni/jvm_loader.cpp
```diff
@@ -69,7 +67,7 @@ String jni::JvmLoader::get_jvm_lib_path() {
 }
 
 String jni::JvmLoader::get_path_to_locally_installed_jvm() {
-    String javaHome{getenv("JAVA_HOME")};
+    String javaHome{OS::get_singleton()->get_environment("JAVA_HOME")};
 
     if (javaHome.empty()) {
         LOG_ERROR("JAVA_HOME is not defined! Exiting...")
```
raniejade, line 73: @chippmann Outside of this PR, but this probably causes the crash described in #133. Do we need JAVA_HOME to be always defined? or can we used the embedded jre?

chippmann: Yes this is definitely the issue. See the exit(1). we did this on purpose. IIRC we discussed this once and decided that during development we should use the locally installed jdk. But i guess we could also change that.

raniejade: Ahh, I guess that's fine as long as it is mentioned in the docs.

PR #142, opened 2021-04-16
Hotfix/fix from godot to kotin basis

### LINKED ISSUE #132: Basis doesn't return correct values
I implemented the 3D platform demo with our current alpha version.
It runs properly except for the camera doing totally inaccurate movements. Looking at the Camera script we use Basis a lot in it so I suspect one or several methods of this core type were not properly translated from C++ to Kotlin.

piiertho:
There several bugs in cpp to kotlin sending:
- Basis data were not set correctly (confusion between axis and elements)
- Variant type for plane was not sent through buffer, which does not respect protocol.

This adds core types identity tests for cpp to kotlin.

This resolves #132

### DIFF
--- a/harness/tests/build.gradle.kts
+++ b/harness/tests/build.gradle.kts
@@ -1,7 +1,7 @@
 
 plugins {
     kotlin("jvm") version "1.4.10"
-    id("com.utopia-rise.godot-kotlin-jvm") version "0.1.2-3.2.3"
+    id("com.utopia-rise.godot-kotlin-jvm") version "0.1.3-3.2.3-SNAPSHOT"
 }
 
 repositories {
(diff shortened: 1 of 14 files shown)


chippmann (approved):

CedNaru (changes requested):
I would make a small change as well to basis:

The kotlin array operator doesn't match the C++ one:

Kotlin return the columns using getAxis.
image

But CPP returns the rows by using the internal array.
(The 3 vector3 in C++ are put into an array. I can't do that in Kotlin because Array is not primitive which would add an additional instance and indirection in Basis.)

image

piiertho:
> I would make a small change as well to basis:
> 
> The kotlin array operator doesn't match the C++ one:
> 
> Kotlin return the columns using getAxis.
> image
> 
> But CPP returns the rows by using the internal array.
> (The 3 vector3 in C++ are put into an array. I can't do that in Kotlin because Array is not primitive which would add an additional instance and indirection in Basis.)
> 
> image

For now we follow same as GDScript. I added small tests on get and set of `Vector3` and `Basis` which ensure this convention.

raniejade (approved):

CedNaru (approved):

PR #164, opened 2021-04-21
NativeBuffer is now way smaller

### LINKED ISSUE #145: NativeBuffer is too big.
Currently, we hardcoded a 20MB buffer for exchanging data between the JVM and C++
The maximum number of arguments for a function currently is 5 parameters so the logic would be to have a buffer of the size:
Biggest CoreType x 5.
Sadly String can have an arbitrary length so we just arbitrary set a huge set to accept big strings.

The issue is the huge amount of wasted space. The buffer is thread-local so each thread is going to have its own 20MB buffer.
If people start to want to play with coroutines with a pool of threads like several told us on Reddit/Discord, we will have quickly reach 100MB taken just to store 5 variables.

My current idea is to allow a maximum size for String ( like 1K).
If a String is below that limit, it will be sent normally with the buffer. 
Above that limit, it's going to be sent with a regular JNI call. 

To implement that, an additional bit is going to be added in the buffer when a string is sent.  0 for a short String, 1 for a long String.
When it's 0 we proceed like we do today.
When it's 1 we make a JNI call, that call is going to push the string into a new String Queue in TransfertContext. When the receiver reads the bit, instead of reading the rest of the buffer, the queue will be poped to retrieve the String.
We need that queue in both Kotlin and C++, it's a Queue because several strings can potentially be sent in the same function.

CedNaru:
Resolves #145 
This PR got 2 major changes:

- I had to change the way the ClassLoader is created because my current rework was harder without that change. We were relying too much on GDKotlin as a singleton. So I separated the two. Now ClassLoader is its own singleton. 
- There is now a maximum size allowed when we want to exchange strings between C++ and Kotlin. When we are below that maximum, the string is sent through the buffer like before. If we go above it, it switches to a regular JNI call. When it's done that way. The string is stored in a new LongStringQueue singleton so we can retrieve the strings from it instead of reading the buffer.

The current maximum is set to 512 but it can be changed with an argument when we start the engine.

The new formula for the size of the buffer is :  (StringMaxSize + 12) * ARGUMENT_MAX_COUNT + 4
The new buffer size is around 2.6 KB.
It was previously 20MB.


harness/tests/src/main/kotlin/godot/tests/coretypes/StringTest.kt
```diff
@@ -0,0 +1,42 @@
...
+        "\n" +
+        "I'll optimize the object-oriented DOM state, that should open source the API abstraction!\n" +
+        "\n" +
+        "So bubble-sorting the Stack Overflow won't do anything, we need to graph the minimum viable FIFO algorithm!\n" +
+        "\n" +
+        "The OOP module is down, configure the behavior-driven context so we can inherit the DOM neck beard!\n" +
+        "\n" +
+        "The SOAP middleware is down, f*** the serverless stream so we can circle back the LGTM constant!\n" +
+        "\n" +
+        "You can't ship the continuous integration without blogging the junior IoT Stack Overflow!"
```
chippmann, line 41: Why not just a multiline string?

```kotlin
"""
cool
string
here
""".trimIndent()
```

harness/tests/test/unit/test_string.gd
```diff
@@ -0,0 +1,41 @@
...
+
+So bubble-sorting the Stack Overflow won't do anything, we need to graph the minimum viable FIFO algorithm!
+
+The OOP module is down, configure the behavior-driven context so we can inherit the DOM neck beard!
+
+The SOAP middleware is down, f*** the serverless stream so we can circle back the LGTM constant!
+
+You can't ship the continuous integration without blogging the junior IoT Stack Overflow!"""
+
+var longest_short_str: String = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
chippmann, line 24: maybe add a comment stating why and what the exact char count is

chippmann: Also so we don't forget to update it when changing sizes

harness/tests/test/unit/test_string.gd
```diff
@@ -0,0 +1,41 @@
...
+
+var scriptpath: Script = load("res://src/main/kotlin/godot/tests/coretypes/StringTest.kt")
```
chippmann, line 26: We should not use the `load` functions in the tests IMO
Even if it does not run for you, for CI and all others it does

harness/tests/test/unit/test_string.gd
```diff
@@ -0,0 +1,41 @@
...
+
+
+func test_long_string() -> void:
+	var script = scriptpath.new()
+	assert_eq(long_str, script.get_long_string(), "String on JVM side should be same as gdscript one.")
+	assert_eq(long_str, script.identity(long_str), "String on JVM side should be same as gdscript one.")
+	
+func test_flood_string() -> void:
+	var script = scriptpath.new()
+	assert_eq(script.fill_the_buffer(longest_short_str, longest_short_str, longest_short_str, longest_short_str, longest_short_str), true, "Should return true when buffer is filled")
```
chippmann, line 41: I'm not sure what this test should do as you always return true on the kotlin side?

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -124,17 +124,33 @@ enum class VariantType(
...
+                    val stringSize = buffer.int
+                    val charArray = ByteArray(stringSize)
+                    buffer.get(charArray, 0, stringSize)
+                    val str = String(charArray, Charsets.UTF_8)
+                    str
+                }
             },
             { buffer: ByteBuffer, any: Any ->
                 buffer.variantType = STRING.ordinal
                 any as String
```
chippmann, line 142: Not introduced by you but why is this not a `require(any is String)`?

src/gd_kotlin.cpp
```diff
@@ -243,10 +178,10 @@ void GDKotlin::init() {
...
             }
-        } else if (cmd_arg.find("--jvm-to-engine-shared-buffer-size") >= 0) {
+        } else if (cmd_arg.find("--jvm_to_engine_max_string_size") >= 0) {
             String result;
             if (split_jvm_debug_argument(cmd_arg, result) == OK) {
-                jvm_to_engine_shared_buffer_size = result.to_int();
+                jvm_to_engine_max_string_size = result.to_int();
                 //TODO: Link to documentation
                 LOG_WARNING(
                         vformat("Warning ! Buffer capacity was changed to %s, this is not a recommended practice",
```
chippmann, line 187: Warning should be changed or removed IMO

chippmann: While at it the link to the docs should be added: https://godot-kotl.in/en/latest/advanced/commandline-args/

src/long_string_queue.h
```diff
@@ -0,0 +1,46 @@
+#ifndef GODOT_JVM_LONG_STRING_QUEUE_H
+#define GODOT_JVM_LONG_STRING_QUEUE_H
+
+#include "java_instance_wrapper.h"
+
+class LongStringQueue : public JavaInstanceWrapper<LongStringQueue> {
```
chippmann, line 6: Shouldn't this class be put in the `bridges` folder?

src/class_loader.h
```diff
@@ -0,0 +1,39 @@
+#ifndef GODOT_JVM_CLASS_LOADER_H
+#define GODOT_JVM_CLASS_LOADER_H
+
+#include "jni/wrapper.h"
+
+class ClassLoader {
```
chippmann, line 6: Shouldn't this class be put in the `jni` folder?

src/long_string_queue.h
```diff
@@ -0,0 +1,46 @@
...
+
+    static void send_string_to_cpp(JNIEnv* p_raw_env, jobject p_instance, jstring p_string);
+
+private:
+
+    static LongStringQueue init();
+
+DECLARE_JNI_METHODS(
+        JNI_METHOD(QUEUE_STRING, "queueString", "(Ljava/lang/String;)V")
+        JNI_METHOD(SET_STRING_MAX_SIZE, "setStringMaxSize", "(Ljava/lang/String;)V")
```
chippmann, line 41: Wrong parameter type. Should be `Int` -> `I` and not `String` -> `Ljava/lang/String;`

chippmann (changes requested):
You also need to update the command line docs here: https://godot-kotl.in/en/latest/advanced/commandline-args/
Also it would be good if you could write a little writeup how the buffer works in the `knowledge-base` but this is purely optional :-)

harness/tests/src/main/kotlin/godot/tests/coretypes/StringTest.kt
CedNaru: because that's what Idea need automatically when I pasted the string . :p

harness/tests/test/unit/test_string.gd
CedNaru: I didn't find a assert_not_crashing so I just made it so it returns true. If the buffer doesn't have the right size, it will throw an error and not return true.

chippmann: :+1:

chippmann (approved):
LGTM. The cpp part has to be looked at more closely by @piiertho or @raniejade though.

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -124,17 +124,33 @@ enum class VariantType(
...
+                    val charArray = ByteArray(stringSize)
+                    buffer.get(charArray, 0, stringSize)
+                    val str = String(charArray, Charsets.UTF_8)
+                    str
+                }
             },
             { buffer: ByteBuffer, any: Any ->
                 buffer.variantType = STRING.ordinal
-                any as String
+                require(any is String)
```
piiertho, line 142: I don't like this additional check in buffer context. Just cast it, if it is not right type, it will crash later.

src/gd_kotlin.cpp
```diff
@@ -243,13 +178,13 @@ void GDKotlin::init() {
                     jvm_jmx_port = "9010";
                 }
             }
-        } else if (cmd_arg.find("--jvm-to-engine-shared-buffer-size") >= 0) {
+        } else if (cmd_arg.find("--jvm_to_engine_max_string_size") >= 0) {
```
piiertho, line 181: Other command lines argument are with `-`. For coherence, this should be `--jvm-to-engine-max-string-size`

src/gd_kotlin.cpp
```diff
@@ -327,22 +262,26 @@ void GDKotlin::init() {
...
     jni::Env env{jni::Jvm::current_env()};
-    jni::JObject current_thread = get_current_thread(env);
-    class_loader = create_class_loader(env, bootstrap_jar, jni::JObject(nullptr)).new_global_ref<jni::JObject>(env);
-    set_context_class_loader(env, current_thread, class_loader);
+
+    jni::JObject class_loader = ClassLoader::provide_loader(env, bootstrap_jar, jni::JObject(nullptr));
+    ClassLoader::set_default_loader(class_loader);
+    class_loader.delete_local_ref(env);
+
+    class_loader = ClassLoader::get_default_loader();
```
piiertho, line 270: Do we need to set it twice ?

src/gd_kotlin.cpp
```diff
@@ -327,22 +262,26 @@ void GDKotlin::init() {
 
     LOG_INFO(vformat("Loading bootstrap jar: %s", bootstrap_jar))
     jni::Env env{jni::Jvm::current_env()};
-    jni::JObject current_thread = get_current_thread(env);
-    class_loader = create_class_loader(env, bootstrap_jar, jni::JObject(nullptr)).new_global_ref<jni::JObject>(env);
-    set_context_class_loader(env, current_thread, class_loader);
+
+    jni::JObject class_loader = ClassLoader::provide_loader(env, bootstrap_jar, jni::JObject(nullptr));
```
piiertho, line 266: Uniform init.

src/jni/class_loader.cpp
```diff
@@ -0,0 +1,81 @@
+#include "class_loader.h"
+#include <cassert>
+
+ClassLoader::ClassLoader(): class_loader{jni::JObject()} {
+
+}
+
+jni::JObject to_java_url(jni::Env& env, const String& bootstrapJar) {
+    jni::JClass cls = env.find_class("java/io/File");
```
piiertho, line 9: I know this was this way before in GDKotlin, but maybe that's the good time to make use of uniform init.

piiertho (changes requested):

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
CedNaru: But it will crash anyway because we will have a TypeCastException

src/gd_kotlin.cpp
CedNaru: still ugly

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
piiertho: Yep, so no need for require

src/jni/class_loader.cpp
CedNaru: even uglier

src/gd_kotlin.cpp
piiertho: Did not mais attention to delete before

kt/godot-runtime/src/main/kotlin/godot/core/Variant.kt
CedNaru: Finally we put require everywhere because it's actually more efficient after benchmarking the 2 solutions

piiertho (approved):

raniejade (approved):
Do we need to set a minimum limit to the value the user provides?

CedNaru:
Good catch.
I added a minimum size for the buffer so it's big enough for other types even if we set the string size to 0.
In itself, there is nothing wrong with a string size set to 0, it just means that all of them will be sent with JNI.
In that case, we get the smaller buffer possible: 264 Bytes

PR #174, opened 2021-04-23
Feature/godot 3.3.0

piiertho:
This updates module to Godot 3.3.
Please test it carefully on your side before we merge it.


harness/benchmarks/godot-project.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Godot.NET.Sdk/3.2.3">
+<Project Sdk="Godot.NET.Sdk/3.3">
```
chippmann, line 1: This will not work as the target framework version also need to change IIRC
Once i have mono setup again, I'll push a fix to this PR

chippmann (changes requested):

chippmann (approved):

docs/src/doc/versioning.md
```diff
@@ -1,7 +1,7 @@
 The module uses semantic versioning for its own versions but adds a suffix for the supported godot version:
 
-Full version: `0.1.0-3.2.3`
+Full version: `0.1.0-3.3.0`
```
raniejade, line 3: nit: I don't think its scalable to change this documentation every time we bump the engine version. The versions here are an example, so this specific change is unnecessary.

raniejade (approved):

CedNaru (approved):

PR #175, opened 2021-04-23
Core type helper functions

### LINKED ISSUE #146: Add coreType helper functions
Reimplement https://godot-kotlin.readthedocs.io/en/latest/api-differences/#core-types

### LINKED ISSUE #147: Add a local copy check for CoreType in Idea plugin
Related to #146  

In Kotlin, similar to C#, we can't write something like:
`spatial.rotation.x = 2`

instead, you have to write:
`val rot = spatial.rotation`
`rot.x = 2`
`spatial.rotation = rot`

The reason is that rotation is a Kotlin side copy of the true native rotation property on the C++ side.
It can be misleading for a lot of users because they won't realize they are modifying a temporary value that won't affect the native value.

To fix this, adding an additional check in the Idea plugin that can throw a warning would let the user know their mistake.

I think it's better to have the plugin handle that instead of trying to fix that local copy in the first place. Doing so would be too complex and make the core types heavier.

chippmann:
This re adds all core type helper functions with accompanying documentation.
This also adds a first version of IDE checks for CoreType copy modifications:
  
Reference handling:

  
Different return branches handling:

Resolves #146 and resolves #147
Depends on: https://github.com/utopia-rise/godot-kotlin-api-generator/pull/9


kt/godot-runtime/src/main/kotlin/godot/annotation/CoreTypeHelper.kt
```diff
@@ -0,0 +1,10 @@
+package godot.annotation
+
+/**
+ * Only applicable to godot core type helper functions generated in the api gen
+ *
+ * Here for the IDE checks to differentiate mutation function calls (these helper functions) from regular function calls
+ */
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class CoreTypeHelper
```
raniejade, line 10: make this internal? I don't think this will be used by user code.

raniejade (approved):

kt/godot-runtime/src/main/kotlin/godot/annotation/CoreTypeHelper.kt
chippmann: Not possible. The generated code is in `godot-library` but the annotation is in `godot-runtime` because of the core types. I cannot have it internal. Then it's not visible in `godot-library` 🙈

piiertho (approved):

CedNaru (approved):

Issue #183, opened 2021-04-24
`getNode` extension function shadowed by member


chippmann:
Currently our generic `getNode` extension function is shadowed by the same not generic member function which leads to the IDE not suggesting the function and not providing imports.
The extension function should be renamed.

PR #187, opened 2021-04-24
Rename node helpers

CedNaru:
Extensions are now in their own package and I split them into 2 files, one for Object, one for Node.

Also because of a conflict with the "getNode" name, the helpers have been rename simple "node".

related issue : #183

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
+++ b/kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
@@ -1,0 +1,10 @@
+package godot.extensions
+
+import godot.Node
+import godot.core.NodePath
+
+@Suppress("NOTHING_TO_INLINE", "UNCHECKED_CAST")
+inline fun <T : Node> Node.node(path: String) = getNode(NodePath(path)) as T?
+
+@Suppress("NOTHING_TO_INLINE", "UNCHECKED_CAST")
+inline fun <T : Node> Node.node(nodePath: NodePath) = getNode(nodePath) as T?
--- a/kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
+++ b/kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
@@ -4,7 +4,7 @@ import godot.Node
 import godot.core.NodePath
 
 @Suppress("NOTHING_TO_INLINE", "UNCHECKED_CAST")
-inline fun <T : Node> Node.node(path: String) = getNode(NodePath(path)) as T?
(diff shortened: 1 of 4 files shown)


kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
```diff
@@ -0,0 +1,10 @@
+package godot.extensions
+
+import godot.Node
+import godot.core.NodePath
+
+@Suppress("NOTHING_TO_INLINE", "UNCHECKED_CAST")
+inline fun <T : Node> Node.node(path: String) = getNode(NodePath(path)) as T?
+
+@Suppress("NOTHING_TO_INLINE", "UNCHECKED_CAST")
+inline fun <T : Node> Node.node(nodePath: NodePath) = getNode(nodePath) as T?
```
chippmann, line 10: Why is the rename necessary? It works for me if they just have a different package path.

chippmann (changes requested):

kt/godot-library/src/main/kotlin/godot/extensions/NodeExt.kt
CedNaru: It works for getNode(String) but not for getNode(NodePath) in my case.

chippmann: Hmm it does for me

CedNaru: it's suggested but when I select it, it still doesn't work, the import is not made so I just get a red line.

raniejade: might be good to test out latest kotlin versions and enabling JVM IR - from what I understand the code still compiles but the only IDE marks it as red?

raniejade (approved):
I like the package change, but I'm not too sure about the rename.

raniejade:
I just did a test, moving to a different package only hides the issue - when you start using the helpers the problem can occur - so a rename is a must. I think we need a way to tell the helper is connected to `getNode`, WDYT about using something like `getNodeOfType` or maybe `getNodeAs`?

chippmann:
Hmm weird. I dont seem to have those issues. Anyways. Then I'm all in for either just `node` or `getNodeAs`.
The latter would even be more "normal" for generic functions i think

chippmann (approved):

raniejade:
`getNodeAs` is a bit a more descriptive since the helper justs calls `getNode` and attempts to cast it to the reified type.

CedNaru:
GetNodeAs seems nice, it creates a whole sentence that way `getNodeAs<Button>`

PR #188, opened 2021-04-25
Scripts created from Godot have a package set.

### LINKED ISSUE #77: Signal connect is broken
When connecting a signal from within a class extending `godot.Object` the `instance` is wrong. Instead of the object that owns the signal, a connection to the caller of the function is made as `this` in that context refers to the caller of the function rather than the owner of the signal.

Example:
```kotlin
@RegisterClass
class TestingClass: Spatial() {

	@RegisterProperty
	lateinit var buttonNodePath: NodePath
	private val button by lazy { getNode(buttonNodePath) as Button }

	@RegisterFunction
	override fun _ready() {
		button.signalPressed.connect(this, this::onButtonPressed)
	}

	@RegisterFunction
	fun onButtonPressed() {
		println("huhu")
	}
}
```

does not work as `this@Object` here:
```kotlin
  inline fun <reified K : () -> Unit> Signal0.connect(
    target: Object,
    method: K,
    binds: VariantArray<Any?>? = null,
    flags: Long = 0
  ) {
    val methodName = (method as KCallable<*>).name.camelToSnakeCase()
    connect(this@Object, target, methodName, binds, flags) // <- HERE
  }
```
will be the `TestingClass` instance rather than the `Button` instance.
So it tries to connect an inexistent signal `pressed` from itself to itself, rather than connect `pressed` from `Button` to itself.

---

This `button.connect("pressed", this, "on_button_pressed")` on the other hand will work as `connect` is explicitly called on the `Button` instance and not on the `TestingClass` instance.

---

Reproduction scene at `bugfix/signal_connection`

### LINKED ISSUE #177: Incorrect package name when creating a script using "Attach Script" dialog inside godot editor
```kotlin
package fixme // this should be a package name

import godot.Spatial
import godot.annotation.RegisterClass
import godot.annotation.RegisterFunction

@RegisterClass
class RootScene : Spatial() {

	// Declare member variables here. Examples:
	// val a = 2;
	// val b = "text";

	// Called when the node enters the scene tree for the first time.
	@RegisterFunction
	override fun _ready() {
		
	}

	// Called every frame. 'delta' is the elapsed time since the previous frame.
	@RegisterFunction
	override fun _process(delta: Double) {
		
	}
}
```

CedNaru:
Fix #177 

image

The package name is inferred from the file path of the newly created script.
For now, it only works if you the main source directory.
A package name will still be written if you use a custom source dir but it won't be correct. Unless we find a way to set the path of the custom dir in the C++ code I can't fix that particular case.

If the script is at the root of the source directory no package is written at all as well, as it should be.

image

### DIFF
--- a/src/kotlin_script.cpp
+++ b/src/kotlin_script.cpp
@@ -168,7 +168,6 @@ void KotlinScript::set_path(const String& p_path, bool p_take_over) {
     String package = p_path.replace("src/main/kotlin/", "")
             .trim_prefix("res://")
             .trim_suffix(get_name() + "." + KotlinLanguage::get_instance().get_extension())
-            .trim_prefix("/")
             .trim_suffix("/")
             .replace("/", ".");
 
@@ -181,7 +180,6 @@ void KotlinScript::set_path(const String& p_path, bool p_take_over) {
 
 #ifndef TOOLS_ENABLED
     if (!kotlin_class) {
-
         kotlin_class = GDKotlin::get_instance().find_class(p_path);
     }
 #endif
--- a/src/kotlin_script.cpp
+++ b/src/kotlin_script.cpp
@@ -165,6 +165,11 @@ Variant KotlinScript::_new(const Variant** p_args, int p_argcount, Variant::Call
 void KotlinScript::set_path(const String& p_path, bool p_take_over) {
     Resource::set_path(p_path, p_take_over);
 
+#ifndef TOOLS_ENABLED
+    if (!kotlin_class) {
+        kotlin_class = GDKotlin::get_instance().find_class(p_path);
+    }
+#else
     String package = p_path.replace("src/main/kotlin/", "")
             .trim_prefix("res://")
             .trim_suffix(get_name() + "." + KotlinLanguage::get_instance().get_extension())
@@ -177,11 +182,6 @@ void KotlinScript::set_path(const String& p_path, bool p_take_over) {
 
     String source_code = get_source_code().replace("%PACKAGE%", package);
     set_source_code(source_code);
-
-#ifndef TOOLS_ENABLED
-    if (!kotlin_class) {
-        kotlin_class = GDKotlin::get_instance().find_class(p_path);
-    }
 #endif
 }
 
--- a/src/kotlin_script.cpp
+++ b/src/kotlin_script.cpp
@@ -170,11 +170,11 @@ void KotlinScript::set_path(const String& p_path, bool p_take_over) {
         kotlin_class = GDKotlin::get_instance().find_class(p_path);
     }
 #else
-    String package = p_path.replace("src/main/kotlin/", "")
-            .trim_prefix("res://")
-            .trim_suffix(get_name() + "." + KotlinLanguage::get_instance().get_extension())
-            .trim_suffix("/")
-            .replace("/", ".");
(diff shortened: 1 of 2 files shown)


chippmann:
Hmm IIRC we even had it once that i generate the used source dirs to the entry file and we read them at startup. IIRC this was before we were using the full res path for loading.
We scrapt that for perf resons but this time its a editor only thing so no issues. I suggest i generate a list in the entry file with all used src dirs in it. Then we query it at startup and you can use it in cpp to check it. WDYT?

CedNaru:
But if you do it by using the entry file, you will only find a source directory if there is already a script in it.
So let's say you create a new empty source directory and write the correct statement in the build. gradle. Because it's empty, no entry file will be generated for it.  So the package will be wrong for the very first script you create there.

chippmann:
True but to be honest, most of the times creation of kotlin files should be done inside the IDE IMO.
It would add an unnecessary level of complexity to catch all cases. And with this we at least have some cases covered.

chippmann:
But I'm also ok with adding this in a separate PR if we want to add this

CedNaru:
Well then we can go with the PR in its current state.

chippmann (approved):

raniejade (approved):
We probably don't need the catch all the cases here - I think at least it should work with the default configuration (i.e no default source dirs).

If really want to implement that feature, we probably can query gradle for the configured custom source dir - but that can be costly specially if you are just creating a file.

src/kotlin_script.cpp
```diff
@@ -164,10 +164,24 @@ Variant KotlinScript::_new(const Variant** p_args, int p_argcount, Variant::Call
 
 void KotlinScript::set_path(const String& p_path, bool p_take_over) {
     Resource::set_path(p_path, p_take_over);
+
 #ifndef TOOLS_ENABLED
     if (!kotlin_class) {
         kotlin_class = GDKotlin::get_instance().find_class(p_path);
     }
+#else
+    String package = p_path.replace("src/main/kotlin/", "")
```
piiertho, line 173: Cpp uniform init can be used here :)

src/kotlin_script.cpp
```diff
@@ -178,7 +192,7 @@ KotlinScript::KotlinScript() : kotlin_class(nullptr) {
 PlaceHolderScriptInstance* KotlinScript::placeholder_instance_create(Object* p_this) {
 #ifdef TOOLS_ENABLED
     PlaceHolderScriptInstance* placeholder{
-        memnew(PlaceHolderScriptInstance(&KotlinLanguage::get_instance(), Ref<Script>(this), p_this))
+            memnew(PlaceHolderScriptInstance(&KotlinLanguage::get_instance(), Ref<Script>(this), p_this))
```
piiertho, line 195: auto format ?

piiertho (approved):

src/kotlin_script.cpp
CedNaru: yes

chippmann:
> We probably don't need the catch all the cases here - I think at least it should work with the default configuration (i.e no default source dirs).
> 
> If really want to implement that feature, we probably can query gradle for the configured custom source dir - but that can be costly specially if you are just creating a file.

Yeah that would be too costly. Especially when you first have to start a gradle deamon

PR #196, opened 2021-04-25
Fix module wizard submodule handling

### LINKED ISSUE #179: [IDEA] Generating a project as a submodule results in putting all files in a root project directory instead of inside the submodule
**Steps:**
```
GIVEN I have a random project opened in IDEA, for example "D:\Repositories\random_project"
AND I click right click on a root project directory name
THEN I click New -> Module
AND I select "Godot Kotlin JVM"
AND I click "Next" on a dialog creation window
AND I make sure that the "Content root" and "Module file location" are a nested directories inside current project, for example: "D:\Repositories\random_project\client-jvm"
THEN I click "Finish"
```

**Expected result:**
All the files created inside "D:\Repositories\random_project\client-jvm"

**Actual result:**
All the project files are not inside "D:\Repositories\random_project\client-jvm", but rather "D:\Repositories\random_project"

chippmann:
Resolves #179

### DIFF
--- a/kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/module/GodotModuleBuilder.kt
+++ b/kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/module/GodotModuleBuilder.kt
@@ -16,6 +16,7 @@ import com.intellij.ui.layout.GrowPolicy
 import com.intellij.ui.layout.panel
 import com.intellij.ui.layout.selected
 import godot.intellij.plugin.GodotPluginBundle
+import godot.utils.GodotBuildProperties
 import org.jetbrains.plugins.gradle.util.GradleUtil
 import java.io.File
 import javax.swing.JCheckBox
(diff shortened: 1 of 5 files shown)


kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/module/GodotModuleBuilder.kt
```diff
@@ -114,36 +118,86 @@ class GodotModuleBuilder : ModuleBuilder(), ModuleBuilderListener {
...
+                        .let { content ->
+                            if (module.parentProjectAlreadyContainsDependency(wizardContext, "godot-library")) {
+                                content.replace(
+                                    "GODOT_KOTLIN_DEPENDENCY",
+                                    "id(\"com.utopia-rise.godot-kotlin-jvm\")"
+                                )
+                            } else {
+                                content.replace(
+                                    "GODOT_KOTLIN_DEPENDENCY",
+                                    "id(\"com.utopia-rise.godot-kotlin-jvm\") version \"0.1.4-3.2.3\""
```
CedNaru, line 184: Isn't that error pron to hardcode the version here ? We might forget it when we bump the version?

chippmann: I had that once but removed it again because the build props were causing an issue in the ide plugin. Don't remember what. I'll try again

raniejade (approved):
LGTM pending the removal of the embedded version.

piiertho (approved):

CedNaru (approved):

PR #197, opened 2021-04-25
GodotStatic QOL and Test

CedNaru:
Added a GodotStatic abstract class so we don't need to manually do the registering.
Also added a singleton to the test to see if there is no leak when are done with the tests.

Small fix to isInstanceValid as well.


piiertho (approved):

docs/src/doc/user-guide/advanced/kotlin-singleton.md
```diff
@@ -1,7 +1,24 @@
-Sometimes you need to store some Godot objects or references in a Kotlin singleton. This can cause some leaks when the program ends because you have to manually free objects and have the references collected. The `GodotStatic` interface lets you implement the function `collect` so you can write the appropriate cleanup code. Then you just need to call `registerAsSingleton` so it will be hooked to the Garbage Collector. 
+Sometimes, you need to store some Godot objects or references in a Kotlin singleton.
+This can cause some leaks when the program ends because you have to manually free objects and have the references collected.
+The `GodotStatic` abstract class lets you implement the function `collect` so you can write the appropriate cleanup code.
```
chippmann, line 3: Ranie is going to kill you with the line breaks xD
He explicitly removed them xD

docs/src/doc/user-guide/advanced/kotlin-singleton.md
```diff
@@ -1,7 +1,24 @@
...
+
+    override fun collect(){
+        obj.free()
+        ref = null
+    }
+}
+```
+
+You can also directly implement the interface `IGodotStatic` but you need to do the registering yourself.
+You just need to use the `registerAsSingleton` so it will be hooked to the Garbage Collector.
```
chippmann, line 18: Same here

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -102,8 +103,10 @@ object GarbageCollector {
         nativeCoreTypeMap[rawPtr] = NativeCoreWeakReference(nativeCoreType, nativeReferenceQueue, variantType)
     }
 
-    fun registerStatic(instance: GodotStatic) {
-        staticInstances.add(instance)
+    fun registerStatic(instance: IGodotStatic) {
+        if(instance !in staticInstances){
+            staticInstances.add(instance)
+        }
```
chippmann, line 109: Why not just a set then?

chippmann (changes requested):

chippmann (approved):

docs/src/doc/user-guide/advanced/kotlin-singleton.md
```diff
@@ -1,7 +1,21 @@
-Sometimes you need to store some Godot objects or references in a Kotlin singleton. This can cause some leaks when the program ends because you have to manually free objects and have the references collected. The `GodotStatic` interface lets you implement the function `collect` so you can write the appropriate cleanup code. Then you just need to call `registerAsSingleton` so it will be hooked to the Garbage Collector. 
+Sometimes, you need to store some Godot objects or references in a Kotlin singleton. This can cause some leaks when the program ends because you have to manually free objects and have the references collected. The `GodotStatic` abstract class lets you implement the function `collect` so you can write the appropriate cleanup code.
```
raniejade, line 1:
```suggestion
Sometimes you need to store some Godot objects or references in a Kotlin singleton, this can cause some leaks when the program ends because you have to manually free objects and have the references collected. The `GodotStatic` abstract class lets you implement the function `collect` so you can write the appropriate cleanup code.
```

raniejade (changes requested):
I'm not fond of the `I` prefix in the interface. One idea is to do:

```kt
interface GodotStatic {
  ...
}

abstract class BaseGodotStatic() : GodotStatic { // or maybe AbstractGodotStatic
  ...
}
```

raniejade (changes requested):
I'm not fond having more than one way of doing the same thing - can we just stick to one? I think the delegate approach is much simpler and users don't have to worry about manually writing the `collect` method. Since `GodotStatic` is the heart of the implementation, we can just make this private perhaps?

piiertho:
Why is `kt/godot-runtime/out/production/classes/godot/util/UtilsKt.class` added ?

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
```diff
@@ -254,9 +254,16 @@ object GarbageCollector {
...
     fun cleanUp() {
-        for (instance in staticInstances) {
-            instance.collect()
+        info("Start collecting static instances")
+        while(staticInstances.size > 0){
+            val iterator = staticInstances.iterator()
+            staticInstances = mutableSetOf()
+            for (instance in iterator) {
+                println("Collect ${instance::class}")
+                instance.collect()
```
piiertho, line 263: We use GD singleton to check if instance is valid. What would happen if GD singleton is collected before other singleton ?

piiertho:
> I'm not fond having more than one way of doing the same thing - can we just stick to one? I think the delegate approach is much simpler and users don't have to worry about manually writing the `collect` method. Since `GodotStatic` is the heart of the implementation, we can just make this private perhaps?

I agree, if we can have only one way, it will be easier for user and for us.

kt/godot-runtime/src/main/kotlin/godot/core/GarbageCollector.kt
CedNaru: No longer an issue with the new delegate system.

chippmann (approved):

docs/src/doc/user-guide/advanced/kotlin-singleton.md
```diff
@@ -1,17 +1,14 @@
-Sometimes you need to store some Godot objects or references in a Kotlin singleton. This can cause some leaks when the program ends because you have to manually free objects and have the references collected. The `GodotStatic` interface lets you implement the function `collect` so you can write the appropriate cleanup code. Then you just need to call `registerAsSingleton` so it will be hooked to the Garbage Collector. 
+Sometimes you need to store some Godot objects or references in a Kotlin singleton. This can cause some leaks when the program ends because you have to manually free objects and have the references collected. 
+This issue is fixed by using the delegate `godotStatic` on singleton properties. Those properties will be freed once the JVM ends. They automatically handle `Object` and `Reference`. You can also freely set a new value and the previous one will be immediatly freed.
+Warning: Only use it on a singleton, otherwise all the properties of all instances are going to be kept alive until the end of the JVM.
```
raniejade, line 3:
```suggestion
!!! warning
	Only use it on a singleton, otherwise all the properties of all instances are going to be kept alive until the end of the JVM.
```

nit: might be good to use an warning block here:

raniejade (approved):

PR #199, opened 2021-04-26
Fix: generate right api for godot 3.3

piiertho:
I messed up with api json when updating to 3.3
I regenerated api from 3.3.
This needs https://github.com/utopia-rise/godot-kotlin-api-generator/pull/11


CedNaru (approved):

chippmann:
@piiertho are you sure this is correct? Because this removes `AspectRatioContainer.kt`. But i thought it was introduced in `3.3`?

kt/godot-library/src/main/kotlin/godot/gen/godot/Area.kt
```diff
@@ -66,7 +66,7 @@ open class Area : CollisionObject() {
    * `local_shape` the index of the [godot.Shape] of this Area used by the [godot.PhysicsServer].
    */
   val areaShapeEntered: Signal4<Long, Area, Long, Long> by signal("area_id", "area", "area_shape",
-      "local_shape")
+      "self_shape")
```
chippmann, line 69: also in the current class ref these params are named `local_shape`
https://docs.godotengine.org/en/3.3/classes/class_area.html#signals
So i think the generation is not correct yet.

piiertho: it seems so, le me re dot it

chippmann (approved):

PR #213, opened 2021-04-30
Remove 0 at end of strings.

### LINKED ISSUE #212: Strings from godot in kotlin ends with zero char
Hello, 

I found a bug with strings which we receive from Godot side:

image

image

This is not specific for Dictionary. I found the same problem with file names while iterating by files in directory:

```
val dir = Directory()
dir.open("res://some_dir")
dir.listDirBegin(skipNavigational = true, skipHidden = true)
var file = dir.getNext() // file name also has zero at the end
while (file != "") {
   ...
   file = dir.getNext()
}
```

CedNaru:
fix #212


piiertho (approved):

raniejade (approved):

raniejade:
Can we add some tests for this?

PR #222, opened 2021-05-02
Feature/embedd api generator

### LINKED ISSUE #160: Include api generator in project
Related to #154 .

piiertho:
This resolves #160 by including api generator as include build but in main repo.
This also closes #199 in favor of this one.


CedNaru (approved):

raniejade (approved):
Do we still need to publish the api generator? The only time we need to use it is to build godot-library, can it be used maybe as an included build? I just don't see value in publishing it since it's part of the main project now and it will only ever be used internally.

CedNaru:
For now it's okay but later people will need it if they want to generate their own API because they have a Godot with different modules than the official. And the people with a custom Godot are not necessary C++ dev, just regular game designers who were given a custom version.

raniejade:
They have to pull in this repo either way right? i.e They need to rebuild godot-library.

piiertho:
Yes they would have to deploy their own artifacts. This is a job for dev team, not game designers.

raniejade:
Why do we need to publish the api generator then? i.e The uses of the api generator is within this repo. I don't quite follow how game designers vs game developers apply here.

raniejade:
Adding a bit more context, we publish the entry generator because it is used in the a users project. Now, with the api generator - it's not gonna be used in users projects but only within this repo. So, why do we need to publish it when in fact it is an internal tool?

piiertho:
> Adding a bit more context, we publish the entry generator because it is used in the a users project. Now, with the api generator - it's not gonna be used in users projects but only within this repo. So, why do we need to publish it when in fact it is an internal tool?

I did not say we need, I just did not removed it now, I agree with you.

piiertho:
@raniejade, we were not deploying it. I have no artifact on sonatype and no job in CI.

chippmann (approved):

PR #228, opened 2021-05-07
Fix: Harcode godot version in CI as it is inconsistency

piiertho:
Godot version is inconsistent, this cause trouble to deploy new release.
Quick fix is to hardcode version in CI.


chippmann (approved):
Ok for this release but we need an issue to fix it properly.
The reason we did it so "complicated" is so we can never forget to bump the version code in every place because we only have one.
It should be "as simple" as just changing the regexes in the workflows.

PR #241, opened 2021-06-10
fix: Getter now calls method from parent even if getter starts with is

### LINKED ISSUE #216: `InputEventMouseButton` has no `pressed` property getter
Well, it has, but it looks like this:

```kotlin
open var pressed: Boolean
	  get() {
	    throw UninitializedPropertyAccessException("Cannot access property pressed: has no getter")
	  }
```

piiertho:
Generated getters were not calling parent method if method was starting with `is`.
This resolves #216

### DIFF
--- a/kt/api-generator/src/main/kotlin/godot/codegen/generationEntry.kt
+++ b/kt/api-generator/src/main/kotlin/godot/codegen/generationEntry.kt
@@ -34,17 +34,35 @@ fun File.generateApiFrom(jsonSource: File, docsDir: File? = null) {
 
     classes.forEach { clazz ->
         clazz.properties.forEach { property ->
-            val method = Method(
-                property.getter.convertToSnakeCase(),
-                property.type,
-                isVirtual = false,
-                hasVarargs = false,
-                arguments = listOf()
-            )
-            val parentClassAndMethod = tree.getMethodFromAncestor(clazz, method)
-            if (parentClassAndMethod != null && !property.hasValidGetter) {
-                property.shouldRenameJvmProperty = true
+            fun inferMethodAccessorFromParent(isSetter: Boolean = false) {
+                val methodName = if (isSetter) property.setter else property.getter
+                val returnType = if (isSetter) "void" else property.type
+                val arguments = if (isSetter) listOf(Argument(property.oldName, property.type)) else listOf()
+
+                val method = Method(
+                    methodName.convertToSnakeCase(),
+                    returnType,
+                    isVirtual = false,
+                    hasVarargs = false,
+                    arguments = arguments
+                )
+
+                val parentClassAndMethod = tree.getMethodFromAncestor(clazz, method)
+                val hasValidAccessor = if (isSetter) property.hasValidSetter else property.hasValidGetter
+                if (parentClassAndMethod != null && !hasValidAccessor) {
+                    if (isSetter) {
+                        property.shouldUseSuperSetter = true
+                    } else {
+                        property.shouldUseSuperGetter = true
+                    }
+                }
             }
+
+            inferMethodAccessorFromParent()
+
+            // It does not seems to have any case where setter should call parent in api. But in case this happen in
+            // future, this is here.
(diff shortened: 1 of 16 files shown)


kt/api-generator/src/main/kotlin/godot/codegen/generationEntry.kt
```diff
@@ -35,17 +36,18 @@ fun File.generateApiFrom(jsonSource: File, docsDir: File? = null) {
...
                 property.type,
                 isVirtual = false,
                 hasVarargs = false,
                 arguments = listOf()
             )
             val parentClassAndMethod = tree.getMethodFromAncestor(clazz, method)
             if (parentClassAndMethod != null && !property.hasValidGetter) {
-                val parentMethodName = "get${parentClassAndMethod.first.newName}${property.newName.capitalize()}"
+                val isBoolGetter = property.getter.startsWith("is")
+                val parentMethodName = "${if (isBoolGetter) "is" else "get"}${parentClassAndMethod.first.newName}${property.newName.capitalize()}"
```
chippmann, line 48: I would separate the property name from the parent method call in the getter. Because you share this variable for both property name and parent method call the properties names are quite redundant.
Example (see in the generated files in this PR):
`isFontDistanceField` instead of `isDistanceFieldHint` inside the `Font` class. `Font` in the property name is redundant.
`isInputEventEcho` instead of `isEcho` inside the `InputEvent` class. `InputEvent` in the property name is redundant.

kt/godot-library/src/main/kotlin/godot/gen/godot/Font.kt
```diff
@@ -141,7 +141,7 @@ open class Font : Resource() {
   /**
    *
    */
-  open fun isDistanceFieldHint(): Boolean {
+  open fun isFontDistanceField(): Boolean {
```
chippmann, line 144: Weird property name because of comment above

kt/godot-library/src/main/kotlin/godot/gen/godot/InputEvent.kt
```diff
@@ -122,7 +122,7 @@ open class InputEvent : Resource() {
   /**
    * Returns `true` if this input event is an echo event (only for events of type [godot.InputEventKey]).
    */
-  open fun isEcho(): Boolean {
+  open fun isInputEventEcho(): Boolean {
```
chippmann, line 125: Same

kt/godot-library/src/main/kotlin/godot/gen/godot/InputEvent.kt
```diff
@@ -131,7 +131,7 @@ open class InputEvent : Resource() {
   /**
    * Returns `true` if this input event is pressed. Not relevant for events of type [godot.InputEventMouseMotion] or [godot.InputEventScreenDrag].
    */
-  open fun isPressed(): Boolean {
+  open fun isInputEventPressed(): Boolean {
```
chippmann, line 134: And so on. Will not comment the other appearances :-)

chippmann (changes requested):

kt/api-generator/src/main/kotlin/godot/codegen/generationEntry.kt
piiertho: This redundancy is desired, because if we do not change the name of parent getter we end up with conflicts between getter of properties and get methods from parent in api. 
Is is a particular case where this would not occur. But this happens with variable starting with get :/

piiertho: @CedNaru gave me a better idea. I'll use @JvmName (which I forgot it exists as I do not write java anymore xD)

CedNaru:
@JvmName should be your friend.

chippmann (approved):
Yeah with `@JvmName` it's fine now :-)

Edit: hoh `@`JvmName is a user on github xD i guess he'll get's mentioned a lot because of this xD

CedNaru (approved):

PR #244, opened 2021-06-23
Implement KSP based entry gen

chippmann:
# TL;DR
Reimplementation of the entry gen. 
Compatible with Kotlin 1.5.x.
Similar feature set as the previous entry gen but not yet fully completed.
Same input/output as the previous entry gen.
No changes for the user.

# General
This PR is the first step of reimplementing the entry gen to be more robust against compiler api changes as well as general abstraction and ease of use (with keeping in mind eventual future support of other languages, or at least make it simpler to implement).
It is not the full reimplementation. I first focused on feature parity and Kotlin 1.5.x support so the graalvm implementation of iOS is not blocked anymore. Also to keep the changeset a bit smaller. Stuff i want to change/add in subsequent PR's are:
- [ ] Complete language independence (some extension functions and hardcoded FqNames are still bound to the kotlin language).
- [ ] Moving sanity checks into own module and abstract them further to be able to share them between the Intellij Plugin and the entry gen
- [ ] Add the possibility of pluggable entry gen extensions (not the full implementation yet, we have more important stuff to do first)

# Why KSP (Kotlin Symbol Processing API)
This reimplementation of the entry gen is based on [KSP (Kotlin Symbol Processing API)](https://github.com/google/ksp).

The usage of KSP over [mpapt](https://github.com/Foso/MpApt) has the advantage of a cleaner api and the abstraction of compiler classes. Which makes our own implementation more robust against compiler api changes as these are handled by KSP.

It also has the capability of detecting generated sources by itself (with correct configuration) and triggering a new compilation round including the new sources. This basically means we're finally getting rid of the additional dummy compilation's and can compile the users code in one go.

# Rough overview
The following modules are removed and not needed anymore:
- `godot-kotlin-compiler-plugin`
- `godot-kotlin-compiler-plugin-common`

The following modules are renamed/reimplemented:
- `godot-annotation-processor` -> `godot-kotlin-symbol-processor`
- `godot-kotlin-entry-generator` -> `godot-entry-generator`

## godot-gradle-plugin
The removal of `godot-kotlin-compiler-plugin` and `godot-kotlin-compiler-plugin-common` means that our plugin no longer needs to be a `KotlinCompilerPluginSupportPlugin`. It can be a normal `Plugin<Project>` which makes the configuration simpler.
Also all the additional configurations for `main` and `game` are removed as everything can be compiled in one go and we don't need multiple compilation targets anymore.
The plugin adds all additional repos, plugins and extension configurations for the following new plugins and dependencies we automatically add to make the entry generation work:
- KSP gradle plugin -> replaces our own compiler plugin
- KSP dependency -> needed for the gradle plugin above
- KSP extension configuration -> replaces the configuration for our own compiler plugin
- idea gradle plugin -> to mark generated sources as such so intellij recognises them
- google() repository -> needed as KSP is only hosted on the google repositories and not on mavenCentral

## godot-entry-generator
Only contains the actual entry generation logic (and for now the sanity checks). It provides a set of model classes which it expects to receive when called. It no longer directly interacts with compiler classes or any other classes from external dependencies anymore (except kotlin poet).
To achieve this, the default value extraction, as well as the compiler information gathering are offloaded to the caller of the entry gen. For kotlin this is `godot-kotlin-symbol-processor`.
The sole responsibility of the entry gen is to actually generate the entry sources based on the source abstraction (and currently the sanity checks but they'll get extracted as well in the near future as written earlier).

## godot-kotlin-symbol-processor
The `godot-kotlin-symbol-processor` is the one actually responsible for gathering the source code information and for default value extraction.
It uses KSP to gather all needed information about the compiled classes and maps only the needed information into the model classes of the entry gen.
It also provides the file output streams for each generated class to the entry gen, so ksp can keep track of which files are generated and if they need to be invalidated based on the source code changes.

With this abstraction and the further abstractions planned, the only thing which needs to be reimplemented to support additional languages (as far as entry generation goes) is the `godot-kotlin-symbol-processor`.


piiertho (changes requested):
It seems some conversions methods we use are now deprecated:
```
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 42): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 56): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (747, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (749, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (753, 46): 'toChar(): Char' is deprecated. Direct conversion to Char is deprecated. Use toInt().toChar() or Char constructor instead.
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (185, 50): Unchecked cast: List<P> to L
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (781, 9): Parameter 'isTool' is never used
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\util\Utils.kt: (17, 7): 'toLowerCase(): String' is deprecated. Use lowercase() instead.
```
```
> Task :kt:godot-library:compileKotlin
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\extensions\PackedSceneExt.kt: (7, 46): Parameter 'editState' is never used
w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\global\GDCore.kt: (214, 24): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
```

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import godot.entrygenerator.model.RegisteredProperty
+
+
+class EnumFlagHintStringGenerator(
+    registeredProperty: RegisteredProperty
+) : PropertyHintStringGenerator(registeredProperty) {
+
+    override fun getHintString(): String =
+        throw UnsupportedOperationException("Hint string for enums is handled by the binding at runtime.")
```
CedNaru, line 11: Isn't that weird to have a hint that will always throw an error? It's kinda confusing for the users.
It's like "Ho yeah we support this feature but don't call it explicitly".

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGenerator.kt
```diff
@@ -0,0 +1,15 @@
+package godot.entrygenerator.generator.hintstring
+
+import godot.entrygenerator.model.PropertyHintAnnotation
+import godot.entrygenerator.model.RegisteredProperty
+
+abstract class PropertyHintStringGenerator(
+    val registeredProperty: RegisteredProperty
+) {
+    protected val propertyHintAnnotation = registeredProperty.annotations.firstOrNull { it is PropertyHintAnnotation } as PropertyHintAnnotation?
```
CedNaru, line 9: Are we sure there is no case of several HintAnnotations needed?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGeneratorProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+import godot.entrygenerator.model.RegisteredProperty
+import godot.entrygenerator.model.TypeKind
+
+
+object PropertyHintStringGeneratorProvider {
+
+    fun provide(
+        registeredProperty: RegisteredProperty
+    ): PropertyHintStringGenerator {
+        return when {
```
CedNaru, line 19: Wouldn't that be simpler for future hint string additions to iterate on the list of annotation and do the 'when' on the annotation class itself instead of the class of the property?
That way we can manage the when blocks on a per annotation basis so we don't have the same annotation repeated several times across the file.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/typehint/PropertyTypeHintGenerator.kt
```diff
@@ -0,0 +1,12 @@
+package godot.entrygenerator.generator.typehint
+
+import com.squareup.kotlinpoet.ClassName
+import godot.entrygenerator.model.PropertyHintAnnotation
+import godot.entrygenerator.model.RegisteredProperty
+
+abstract class PropertyTypeHintGenerator(
+    registeredProperty: RegisteredProperty
+) {
+    protected val propertyHintAnnotation = registeredProperty.annotations.firstOrNull { it is PropertyHintAnnotation } as PropertyHintAnnotation?
```
CedNaru, line 10: Same as stated earlier

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/typehint/PropertyTypeHintProvider.kt
```diff
@@ -0,0 +1,71 @@
...
+import godot.entrygenerator.model.PlaceHolderTextHintAnnotation
+import godot.entrygenerator.model.RegisteredProperty
+import godot.entrygenerator.model.TypeKind
+
+object PropertyTypeHintProvider {
+
+    fun provide(
+        registeredProperty: RegisteredProperty
+    ): ClassName {
+        return when {
```
CedNaru, line 22: same as stated before

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Clazz.kt
```diff
@@ -0,0 +1,22 @@
+package godot.entrygenerator.model
+
+import godot.entrygenerator.ext.hasAnnotation
+
+open class Clazz(
+    open val fqName: String,
+    open val supertypes: List<Clazz> = emptyList(),
```
CedNaru, line 7: I guess it contains the list of ancestors in order from child to parent.
Is that to avoid the use of a recursive function in the case you only store the direct parent?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
```diff
@@ -0,0 +1,13 @@
+package godot.entrygenerator.model
+
+data class RegisteredProperty(
+    val fqName: String,
+    val type: Type,
+    val isMutable: Boolean = true,
```
CedNaru, line 6: Nice we are even prepared to the case Godot add read-only properties
(Sadly they are against the keyword "val" as it's too close to "var" :( )

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
```diff
@@ -0,0 +1,13 @@
...
+    val isLateinit: Boolean = false,
+    val isOverridee: Boolean = false,
```
CedNaru, line 8: typo or my broken English ?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Type.kt
```diff
@@ -0,0 +1,19 @@
...
+
+data class Type(
+    val fqName: String,
+    val kind: TypeKind,
+    val isNullable: Boolean,
+    val supertypes: List<Type>,
+    val arguments: () -> List<Type>
+)
+
+enum class TypeKind {
```
CedNaru, line 11: Are some of those types just in case of future Godot functionalities?

For example, I hardly see how a type alias can be used for entry gen for now.

CedNaru (commented):
Suggestion:
Wouldn't that be better to have the model part of the entry generation as its own module and then use it as a dependency for the generator?
People wanting to implement other languages will only need to import that model module when implementing their own processor instead of the whole entry generator.
Given that model is kinda the bridge between the source language and the actual entry file. It seems natural to me that it's its own module that both sides import as a dependency.

Example: With the new Godot extension system in 4.0, we might need to handle 2 types of entry files, one for scripts, one for extensions.
We will be able to use the same Kotlin source processor for both, but not the same generator will be used.

CedNaru:
Finally done !
Overall it's amazing. Responsibilities are separated way more than before. The intermediate model you created is very simple to understand.
Separating the source code processor from the rest lets us enjoy the rest of the code without having the complexity of whatever library we use weights on everything else.
I was always lost when I was reading cod related to the previous entry generator but now I feel that even I would be able to do maintenance on it.

chippmann:
> It seems some conversions methods we use are now deprecated:
> 
> ```
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 42): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 56): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (747, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (749, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (753, 46): 'toChar(): Char' is deprecated. Direct conversion to Char is deprecated. Use toInt().toChar() or Char constructor instead.
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (185, 50): Unchecked cast: List<P> to L
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (781, 9): Parameter 'isTool' is never used
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\util\Utils.kt: (17, 7): 'toLowerCase(): String' is deprecated. Use lowercase() instead.
> ```
> 
> ```
> > Task :kt:godot-library:compileKotlin
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\extensions\PackedSceneExt.kt: (7, 46): Parameter 'editState' is never used
> w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\global\GDCore.kt: (214, 24): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> ```

I suggest that I'll update usages related to the entry gen in this PR and all other occurrences should be done in a separate PR.
If you are ok with it I would create the corresponding issue.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
chippmann: This error should never get thrown in the first place. If the entry gen code works correctly, it should generate different code for enums which should never rely on hint string generators in the first place. These are generated at runtime here: https://github.com/utopia-rise/godot-kotlin-jvm/blob/ed18f34f64316d8e0c022eccca551531dff3f043/kt/godot-runtime/src/main/kotlin/godot/runtime/Registration.kt#L142
This exception is just here so that it fails fast if we change something in the entry gen and not handle enums correctly anymore.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGenerator.kt
chippmann: Yes. The code in the inspector which handles hint strings (https://github.com/godotengine/godot/blob/4fcc5891450a7dca9c8ebae6ff3765f5f060c442/editor/editor_properties.cpp#L2701) only handles one and this one according to the property hint: https://github.com/godotengine/godot/blob/dcd11faad3802679a43b27155f1b6bc59aa39b60/core/object.h#L56.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
CedNaru: But it still seems weird to not use the GetHintString of a generator, just to generate it in another way inside the registration code.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGeneratorProvider.kt
chippmann: Hmm not sure if possible. For some hint strings i need information about the property itself not only the annotation.
For example for the `ArrayHintStringGenerator`: https://github.com/utopia-rise/godot-kotlin-jvm/blob/37914fc35e6649c08f6dec120f06296a2c33a99a/kt/entry-generation/godot-kotlin-entry-generator/src/main/kotlin/godot/entrygenerator/generator/property/hintstring/ArrayHintStringGenerator.kt#L11

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Clazz.kt
chippmann: Yes. As i would need such a recursive function to be executed at multiple places.
I thought about making it lazy like I've done for stuff like the default value extraction, but decided that it was not worth it as i already have that information in the `godot-kotlin-symbol-processor`.
But if you think a recursive function makes more sense, we can discuss this of course

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
chippmann: Yeah but the main reason was that i need that information for the export check: https://github.com/utopia-rise/godot-kotlin-jvm/blob/ed18f34f64316d8e0c022eccca551531dff3f043/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/ExportedMutablilityCheck.kt#L14

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
chippmann: Actually probably both of us have broken english xD
I wondered that as well. But ksp names it that way and IIRC in PSI there is also a property or function that names it like that. An alternative name in this case would probably be `isOverriding` or something like that.
I'm open for suggestions :-)

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGeneratorProvider.kt
CedNaru: Ho you would still have the same data available I was just talking about how you currently select the right genererator:
-Check the type of the property
-Check if  it owns one of the annotations
-If yes return the right generator

I was proposing the reverse:
- Check if  it owns one of the annotations
- Check the type.
- If yes, return the right generator.

It's just so everything related to one particular annotation is handled in the same place.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
chippmann: Maybe a list of modifiers would actually make more sense. So we can replace `isMutable`, `isLateinit` and `isOverridee` with a property which is a list of modifiers.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Type.kt
chippmann: The idea was so i can check whether a type is a type alias and get the underlying type if necessary. IIRC I'll need this for the intellij plugin to properly resolve if a type is allowed as a default value.
But other than that this could be removed i think.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Clazz.kt
CedNaru: No, no. It was just to be sure. We already got stackoverflow errors because of recursive functions, so it's nice to avoid them.
I was just wondering if it wasn't something like storing everything the class inherits/implements (so also interfaces).

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
CedNaru: I guess it's similar to cases where we write Clazz

chippmann:
> Suggestion:
> Wouldn't that be better to have the model part of the entry generation as its own module and then use it as a dependency for the generator?
> People wanting to add another language will only need to import that model module when implementing their own processor instead of the whole entry generator.
> Given that model is kinda the bridge between the source language and the actual entry file. It seems natural to me that it's its own module that both sides import as a dependency.
> 
> Example: With the new Godot extension system in 4.0, we might need to handle 2 types of entry files, one for scripts, one for extensions.
> We will be able to use the same Kotlin source processor for both, but not the same generator will be used.

Yes that makes more sense and is actually what i want to do when extracting the checks as well.
Regarding the separate entry gen; I don't think it will be necessary to have the entry gen separated when handling the different script types. I think this can all be done in one. But we'll see once we are there

chippmann:
> Finally done!
> Overall it's amazing. Responsibilities are separated way more than before. The intermediate model you created is very simple to understand.
> Separating the source code processor from the rest lets us enjoy the rest of the code without having the complexity of whatever library we use weights on everything else.
> I was always lost when I was reading cod related to the previous entry generator but now I feel that even I would be able to do maintenance on it.
> 
> I, when I started reading and saw the number of files:
> https://www.youtube.com/watch?v=a2Ku1KSUqUo

Thx :-)
And dafuq? xD

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
chippmann: I had to earlier because it was not possible to just get the enum values the way our kotlin compiler plugin was set up before. We were also getting the properties and functions which might be defined inside the enum class.
So a reliable getting of just the enum values was not possible (at least in a easy to understand and maintainable way).
Not sure if this is still the case for ksp though. I'll look into it. Maybe we can merge the enumProperty and the normal property registration again.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGeneratorProvider.kt
chippmann: ah. Now i understand. Yeah that makes more sense. I'll look into it.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Type.kt
CedNaru: Ho so it's not automatically resolved to the actual underlying type? I thought it was pure syntactic sugar without any technical effect.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Clazz.kt
chippmann: No it does not contain interfaces as currently we don't really care about them: https://github.com/utopia-rise/godot-kotlin-jvm/blob/ed18f34f64316d8e0c022eccca551531dff3f043/kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/ext/ksClassDeclarationExt.kt#L45

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
chippmann: WDYM?

CedNaru: Well, we don't write  "Class" directly to name stuff because it's already taken as a keyword, that's why we write "Clazz" instead. 
So I was wondering if they were using "Overridee" for the same reason.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
chippmann: Ok it is possible but I'm not sure if it's worth it.
Because to be able to implement it i would need to add another callback param to the entry gen entry point here: https://github.com/utopia-rise/godot-kotlin-jvm/blob/7f979104bae5123a8a83ba99ade4073ac766a362/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/EntryGenerator.kt#L20
The purpose of the callback would be to resolve a `Type` to a `Clazz` so we can read the properties, functions and in the case of an enum; subclasses which are the enum values.
So the `godot-kotlin-symbol-processor` would need to keep track of the mapping of KSP types it mapped to entry gen `Type` so it can do that mapping when requested through the callback.
All of that is not complicated and shouldn't take too long to implement, but I'm not sure if it's worth it just for this case.
I'd rather just remove this hintstring provider entirely if it bothers you but if you think otherwise I'm fine with implementing this resolve function as well.

CedNaru: Can't you store the information directly in the EnumFlagHintAnnotation, just like you do for some others like RangeHint.

chippmann: I can but i don't want to as then the user would have to write the names of the enums everytime by hand. Now it's done automatically. On top of that I would need to add additional checks to ensure the number of names match the number of enums in the case of enum, lists. Also i would need to check if no name is entered twice. Additionally i would need to implement this error reporting in the idea plugin as well.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Type.kt
chippmann: Nope it's not. The actual underlying type is passed as a super type. So in the case of nesting type aliases, i receive them the same way as a normal class hierarchy.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/PropertyHintStringGeneratorProvider.kt
chippmann: done. feel free to unresolve if what i've done is not what you expected

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/RegisteredProperty.kt
chippmann: Ah yeah probably. An additional reason might be because ksp provides a method to get the nearest overridden member. This function is called `findOverridee`. It's probably to match that as well.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/model/Type.kt
chippmann: Removed

chippmann:
> Suggestion:
> Wouldn't that be better to have the model part of the entry generation as its own module and then use it as a dependency for the generator?
> People wanting to add another language will only need to import that model module when implementing their own processor instead of the whole entry generator.
> Given that model is kinda the bridge between the source language and the actual entry file. It seems natural to me that it's its own module that both sides import as a dependency.
> 
> Example: With the new Godot extension system in 4.0, we might need to handle 2 types of entry files, one for scripts, one for extensions.
> We will be able to use the same Kotlin source processor for both, but not the same generator will be used.

If you are ok with it, i do the separation in a separate PR when i extract the checks anyways

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
CedNaru: I don't get it. Can't you just get this information directly from the KSP without the user writing anything?

Currently, in the registration, you just take the Enum and iterate all the values to create the string, can't you simply move that part directly in the symbol processing?

chippmann: Yeah i could, but either I'd have to implement such a resolve callback like i mentioned earlier, or i cheat by enriching the `EnumHintAnnotation` model of the entry gen with values which are not actually present on the "real" annotation which i don't like to do beacause then the model does not represent the actual source code anymore.

So basically two options to do such a thing automatically during entry generation:
- Adding a resolve callback to the interface of the entry gen so calling tools like `godot-kotlin-symbol-processor` can implement it to resolve a `Type` to a `Clazz`.
- Or cheat by adding the needed information to one of the existing models to which it does not belong

Because the information about what enum values an enum has, is information on class level, so it should belong into the `Clazz` model.
But when checking the type of a property it is not of type `Clazz` it's of type `Type`. In compiler classes (or ksp for that matter) you then have a `resolve` method to get the class declaration. This is not present in my abstraction because it's not needed and has to be implemented by the calling tool (`godot-kotlin-symbol-processor` in the case of kotlin). So this would be the first option.

CedNaru: The second option is the one I was thinking about. It's easy to implement and I don't see how it doesn't follow the Model. Imo, it's perfectly in line. We want what Godot considers to be an int with an EnumFlag hint. So there is nothing wrong with storing the different enum values inside the EnumFlagAnnotation.
Maybe the difference is in how we see the model. The actual implementation we use in Kotlin is a Set of Enum which is then converted back and forth to an Int but it only matters for Kotlin, another JVM language might not use the same kind of trick in the first place.  
The model should reflect the Godot side which is the reference, not the language source which can be interchanged by another one. We are the one cheating in the first place already by using a Set<Enum>.
I guess this is what you call cheating because the Kotlin source is a Set<Enum> property with an annotation without parameters. But the Godot side is an Int with an annotation holding multiples values (I say annotations in the context of godot both for the current export syntax in GDScript and the true annotations we will get in GDSCript 2.0 in Godot 4.0)

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/ConstructorOverloadingCheck.kt
```diff
@@ -0,0 +1,20 @@
...
+    override fun execute() {
+        sourceFiles
+            .flatMap { it.registeredClasses }
+            .forEach { registeredClass ->
+                registeredClass
+                    .constructors
+                    .groupBy { it.parameters.size }
+                    .filter { it.value.size > 1 }
+                    .forEach { (argCount, overloadingConstructors) ->
+                        logger.error("Constructor overloading is not yet supported! Overloading constructors in class ${registeredClass.fqName}: ${overloadingConstructors.joinToString { it.fqName }} with arg count: $argCount")
```
piiertho, line 16: Maybe specify that overloading is not supported only on registered constructor ?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/StringExtensions.kt
```diff
@@ -0,0 +1,7 @@
+package godot.entrygenerator.ext
+
+fun String.camelToSnakeCase(): String {
```
piiertho, line 3: note for future, should be merged with string operations we have in api generator.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
```diff
@@ -0,0 +1,176 @@
+package godot.entrygenerator.ext
```
piiertho, line 1: Same here, should be merged in a near future with what we have in api generator.

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/typehint/coretypes/JvmCoreTypeTypeHintGenerator.kt
```diff
@@ -0,0 +1,23 @@
...
+import godot.entrygenerator.model.ColorNoAlphaHintAnnotation
+import godot.entrygenerator.model.RegisteredProperty
+
+class JvmCoreTypeTypeHintGenerator(
+    private val registeredProperty: RegisteredProperty
+) : PropertyTypeHintGenerator(registeredProperty) {
+    override fun getPropertyTypeHint(): ClassName {
+        return when (propertyHintAnnotation) {
+            ColorNoAlphaHintAnnotation -> ClassName("godot.core.PropertyHint", "COLOR_NO_ALPHA")
+            //TODO: implement ImageCompressLossy
```
piiertho, line 15: TODOs are normal, or implems are forgotten ?

kt/godot-library/src/main/kotlin/godot/gen/godot/ARVRController.kt
```diff
@@ -37,21 +38,21 @@ import kotlin.Suppress
  * The position of the controller node is automatically updated by the [godot.ARVRServer]. This makes this node ideal to add child nodes to visualize the controller.
  */
 @GodotBaseType
-open class ARVRController : Spatial() {
+public open class ARVRController : Spatial() {
```
piiertho, line 41: Is it a change from Kotlin Poet new version ?

kt/utils/godot-build-props/build.gradle.kts
```diff
@@ -43,3 +43,5 @@ publishing {
 }
 
 project.extra["artifacts"] = arrayOf("buildProps")
+
+tasks.withType<Copy> { duplicatesStrategy = DuplicatesStrategy.INCLUDE}
```
piiertho, line 47: What is this for ?

piiertho:
> > It seems some conversions methods we use are now deprecated:
> > ```
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 42): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 56): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (747, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (749, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (753, 46): 'toChar(): Char' is deprecated. Direct conversion to Char is deprecated. Use toInt().toChar() or Char constructor instead.
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (185, 50): Unchecked cast: List<P> to L
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (781, 9): Parameter 'isTool' is never used
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\util\Utils.kt: (17, 7): 'toLowerCase(): String' is deprecated. Use lowercase() instead.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > > Task :kt:godot-library:compileKotlin
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\extensions\PackedSceneExt.kt: (7, 46): Parameter 'editState' is never used
> > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\global\GDCore.kt: (214, 24): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > ```
> 
> I suggest that I'll update usages related to the entry gen in this PR and all other occurrences should be done in a separate PR.
> If you are ok with it I would create the corresponding issue.

I'm not ok on this one. I know this PR is already huge. But here we speak about merging in master with deprecated code :/

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
chippmann: True we already kinda cheat with the enum set. But the model does not cheat yet. The model is meant to be a abstracted representation of the actual source code. Which is the case atm.
If we add that additional information to the models `EnumHintAnnotation` or `EnumFlagHintAnnotation` it does not represent the actual source code anymore, because there that information is not present in the annotation.

That's what i mean by cheating. I just ask myself if it's worth it to let the model be different from the source code just for this one case.

So yeah i guess the difference is how we see the model. I saw it as a abstracted, language independent representation of the source code. And according to your comment, you rather see it as an conversion from the language to the godot side of things.

So i guess we first need to find a common ground what the model should represent. I'm still more keen to see it as an abstraction of the actual source code. And IMO it's the job of the entry gen to use that abstracted information to convert/generate it to something godot can understand and use.
So from my point of view, if we really want to change the enum registration, we should do it with a `resolve` function and not with a model change.

But of course I'm open to change my view on how we see the model. But atm i just don't see why it should represent the godot side and not the source code side of things

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/ConstructorOverloadingCheck.kt
chippmann: Agree

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/typehint/coretypes/JvmCoreTypeTypeHintGenerator.kt
chippmann: They are "normal". If you look at https://github.com/utopia-rise/godot-kotlin-jvm/blob/master/kt/godot-library/src/main/kotlin/godot/annotation/propertyHintAnnotations.kt All those annotations are commented. We previously decided that we do not yet implement hint annotations from which we do not know the purpose or how they function.
So i kept them as we had them before. With comments and todos.
But i can of course also completely remove them for now if you prefer

kt/godot-library/src/main/kotlin/godot/gen/godot/ARVRController.kt
chippmann: yeah. I did not touch anything about the api gen

kt/utils/godot-build-props/build.gradle.kts
chippmann: Damn. Should have commented why i added that xD
Builds failed because of an issue with gradle 7.x
Something about property files from different modules IIRC.
I'll see if A it is still needed, and B if i find the issue again where i found this solution and comment it accordingly.

chippmann:
> > > It seems some conversions methods we use are now deprecated:
> > > ```
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 42): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (452, 56): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (454, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 31): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (458, 45): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (747, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (749, 29): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\core\Color.kt: (753, 46): 'toChar(): Char' is deprecated. Direct conversion to Char is deprecated. Use toInt().toChar() or Char constructor instead.
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (185, 50): Unchecked cast: List<P> to L
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\runtime\Registration.kt: (781, 9): Parameter 'isTool' is never used
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-runtime\src\main\kotlin\godot\util\Utils.kt: (17, 7): 'toLowerCase(): String' is deprecated. Use lowercase() instead.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```
> > > > Task :kt:godot-library:compileKotlin
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\extensions\PackedSceneExt.kt: (7, 46): Parameter 'editState' is never used
> > > w: C:\Users\ptmeisels\git\godot\modules\kotlin_jvm\kt\godot-library\src\main\kotlin\godot\global\GDCore.kt: (214, 24): 'toInt(): Int' is deprecated. Conversion of Char to Number is deprecated. Use Char.code property instead.
> > > ```
> > 
> > 
> > I suggest that I'll update usages related to the entry gen in this PR and all other occurrences should be done in a separate PR.
> > If you are ok with it I would create the corresponding issue.
> 
> I'm not ok on this one. I know this PR is already huge. But here we speak about merging in master with deprecated code :/

Hmm yeah. Ok I'll change it in this PR

kt/utils/godot-build-props/build.gradle.kts
chippmann: Ah the problem was with how gradles copy task functions. If you copy multiple files at once you now need to explicitly set a duplication strategy: https://docs.gradle.org/7.0.2/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy
And this is the case for the copy task in build props it seems. So i specified the duplication strategy explicitly here

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/hintstring/EnumFlagHintStringGenerator.kt
chippmann: To sum up and document the call @CedNaru, @piiertho and i just had:
We'll look at the model as an abstracted version of the things Godot needs, not like an AST.
One of the main reasons: stuff like `enumFlags` might be of different type in different languages and we cannot rely on that in the entry gen. Also it keeps the model more in line with changes to the GDScript exports (which will be annotations for Godot 4.0 as well).
So I'll change the enum hint string generation to store the enum values as metadata inside the anonotation in the model (option 2 mentioned above).
So we can get rid of the different registration of enum types.

The only thing that still will be registered differently, is the `enumFlag` type.
This one will be reworked once we have the entry gen type configuration mentioned below:

With a later PR, I'll make the entry gen configurable so it knows about built in types.
Currently fqNames of kotlin builtin types are hardcoded inside the entry gen. Also some other types are hardcoded.
These should be configurable when using the entry gen, so it knows what type is an Integer language independent

chippmann: done

chippmann:
@piiertho and @CedNaru please also retest it on Windows and MacOS after my change with the enumAnnotationMetadata

piiertho (approved):

piiertho (approved):

CedNaru (approved):

PR #249, opened 2021-07-26
Fix android build

chippmann:
This fixes the android build by using the newer `d8` tool rather than the deprecated `dx` tool.
This also has the advantage of smaller resulting jars, as the unnecessary `class` files are not bundled into it.
Just the `classes.dex` file (and the service file in the case of `main-dex.jar`).

One config field was added to the `godot` config block: `androidCompileSdkDir`. As the `android.jar` contained in that dir, is needed by the `d8` tool. The path to that dir depends on the target compilation sdk version. In most cases this is the most recent sdk version. Corresponding documentation was added.

Please test this especially on windows, as I'm unsure if the gradle task `createMainDexJar` is properly configured for windows.


docs/src/doc/user-guide/exporting.md
```diff
@@ -46,16 +46,31 @@ In order to build for Android, set the `isAndroidExportEnabled` flag to `true` i
 On android we do not embed a JVM, we use the existing ART provided by the OS. In order for your game to load the necessary jar files, they need to be converted into dex format. Our gradle plugin will handle this for you, but you need to fulfill the following requirements:
 
 - Android SDK installed
-- `dx` tool resolvable by doing one of the following:
-    - set the `PATH` environment variable to include `<android-sdk-dir>/build-tools/<version>/`
-    - set `dxToolPath` to the file path of `dx`:
+- `d8` tool resolvable by doing one of the following:
+    - (discuraged) only set the `ANDROID_SDK_ROOT` environment variable
```
CedNaru, line 50: typo ?
Also, we only have 2 methods and the one not discouraged is basically hard writing the path in the Gradle build file,  not really portable in a team using different systems.

docs/src/doc/user-guide/exporting.md
```diff
@@ -46,16 +46,31 @@ In order to build for Android, set the `isAndroidExportEnabled` flag to `true` i
...
     godot {
-        isAndroidExportEnabled.set(true)
-        dxToolPath.set("${System.getenv("ANDROID_SDK_ROOT")}/build-tools/30.0.3/dx")
+        d8ToolPath = File("/home/user/Android/Sdk/build-tools/31.0.0/d8")
+    }
+    ```
+    !!! info
+        The d8 tool is located at <android-sdk-root>/build-tools/<build-tool-version>/d8
+- `android sdk dir` set by doing one of the following:
+    - (discuraged) only set the `ANDROID_SDK_ROOT` environment variable
```
CedNaru, line 61: same here

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
```diff
@@ -37,125 +40,194 @@ fun Project.setupConfigurationsAndCompilations(godotExtension: GodotExtension, j
...
-                            } else {
-                                commandLine(godotExtension.dxToolPath.get(), "--version")
+            val checkD8ToolAccessible = with(create("checkD8ToolAccessible")) {
+                group = "godot-kotlin-jvm"
+                description = "Checks if the d8 tool is accessible and executable. Needed for android builds only"
+
+                doLast {
+                    try {
+                        val d8Tool = godotExtension.d8ToolPath
+                            ?: throw IllegalArgumentException("d8 tool not set! Make sure you've either set the ANDROID_SDK_ROOT environment variable or set the d8ToolPath. For more information, visit: https://godot-kotl.in/en/stable/user-guide/exporting/#android")
```
CedNaru, line 87: This one is written 3 times in that files, maybe just create your own Exception class with that text in it.

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
```diff
@@ -37,125 +40,194 @@ fun Project.setupConfigurationsAndCompilations(godotExtension: GodotExtension, j
...
+                archiveVersion.set("")
+                archiveClassifier.set("")
 
-            dependencies {
-                it.exclude(it.dependency("org.jetbrains.kotlin:kotlin-stdlib.*"))
-                it.exclude(it.dependency("com.utopia-rise:godot-library:.*"))
-            }
+                dependencies {
+                    it.exclude(it.dependency("org.jetbrains.kotlin:kotlin-stdlib.*"))
+                    it.exclude(it.dependency("com.utopia-rise:godot-library:.*"))
```
CedNaru, line 74: Do we really need to remove our library here given it's compileOnly in the first place ?

docs/src/doc/user-guide/exporting.md
chippmann: Well it is portable. You just have to setup your gradle script properly.
You can say in your team that for android to be built, one has to have the `ANDROID_SDK_ROOT` environment variable set and then it would look like this:
```kotlin
d8ToolPath.set("${System.getenv("ANDROID_SDK_ROOT")}/build-tools/30.0.3/d8")
```
Which is very portable IMO.

And what typo are you referring to?

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
chippmann: :+1:

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/setupConfigurationsAndCompilations.kt
chippmann: Yes because theoretically one can add our library as well in the `dependencies` block manually to whatever he wants. For example one can set it as `api` so it's a transitive dependency. Which is completely valid. But even then we don't want it to be included in the final jar.

docs/src/doc/user-guide/exporting.md
CedNaru: discuraged

chippmann: Fixed the typo.
Are you ok with my explanation why i think it's portable or do you disagree?

CedNaru: I understand that the "only SDK path" solution is not the most reliable one because the location can change depending on the SDK version. 
But then just write the example you just gave me that combines the SDK path and relative path instead of the "/home/user". The current version gives the feeling you shouldn't use the SDK path at all.

chippmann: Agree. Will change the wording

CedNaru (approved):

piiertho (approved):

PR #251, opened 2021-07-26
Merge godot-library with godot-runtime

### LINKED ISSUE #214: Merging godot-runtime and godot-library
We initially split the Kotlin codebase into 2 modules  `godot-runtime` and `godot-library` with the basic idea that core objects and glue code should be in the former and everything related to the Godot API in the latter.

But because of Kotlin not supporting package visibility like Java. We very often need to access classes located in  `godot-runtime`  from  `godot-library` and sometimes there are classes that we would like to put in  `godot-runtime`  but need Object and Reference so we have to move them to  `godot-library`

The direct consequence is that many  `godot-runtime` classes are made public for the sake of  `godot-library` but this also made them available for the users.
Right now the users can access classes like GarbageCollector or GodotStatic.
We should just put everything in the same module and replace the module name with a package. That way we will finally be able to set to internal everything we need.

chippmann:
This resolves #214


CedNaru (approved):
Nice, finally.
I feel like we will need at a later date, to reorganize the whole module because the current structure is kinda messy. But for now, it's good enough.

piiertho:
Can we rebase this branch on master to ensure all green ?

raniejade (approved):

piiertho (approved):

piiertho:
Just thought right now, should we upgrade from 0.2.0 to 0.3.0 as this modify accessibles methods ?

chippmann:
> Just thought right now, should we upgrade from 0.2.0 to 0.3.0 as this modify accessibles methods ?

Hmm yeah it's a api change so it would justify it. @CedNaru and @raniejade whats your take on this?

raniejade:
Bumping version SGTM.

PR #252, opened 2021-07-27
Kotlin version checks in gradle plugin

### LINKED ISSUE #208: Gradle plugin doesn't apply kotlin plugin by default
At the moment we let the user apply the kotlin plugin, which means they can configure which version to use - unfortunately that wouldn't work as we ship the kotlin stdlibs together with the bootstrap jar. To prevent further issues we need to:

(1) check if the kotlin plugin is applied, if it is we fail the build since we want to a specific version.
(2) check the version of the kotlin plugin (there should be a way to check for it), this can happen if the user specifies the kotlin plugin dependency directly in buildSrc or in the buildscript block. If it doesn't match the version we expect, fail the build.
(3) apply the kotlin plugin.

To fix #201 we can either include reflect by default or just don't ignore the kotlin reflect dependency when the user specifies it. The latter means the user can just do:

```kt
dependencies {
  implementation(kotlin("reflect")) // kotlin helper here is locked down to the version we want if do the steps mentioned above.
}
```

chippmann:
This resolves #208.

The kotlin jvm plugin is now applied by our gradle plugin by default. If a kotlin version is already applied (either by the user explicitly applying the kotlin jvm plugin by himself or through something like `buildSrc`), then we check the applied kotlin version and let the build fail if the detected kotlin version is not the one we're expecting/supporting.

As there are cases were an advanced user is familiar with kotlin compiler plugin constraints and still wants to use another kotlin version, one can suppress the version check by specifying `godot.jvm.suppressKotlinIncompatibility=true` in the `gradle.properties` file.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -62,6 +62,7 @@ nav:
     - Advanced:
       - Custom src dirs: user-guide/advanced/custom-src-dirs.md
       - Custom gradle wrapper path: user-guide/advanced/custom_gradle_wrapper_path.md
+      - Gradle plugin configuration: user-guide/advanced/gradle-plugin-configuration.md
       - Commandline args: user-guide/advanced/commandline-args.md
       - Kotlin singletons: user-guide/advanced/kotlin-singleton.md
   - Contributing:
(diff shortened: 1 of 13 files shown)


CedNaru (approved):
Feels weird to not have a kotlin version in a gradle build file but I guess we have no choice for now.

piiertho (approved):

PR #255, opened 2021-07-27
Fix jvm type registration

### LINKED ISSUE #243: `JVM_` prefixed `VariantType` types not visible in inspector.
Types like `Int` are not visible in inspector because godot doesn't know the ordinal of `JVM_INT`.
Long on the other hand works as the ordinal of `VariantType.LONG` is correct.

chippmann:
Fixes #243.
When implementing the entry generator with ksp i forgot to register the JVM_* types as their actual godot type rather than their KtVariant type. This lead to the inspector not showing properties with the types `Int`, `Float` and `Byte`.
This PR fixes this.


CedNaru:
Noting: This fix is only temporary as we will soon rework the VariantType enum so the JVM types disapear.

chippmann:
> Noting: This fix is only temporary as we will soon rework the VariantType enum so the JVM types disapear.

Indeed. But this change will most likely be in another release. So we still need this fix until then

piiertho (approved):

CedNaru (approved):

PR #260, opened 2021-08-08
bugfix: windows github actions

piiertho:
Scons from version 4.2.0 is not anymore compatible with godot 3 build scripts.  
It was fixed in master branch of godot. So for now let fix scons version to `4.1.0`.


chippmann (approved):

raniejade (changes requested):
You can create a `requirements.txt` containing:

```
scons==4.1.0
```

Then you can use that file by: `python -m pip install -r requirements.txt`. This means we only need to manage the version in a single place.

raniejade (approved):

PR #261, opened 2021-08-19
Show warning when JAVA_HOME is not defined

### LINKED ISSUE #133: Editor crashes when opening a project if JAVA_HOME is not defined

chippmann:
Resolves #133 

This prevents our module from loading if JAVA_HOME is not defined in editor mode instead of letting the editor crash.
It also adds a warning on editor startup about the issue:


src/editor/godot_kotlin_jvm_editor.cpp
```diff
@@ -143,11 +143,20 @@ void GodotKotlinJvmEditor::_notificationv(int p_notification, bool p_reversed) {
...
         menu_pop_up->add_item("About Godot Kotlin JVM");
         editor_base_control->add_child(about_dialog);
+        editor_base_control->add_child(error_dialog);
 
         FileSystemDock* file_system_dock = get_editor_interface()->get_file_system_dock();
         file_system_dock->connect("files_moved", this, "on_file_system_dock_file_moved");
         file_system_dock->connect("file_removed", this, "on_file_system_dock_file_removed");
         file_system_dock->connect("folder_moved", this, "on_file_system_dock_folder_moved");
+
+        if (OS::get_singleton()->get_environment("JAVA_HOME").empty()) {
```
raniejade, line 153: We are technically making the same check here and at `GDKotlin` but displaying the error message differently. What do you think about doing the initialization checks in `GDKotlin` exclusively? Then we can just expose a `GDKotlin::is_initialized()` method that returns true if the module is properly initialized. For error messages, we can just expose `GDKotlin::get_init_errors()` or something then display whatever string it returns.

src/editor/dialog/error_dialog.cpp
```diff
@@ -0,0 +1,35 @@
+
+#ifdef TOOLS_ENABLED
+
+#include <editor/editor_settings.h>
+#include <scene/gui/texture_rect.h>
+#include <scene/gui/rich_text_label.h>
+#include <editor/editor_scale.h>
+#include "error_dialog.h"
+
+ErrorDialog::ErrorDialog(): error_message_label(memnew(RichTextLabel)) {
```
piiertho, line 10: I think it misses call to super class constructor.

piiertho (changes requested):

src/editor/godot_kotlin_jvm_editor.cpp
chippmann: Agreed. Will change that

src/editor/dialog/error_dialog.cpp
chippmann: Hmm i blindly copied it from the About dialog which in turn i copied from the C# one. Didn't check that. I'll look into if it's necessary or not

src/gd_kotlin.h
```diff
@@ -57,6 +59,10 @@ class GDKotlin {
     KtClass* find_class(const StringName& p_script_path);
 
     const GdKotlinConfiguration& get_configuration();
+
+    bool is_initialized;
+
+    Vector<Pair<String, String>> configuration_errors;
```
piiertho, line 65: I would prefer a getter and make those variables private so that it cannot be changed from outside of `GDKotlin`.

piiertho (changes requested):

CedNaru (approved):

piiertho (approved):

src/gd_kotlin.h
```diff
@@ -57,6 +63,10 @@ class GDKotlin {
     KtClass* find_class(const StringName& p_script_path);
 
     const GdKotlinConfiguration& get_configuration();
+
+    bool initialized() const;
+
+    Vector<Pair<String, String>> get_configuration_errors();
```
piiertho, line 69: I would expect a const reference here.

piiertho (changes requested):

piiertho (approved):
Great Job ! Thx !

PR #263, opened 2021-08-21
Refactor gradle plugin

### LINKED ISSUE #259: Refactor `setupConfigurationsAndCompilations.kt`
We have more and more tasks added to `setupConfigurationsAndCompilations.kt` in Gradle Plugin.  
With GraalVM, this file is now becoming big, and will become more if we add new platforms.  
We should split this one in several parts, so that it remains readable.

chippmann:
Resolves #259 ~~by introducing a task registry with which the individual tasks are properly grouped and accessible and configurable in a typesafe way.~~ by splitting the task registration/creation into individual extension functions residing in their own files.

It also splits some functions up into individual parts and removes some parameter passing by introducing extension functions/properties for gradles `Project` class so we can access extensions directly for example.

~~This could further be improved though by using sealed classes or interfaces for the tasks and use them inside the `TaskRegistry` as in it's current state, the `TaskRegistry` could become unreadable if even more tasks would be added in the future. But to keep this more simple until we actually need it, i opted to just keep them in an enum for now.~~

Edit: updated implementation after: https://github.com/utopia-rise/godot-kotlin-jvm/pull/263#pullrequestreview-740434136


piiertho:
I see you did not touch bindings logic but we have a failure on `test_array_any_not_null_append` test. Do you have it locally ? Or maybe we have a deeper problem ?

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
```diff
@@ -0,0 +1,85 @@
...
+                deleteBuildLockTask = deleteBuildLockTask
+            )
+            val packageMainJarTask = packageMainJarTask(
+                createBuildLockTask = createBuildLockTask,
+                deleteBuildLockTask = deleteBuildLockTask,
+                generateEntryServiceFileTask = generateEntryServiceFileTask
+            )
+
+            // START: android specific tasks
+            val checkD8ToolAccessibleTask = checkD8ToolAccessibleTask()
```
raniejade, line 37: I was expecting an `if (isAndroidExportConfigured())` statement here, what happens if the user doesnt' want to build for android?

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
```diff
@@ -0,0 +1,85 @@
...
+                packageBootstrapDexJarTask = packageBootstrapDexJarTask,
+                packageMainJarTask = packageMainJarTask
+            )
+            val packageMainDexJarTask = packageMainDexJarTask(
+                createMainDexFileTask = createMainDexFileTask
+            )
+            // END: android specific tasks
+
+            // START: graal native image specific tasks
+            val checkNativeImageToolAccessibleTask = checkNativeImageToolAccessibleTask()
```
raniejade, line 56: same as above

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/android/checkAndroidJarAccessibleTask.kt
```diff
@@ -0,0 +1,28 @@
+package godot.gradle.tasks.android
+
+import godot.gradle.projectExt.godotJvmExtension
+import org.gradle.api.Project
+import org.gradle.api.Task
+
+fun Project.checkAndroidJarAccessibleTask(): Task {
+    return with(tasks.create("checkAndroidJarAccessible")) {
```
raniejade, line 8: Not a big issue, but have you tried using `register` instead of `create`? With register a task is only configured/resolved when it will participate in the execution.

raniejade (commented):
This looks good, but I have a couple of questions.

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
chippmann: This is currently taking care of in the setup of the `build` task:
```kotlin
if (godotJvmExtension.isAndroidExportEnabled.get()) {
    finalizedBy(packageBootstrapDexJarTask, packageMainDexJarTask)
}
```

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
chippmann: Same

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/android/checkAndroidJarAccessibleTask.kt
chippmann: Ah nice to know. I'll change them

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
raniejade: Ahh, weird way of doing it - we don't really need to configure the android/native image tasks when they are disabled?

raniejade: Anyway, not a big issue. This will change once I'm done with the refactor.

raniejade (approved):

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
chippmann: :+1:

piiertho (approved):

PR #256, opened 2021-07-28
Feature/graal support

piiertho:
This adds GraalVM native image support.
This adds to Gradle plugin option to generate GraalVM native image.
This adds command line argument `--java-vm-type`.
This adds godot kotlin specific project configuration file `godot_kotlin_configuration.json`.

Currently platforms supported are Windown, Linux and MacOS, on AMD64 hosts.

*Warning*: Reloading is not possible with native-image, as it would require to reload JVM.

PR #270, opened 2021-09-08
CPP reloading

### LINKED ISSUE #257: [PROPOSAL] Write reload logic in cpp
Currently we have the bootstrap class which handles file system changes and reloads classes accordingly.
It basically is the sole reason why we have to ship the `godot-bootstrap.jar` alongside the engine executable.

I propose that we implement this filewatching and reloading logic completely in cpp. Looking at the impl of some other cpp filewatching libs (like https://github.com/apetrone/simplefilewatcher/) we should be able to to this kind of filewatching ourselves pretty easily as well even if we have to implement it platform specific. Maybe even godot has some platform independent way of watching filesystem changes.

Also now that we use a buildLock file instead of watching the `build` dir directly, the file watching logic is greatly simplified.
So we would in fact only need to watch one file which further simplifies things a lot.

This would pave the way for better/simpler editor builds and would completely remove the need for two jar files as we could pack everything into one jar. Which would also simplify export targets like android a lot.

### LINKED ISSUE #258: Refactor `GDKotlin`
Follow up issue from #256. Currently `gd_kotlin.cpp` is a mess as it slowly grew in functionality and complexity.
We have to refactor it at some point.

chippmann:
~~Draft PR only because the documentation and github workflows for the editor release are not yet updated. I wait with that until no big changes from any review is pending anymore. The implementation itself can be reviewed already.~~

Resolves #257 but #258 needs to be still done separately even though many cleanup changes are already made in this PR.

This implements the reload logic in cpp rather than rely on java file watching and a executor service in `Bootstrap`.

This has several advantages:
- No need to split `bootstrap` and `main` anymore -> both are now in a single `jar`/`so` called `usercode` (similarly to the graal native image implementation)
- The point above also means that we don't have to ship any other files with our engine builds anymore
- This also means that if graal native image is used for the editor, it now can also be reloaded, which previously was not possible
- The user can define the reload check interval in the editor settings (as the reload file watching is implemented in the editor plugin)
- No reloading code in the final executable (ifdev's in the editor plugin strip that away)
- Gradle task setup is easier now as there are fewer tasks
- No service file anymore

This is implemented roughly in the following way:

**Reloading:**
In the editor plugin there is the `BuildLockWatcher`. It polls for changes in the `usercode.jar` (or `usercode.so` for graal native image) in regular intervals (configurable in editor settings and 0.5s by default) and if there is a change detected, it unloads any existing code and reloads the classes again.
To keep the file "stable" for any sudden file changes (moves/deletions/edits in a gradle clean for example) the `usercode.jar` is copied to a tmp dir before it gets loaded. It is never loaded from the project directory directly.
Sadly there is no file watching logic existing in godot which utilize platform specific filesystem events. To implement something like that we'd either would have to implement it ourselves or rely on third party libraries which both are not ideal. So i settled for a simple timer node approach to poll for changes (like the C# module does).

**Entry file loading:**
This is not done through a service file anymore because this would require to have one classloader existing globally which keeps the base classes loaded (namely the base `godot.registration.Entry` file (note; this is not the generated user entry class, this is it's parent class!)). If this is not the case, when a reload occurs, the classloader does not know that the previously loaded `godot.registration.Entry` class and the newly loaded `godot.registration.Entry` class are the same type of class, and throws a `ServiceConfigurationError` stating that `godot.Entry` is not a subtype of `godot.registration.Entry`. See: https://stackoverflow.com/a/65404325. This problem was the root cause why we splitted and shipped the `bootstrap` jar separately in the first place.
To prevent this issue we could either still keep our `Bootstrap` class in a separate `jar` and ship that with the engine or do it like I've done now; pack everything into one jar and load the `Entry` file through reflection.
As this is one call at the beginning and uses standard java reflection (not kotlin reflection) it should be fine IMO and doesn't even need a reflection config for graal native image as it can automatically detect these reflection calls (see: https://www.graalvm.org/reference-manual/native-image/Reflection/#automatic-detection)

**Module configuration:**
@piiertho introduced `GdKotlinConfiguration` with #256 . But until now only the requested `jni::Jvm::Type` was setup in there and all other configurations were only done through commandline.
I changed that so every configuration setting that we provide, can be defined in this configuration json. This has the advantage that one can define every setting we provide per platform and flavor (debug, production, and so on).
The way this loading works now is:
For the editor: the configuration json is loaded if existent, any supplied cmd args override what's defined in the json
During export: only what's written in the json is copied to the export
During runtime:  same as for the editor but with the exported json of course.

This was tested with the following platforms and flavours:
- x11 debug editor
- x11 release_debug editor
- x11 release_debug
- x11 release
- android debug_release
- android release
- x11 release_debug graal native image
- x11 release graal native image

This still needs testing for osx and windows both in editor and with export templates.


src/gd_kotlin_configuration.cpp
```diff
@@ -97,11 +121,235 @@ void GdKotlinConfiguration::set_max_string_size(int p_max_string_size) {
...
+void GdKotlinConfiguration::override_json_config_with_cmd_args(GdKotlinConfiguration* r_configuration) {
+    const List<String>& cmdline_args{OS::get_singleton()->get_cmdline_args()};
+    const String& arg_prefix{"--"};
+
+    for (int i = 0; i < cmdline_args.size(); ++i) {
+        const String cmd_arg{cmdline_args[i]};
+        if (cmd_arg.find(arg_prefix + vm_type_identifier) >= 0) {
+            String vm_type_as_string;
+            if (split_jvm_debug_argument(cmd_arg, vm_type_as_string) == OK) {
+#ifdef __ANDROID__
```
CedNaru, line 291: Just for practical reasons for when we will also support IOS.
I would add a `sanitize_config` function where we put all the OS-related preprocessors so they are all in one place and call it inside `load_gd_kotlin_configuration_from_json_and_args`

src/path_provider.cpp
```diff
@@ -0,0 +1,153 @@
...
+#elif defined _WIN32 || defined _WIN64
+    String usercode_jar_path{build_lock_dir_path + "\\" + get_usercode_name()};
+#endif
+    return usercode_jar_path;
+#else //TOOLS_ENABLED
+    return ProjectSettings::get_singleton()->globalize_path("user://" + get_usercode_name());
+#endif //TOOLS_ENABLED
+}
+
+bool PathProvider::copy_usercode_jar_if_necessary() {
```
CedNaru, line 21: really weird to have this function in PathProvider.  I would put it back in GDKotlin for now.

raniejade (commented):
I haven't read the PR yet but:

> To prevent this issue we could either still keep our Bootstrap class in a separate jar and ship that with the engine or do it like I've done now; pack everything into one jar and load the Entry file through reflection.
As this is one call at the beginning and uses standard java reflection (not kotlin reflection) it should be fine IMO and doesn't even need a reflection config for graal native image as it can automatically detect these reflection calls 

If we pack everything in a single jar and not needing to ship the bootstrap jar then that means users can know freely use what kotlin version they want no?

piiertho:
As far as I understand this new reloading is done on main thread.  
This can lead to problems on low end hardware, like freeze.  
I think this should be made on another thread.

src/gd_kotlin_configuration.cpp
chippmann: I suggest doing this when doing #258 and we refactor the `GDKotlin` class because i guess we'll rework this part with that as well.

chippmann:
> I haven't read the PR yet but:
> 
> > To prevent this issue we could either still keep our Bootstrap class in a separate jar and ship that with the engine or do it like I've done now; pack everything into one jar and load the Entry file through reflection.
> > As this is one call at the beginning and uses standard java reflection (not kotlin reflection) it should be fine IMO and doesn't even need a reflection config for graal native image as it can automatically detect these reflection calls
> 
> If we pack everything in a single jar and not needing to ship the bootstrap jar then that means users can know freely use what kotlin version they want no?

As discussed on discord, this won't solve the kotlin version issue as we're still bound to the version `KSP` supports

CedNaru (approved):

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/graal/createGraalNativeImage.kt
```diff
@@ -72,7 +69,7 @@ fun Project.createGraalNativeImageTask(
 
                         godotJvmExtension.nativeImageToolPath.get(),
                         "-cp",
-                        "\"${godotBootstrapJar.absolutePath}\";\"${mainJar.absolutePath}\"",
+                        "\"${usercodeJar.absolutePath}\"",
```
piiertho, line 72: I think there's no more need of string template.

src/editor/build/build_lock_watcher.cpp
```diff
@@ -0,0 +1,66 @@
...
+        if (poll_delay_variant.get_type() == Variant::REAL || poll_delay_variant.get_type() == Variant::INT) {
+            poll_delay = poll_delay_variant;
+        }
+        OS::get_singleton()->delay_usec(static_cast<int>(poll_delay * 1000 * 1000));
+    }
+    jni::Jvm::detach();
+}
+
+void BuildLockWatcher::reload_if_needed() {
+    String build_usercode_path = PathProvider::provide_build_usercode_path();
```
piiertho, line 26: I would use const ref here for const correctness and give indication to future developper.

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/graal/createGraalNativeImage.kt
chippmann: For windows it is. There the `"` around the path are needed it seems

src/gd_kotlin_configuration.cpp
```diff
@@ -22,60 +23,83 @@ String GdKotlinConfiguration::to_json() {
...
+    result[jvm_jmx_port_identifier] = jvm_jmx_port;
+    result[is_waiting_for_debugger_identifier] = is_waiting_for_debugger;
+    result[max_string_size_identifier] = max_string_size;
+    result[jvm_force_gc_identifier] = is_gc_force_mode;
+    result[jvm_disable_gc_identifier] = !is_gc_activated;
+    result[jvm_disable_closing_leaks_identifier] = should_display_leaked_jvm_instances_on_close;
 
-    return JSON::print(result);
+    String json = JSON::print(result);
+    LOG_VERBOSE(vformat("Output json: %s", json))
```
piiertho, line 36: I'm pretty sure this was here for debugging purpose, but do we still need it ?

piiertho (commented):
I did not finished to review, but here are few comments.
Will finish to review tomorrow.
Quite a bunch of work you did :D

src/path_provider.cpp
```diff
@@ -0,0 +1,98 @@
...
+    switch (vm_type) {
+        case jni::Jvm::JVM:
+            return jar_extension;
+        case jni::Jvm::ART:
+            return jar_extension_android;
+        case jni::Jvm::GRAAL_NATIVE_IMAGE:
+#ifdef __linux__
+            return dynamic_library_extension_linux;
+#elif __APPLE__
+#include <TargetConditionals.h>
```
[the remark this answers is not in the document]
chippmann: WDYM with stacking?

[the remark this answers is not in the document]
piiertho: I mean set the import on top of cpp file

piiertho (changes requested):
For now I got troubles on windows regarding permissions
image
Here is temp folder content:
image

chippmann:
Closing. We need to postpone this until 4.0 is done and take a new shot at this.
The currently implemented way does not work as expected. We need to create a small PoC outside of this project and reevaluate how we want to achieve this.

PR #286, opened 2021-11-12
feature: register user types variant mappings

piiertho:
This adds registration of user types variant mappings.  
This enables to declare `VariantArray` and `Dictionary` parameterized by a user type. Like:
```kotlin
variantArrayOf<MyUserType>()
```


CedNaru (approved):

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/EntryGenerator.kt
```diff
@@ -42,6 +42,7 @@ object EntryGenerator {
                     )
                 }
             }
+            registerUserTypesVariantMappings(sourceFiles.map { it.registeredClasses }.flatten())
```
chippmann, line 45: Why not just use `flatMap`?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/filebuilder/MainEntryFileBuilder.kt
```diff
@@ -48,4 +60,21 @@ object MainEntryFileBuilder {
...
+        for (registeredClass in registeredClasses) {
+            registerUserTypesVariantMappingsFunSpec.addStatement(
+                "ret.add(%T::class)",
+                ClassName(registeredClass.containingPackage, registeredClass.name)
+            )
+        }
+        registerUserTypesVariantMappingsFunSpec.addStatement(
+            "return ret"
+        )
+    }
```
chippmann, line 79: This could all be a single return statement of an immutable list if you pass in the classes in the constructor of the list.
So the generated code would look like:
```kotlin
return listOf(
    ClassOne::class,
    ClassTwo::class
)
```

So probably the generation code would look something like this (probably not completely correct as I'm just typing that in the browser):
```kotlin
val codeBlock = CodeBlock
    .builder()
    .addStatement("return %M(", MemberName("kotlin.collections", "listOf"))

registeredClasses.forEach { registeredClass ->
    codeBlock.addStatement("%T::class,", ClassName(registeredClass.containingPackage, registeredClass.name))
}
registerUserTypesVariantMappingsFunSpec.addCode(codeBlock.build())
```

harness/tests/test/unit/test_array.gd
```diff
@@ -51,6 +51,12 @@ func test_array_typed_not_null_remove():
 	assert_eq(invocation_script.nav_meshes_size(), 0, "Array size should be 0")
 	invocation_script.free()
 
+func test_create_user_type_variant_array():
+	var invocation_script = godot_tests_Invocation.new()
+	var user_type_variant_array = invocation_script.create_variant_array_of_user_type()
+	assert_not_null(user_type_variant_array, "Should return an empty user typed variant array.")
```
chippmann, line 57: Shouldn't we also check the actual type here? So we're sure that whatever object array we receive here has the actual types we expect?

chippmann (changes requested):
Great work!
Finally this will be possible :-)

harness/tests/test/unit/test_array.gd
piiertho: You can't check type of another language in gdscript

chippmann: Hmm but IIRC you can do this: `if obj extends Class`
So theoretically we should be able to do `assert_true(user_type_variant_array[0] extends godot_tests_subpackage_OtherScript, "Bla bla")` no?

piiertho: Nop, we cannot as right operand is a `KotlinScript`

chippmann: Hmm i always though also scripts written in other languages land in the class db because of the registering.
Well if this is not the case, I'm fine with it

chippmann (approved):

PR #294, opened 2021-11-25
Enhancement/add edit state to instance as extension

piiertho:
This adds editState parameter to `PackedScene.instanceAs` extension.  
Default value is 0 like in api.


chippmann (changes requested):
The version in the docs also needs to be changed.

CedNaru (approved):

chippmann (approved):

PR #295, opened 2021-11-27
feature: enable to set android min api

piiertho:
Add androidMinApi parameter to gradle plugin to enable specify --min-api argument of d8 tool

### DIFF
--- a/docs/src/doc/user-guide/exporting.md
+++ b/docs/src/doc/user-guide/exporting.md
@@ -63,6 +63,13 @@ On android, we do not embed a JVM, we use the existing ART provided by the OS. I
     }
     ```
 
+- Setting the `androidMinApi` (equivalent to `--min-api` argument of `d8` tool), default is `21`:
+    ```kt
+    godot {
+        androidMinApi = 22
+    }
+    ```
+
 !!! warning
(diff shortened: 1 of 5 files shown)


CedNaru:
Maybe we can also add the multidex in this PR as well as it's a closely related matter.

piiertho:
> Maybe we can also add the multidex in this PR as well as it's a closely related matter.

Multidex is a legacy feature for Dalvik.
On art it is enabled by default. (So after Android 21)
https://developer.android.com/studio/build/multidex

CedNaru (approved):

chippmann:
> > Maybe we can also add the multidex in this PR as well as it's a closely related matter.
> 
> Multidex is a legacy feature for Dalvik. On art it is enabled by default. (So after Android 21) https://developer.android.com/studio/build/multidex

Still we need to handle that. AFAIK the d8 tool will generate multiple dex files. Currently we then only add the main dex file to the exported game together with the bootstrap dex file.
But not the other dex files.
But i suggest doing this in a separate PR.

chippmann (dismissed):

piiertho:
> > > Maybe we can also add the multidex in this PR as well as it's a closely related matter.
> > 
> > 
> > Multidex is a legacy feature for Dalvik. On art it is enabled by default. (So after Android 21) https://developer.android.com/studio/build/multidex
> 
> Still we need to handle that. AFAIK the d8 tool will generate multiple dex files. Currently we then only add the main dex file to the exported game together with the bootstrap dex file. But not the other dex files. But i suggest doing this in a separate PR.

I dont think so.
We generate dex files that will be zipped into one jar.  
We should not confuse dex files and jar zipping them.

chippmann (approved):

CedNaru (approved):

PR #296, opened 2021-11-29
Enhancement/set core type interface public

piiertho:
`CoreType` interface has been set `internal` since `0.3.0-3.4.0`.  
This is something that can be helpful to have public.  
As an example @CedNaru and me have a project where we created some extensions that should be called only on `CoreType` inheritor.


chippmann (changes requested):
If this is to be public, then we should at least add a kdoc stating that this interface should not be inherited from by a user's class

chippmann (approved):

CedNaru (approved):

PR #297, opened 2021-12-01
Update intellij plugin

chippmann:
This updates the intellij plugin to work with the new `2021.3.x` versions if IDEA.

It also improves initial indexing performance for the scene file cache by introducing a custom file type for scene files.
Upon this change, we could later add improvements like implementing a psi based parser and lexer for proper integration in the IDE and thus maybe replace https://github.com/utopia-rise/jvm-godot-resource-serialization. This would allow for better refactoring support and better IDE hints regarding scene integrations of scripts like script path and signal connections. Maybe even a hint in code stating with what value a field is overriden from within the inspector.

It also fixes a bug which was noticed by @piiertho where the registered name hint was not correct when the `@RegisterClass` annotation did define a custom class name.

@piiertho Please take a careful look at the version bump so i didn't miss it somewhere.


kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/language/scene/language/GodotSceneLanguage.kt
```diff
@@ -0,0 +1,14 @@
+package godot.intellij.plugin.language.scene.language
+
+import com.intellij.lang.Language
+
+class GodotSceneLanguage private constructor() : Language("godotScene") {
+    override fun getDisplayName(): String = NAME
+    override fun isCaseSensitive(): Boolean = true
+
+    companion object {
```
CedNaru, line 9: Is that for Java compatibility issues ?

chippmann: AFAIK yes.
`GodotSceneFileType` for example is referenced by such a field in the `plugin.xml`:
```xml
<fileType name="GodotSceneFile"
            fieldName="INSTANCE"
            language="godotScene"
            extensions="tscn"
            implementationClass="godot.intellij.plugin.language.scene.filetype.GodotSceneFileType"/>
```

For the `GodotSceneLanguage` i assumed it must be the same as all implementations i checked (all tough all written in plain java) have such an `INSTANCE` field.

kt/plugins/godot-intellij-plugin/src/IJ183/kotlin/godot/intellij/plugin/extension/moduleExt.kt
```diff
@@ -0,0 +1,6 @@
+package godot.intellij.plugin.extension
+
+import com.intellij.openapi.module.Module
+import org.jetbrains.kotlin.idea.framework.isGradleModule
+
+fun Module.isGradleModule(): Boolean = this.isGradleModule()
```
piiertho, line 6: Why is this necessary ?

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/listener/GodotSceneBulkFileListener.kt
```diff
@@ -35,24 +41,29 @@ class GodotSceneBulkFileListener(private val project: Project) : BulkFileListene
...
 
     @Suppress("UnstableApiUsage")
-    private fun getContainingFiles() = FileBasedIndex
-        .getInstance()
-        .getContainingFiles(
-            FileTypeIndex.NAME,
-            PlainTextFileType.INSTANCE,
-            GlobalSearchScope.projectScope(project)
-        )
+    private fun getContainingFiles(callback: (List<VirtualFile>) -> Unit) {
```
piiertho, line 51: I would rename this method as it now execute a callback

piiertho (approved):
Few comments, otherwise ok for merging !
Thx a lot !

kt/plugins/godot-intellij-plugin/src/IJ183/kotlin/godot/intellij/plugin/extension/moduleExt.kt
chippmann: Because jetbrains moved the function `isGradleModule` to another package.
In older sdk versions, it's now at the old path and in newer ones it's at the new path, breaking compatibility.
This file is a wrapper which links to the correct path.
A different version is included now in our plugin based on the target version of IDEA we're building for.

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/listener/GodotSceneBulkFileListener.kt
chippmann: :+1:

piiertho (approved):

CedNaru (approved):

PR #298, opened 2021-12-26
Feature/update to godot 3.4.2

piiertho:
This updates module to 3.4.2 version of godot.


chippmann (changes requested):
One thing we forgot to update each time and needs correcting: here the version is still wrong: https://godot-kotl.in/en/stable/contribution/setup/

Just as a note: remember to move the stable tag so the website get's updated. :-)

piiertho:
> One thing we forgot to update each time and needs correcting: here the version is still wrong: https://godot-kotl.in/en/stable/contribution/setup/
> 
> 
> 
> Just as a note: remember to move the stable tag so the website get's updated. :-)

IIRC CI updates stable tag when releasing

chippmann:
> > One thing we forgot to update each time and needs correcting: here the version is still wrong: https://godot-kotl.in/en/stable/contribution/setup/
> > Just as a note: remember to move the stable tag so the website get's updated. :-)
> 
> IIRC CI updates stable tag when releasing

Yeah it should but IIRC it does not work reliably. Didn't we have some problems with it in the past? When i look at our current documentation website, it doesn't seem up to date to me? For example see the version here: https://godot-kotl.in/en/stable/user-guide/versioning/

piiertho:
> > > One thing we forgot to update each time and needs correcting: here the version is still wrong: https://godot-kotl.in/en/stable/contribution/setup/
> 
> > > Just as a note: remember to move the stable tag so the website get's updated. :-)
> 
> > 
> 
> > IIRC CI updates stable tag when releasing
> 
> 
> 
> Yeah it should but IIRC it does not work reliably. Didn't we have some problems with it in the past? When i look at our current documentation website, it doesn't seem up to date to me? For example see the version here: https://godot-kotl.in/en/stable/user-guide/versioning/

True, did not catch this. But stable tag is up to date :/

chippmann:
> > > > One thing we forgot to update each time and needs correcting: here the version is still wrong: https://godot-kotl.in/en/stable/contribution/setup/
> > 
> > 
> > > > Just as a note: remember to move the stable tag so the website get's updated. :-)
> > 
> > 
> > > 
> > 
> > 
> > > IIRC CI updates stable tag when releasing
> > 
> > 
> > Yeah it should but IIRC it does not work reliably. Didn't we have some problems with it in the past? When i look at our current documentation website, it doesn't seem up to date to me? For example see the version here: https://godot-kotl.in/en/stable/user-guide/versioning/
> 
> True, did not catch this. But stable tag is up to date :/

Probably we need to set the stable tag before the actual version tag?

piiertho:
That's what we do already IIRC

chippmann (dismissed):

CedNaru (approved):

CedNaru (approved):

chippmann (approved):

CedNaru (approved):

PR #299, opened 2022-01-09
bugfix/ci/change to new action zip

piiertho:
It seems `papeloto` changed his name, and someone clever (you can report him) decided to break the world by creating a trolling repo with former name: https://github.com/papeloto/action-zip/issues/1

This changes action to new repo.

### DIFF
--- a/.github/workflows/deploy-godot-editor-release.yaml
+++ b/.github/workflows/deploy-godot-editor-release.yaml
@@ -110,7 +110,7 @@ jobs:
           arguments: godot-library:build
 
       - name: Zip release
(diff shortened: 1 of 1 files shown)


chippmann (approved):
Ok for now but in the long run we should just use the zip libraries per platform to remove such thirdparty libs IMO.

piiertho:
> Ok for now but in the long run we should just use the zip libraries per platform to remove such thirdparty libs IMO.

I think we should setup some acceptance rules for third party dependencies.

CedNaru (approved):

PR #344, opened 2022-07-22
bugfix: embedd JRE in export app

### LINKED ISSUE #266: Add documentation about JRE generation per platform
We need to clarify in our documentation that the jre generation using jlink has do be done on a per platform basis.
Atm this is unclear and leads to export problems like on this nice GameJam entry: https://marquiskurt.itch.io/planting-uneasy-feelings when the developer cannot test the export on each platform (especially relevant for Jam's where the time limit might not allow this).

We could and should also provide a github action which exports the game and generates corresponding jre's per platform.
This github action could then be distributed to the github actions marketplace for easy usage.

piiertho:
This makes difference between JRE for arm and amd systems.
This fixes OSX exports by making use of Plugins app folder.
Only applicable to Godot `3.5+` versions.

This resolves #266


piiertho:
We need to check all exports (all os + hotspot / graal native image) before merging this

piiertho:
Tested on windows for both hotspot and graal native image.

CedNaru:
LGTM
But reminder that at some point we should rework everything related to systems and arch as it's becoming quite messy.

CedNaru (approved):

piiertho:
> LGTM But reminder that at some point we should rework everything related to systems and arch as it's becoming quite messy.

Maybe you can write an issue so that we don't forget ?

piiertho:
Tested on Android.

piiertho:
Tested all exports (arm64, x86_64 and universal) for MacOS.

src/jni/jvm_loader.cpp
```diff
@@ -101,7 +102,22 @@ String jni::JvmLoader::get_embedded_jre_path() {
...
+#if defined(OSX_ENABLED) && !defined(TOOLS_ENABLED)
+                        "../PlugIns/"
+#else
+                        ""
+#endif
+                        ,
+                        jni::JniConstants::CURRENT_RUNTIME_JRE
+                )
+        };
+        jre_path = OS::get_singleton()->get_executable_path()
```
chippmann, line 117: IMO we should use the `res://` folder here for everything that is not an export (like we had before)
Because when i run the game directly with the editor executable, it does not pick up the jre folder as the directory of the "executable" is where the editor executable lies.
Example:
2022-08-12-151944_1332x174_scrot

chippmann (changes requested):
Tested both debug and release exports on linux x64.

But the documentation needs to be updated to mention the changed requirements for generating the jre folders: `jre` -> `jre-amd64` or `jre-arm64`
This also is a good time to add the missing documentation from: #266

Also we need to mention this somewhere (probably on discord) for existing users as this breaks their exports if they do not create new jre folders.

src/jni/jvm_loader.cpp
piiertho: Fixed this by modifying a bit `jvm_loader.cpp`, can you re-test on linux ?

CedNaru (dismissed):

docs/src/doc/contribution/setup.md
```diff
@@ -13,10 +13,14 @@ Make sure that Java is installed and its PATH set in your system as well (at lea
 4. Build sample
     - navigate to `<module-root>/harness/tests`
     - create embedded JVM: `jlink --add-modules java.base,java.logging --output jre`
```
chippmann, line 15: The command in this line can be removed

docs/src/doc/user-guide/debugging.md
```diff
@@ -4,7 +4,8 @@ In order to debug your code using an embedded JRE, make sure to include the foll
 - `jdk.jdwp.agent`
 
 !!! info
-    Example: `jlink --add-modules java.base,java.logging,jdk.jdwp.agent --output jre`
+    Example for ADM64 systems: `jlink --add-modules java.base,java.logging,jdk.jdwp.agent --output jre-amd64`  
+    Example for ARM64 systems: `jlink --add-modules java.base,java.logging,jdk.jdwp.agent --output jre-arm64`
```
chippmann, line 8: Typo: `ADM64` -> `AMD64`

chippmann (changes requested):
Tested on linux;
- Editor binary without jre: OK
- Editor binary with jre: OK
- Release_debug export: OK
- Release export: Ok

Found another issue: #354. But this does not need to be addressed in this PR IMO

Only issues i found are small things in the docs

piiertho:
> Tested on linux;
> 
> - Editor binary without jre: OK
> 
> - Editor binary with jre: OK
> 
> - Release_debug export: OK
> 
> - Release export: Ok
> 
> 
> 
> Found another issue: #354. But this does not need to be addressed in this PR IMO
> 
> 
> 
> Only issues i found are small things in the docs

I agree this should be done in another PR.

chippmann (approved):

Issue #308, opened 2022-03-13
Simplify GC leveraging `objectID` and `validatorCount`


chippmann:
Godot 4 introduced a more sophisticated `objectID` containing many informations about an object. Such as a bit flag for it it's a reference, or a pointer to a `slot` object containing usefule information for us such as the actual pointer to the object and a `validatorCount` for the object index.

With this new information and the changes in https://github.com/utopia-rise/godot-kotlin-jvm/issues/306, we should be able to greatly simplify the GC code.

Issue #306, opened 2022-03-13
Change CPP/JVM object handling to incorporate `objectID` changes


chippmann:
Godot 4 introduced a more meaningful/powerful `objectID` which now contains a bit flag for `Reference` objects and a pointer to a `slot` object containing object information data and the actual pointer of the object.

We need to change our object handling across language borders to utilize this new `objectID`. This should greatly simplify how we handle objects and object pointers across borders.

This will also enable https://github.com/utopia-rise/godot-kotlin-jvm/issues/308 to greatly simplify our GC.

PR #358, opened 2022-09-27
New memory management using instance bindings

CedNaru:
This new model makes use of Godot's instance binding, a feature available for every Godot Object and reworked for 4.0 as well as the ObjectDB.

Instance Bindings allow altering the lifecycle of any Godot object, being creation, destruction, or increment/decrement of the RefCounted counter.
For native Godot objects simply sent to the JVM, this binding will keep a reference to the Kotlin instance, preventing it to be collected. 
Before we had to keep an instance in our custom Garbage collector ourselves to keep them alive, but now that job is done directly by the binding on the C++ side. The moment the native Godot object dies,  so will the JVM instance.

Regarding Refcounted Script, the solution to the cyclic references was inspired by the C# module. The concept is simple: when the counter of a Refcounter reaches 1, we know that it means the JVM is the only one still holding a reference. When it happens, the JNI reference in the binding is transformed into a weak one. The moment the JVM no longer keeps a reference to it, the RefCounted can be disposed of.

With the new ObjectDB comes a new way to handle objects, long story short, it's no longer necessary to treat RefCounted and Object differently so the code was simplified in many places.

Remains to do:

- [x] Make the JVM responsible for deleting the native RefCounted if it is the last one to decrement the counter.
- [x] There is a delay between the creation of the binding and the assignment of its JNI reference. Give the job of assigning the reference to the custom GC.


kt/godot-library/src/main/kotlin/godot/core/KtObject.kt
```diff
@@ -55,17 +53,10 @@ abstract class KtObject {
...
-                GarbageCollector.registerReference(this)
-            } else {
-                GarbageCollector.registerObject(this)
-            }
+            GarbageCollector.registerObject(this)
         }
     }
 
-    @Suppress("FunctionName")
-    open fun ____DO_NOT_TOUCH_THIS_isRef____() = false
```
piiertho, line 67: <3

kt/godot-library/src/main/kotlin/godot/core/StringName.kt
```diff
@@ -1,5 +1,6 @@
 package godot.core
 
+import godot.core.memory.GarbageCollector
```
piiertho, line 3: Still needed ?

src/java_instance_wrapper.h
```diff
@@ -1,77 +1,113 @@
...
+	bool is_weak;
+	jni::JObject wrapped;
+	jni::JObject class_loader;
+
+	static jni::JClass j_class;
 
-    static jni::JClass j_class;
+	JavaInstanceWrapper(const char *p_class_name, jni::JObject p_wrapped, jni::JObject &p_class_loader);
 
-    JavaInstanceWrapper(const char* p_class_name, jni::JObject p_wrapped, jni::JObject& p_class_loader);
```
piiertho, line 28: Should we align on godot formatting ?

src/jni/types.h
```diff
@@ -51,8 +51,13 @@ namespace jni {
         template <class T>
         inline T new_global_ref(Env& env);
         void delete_global_ref(Env& p_env);
+		template <class T>
```
piiertho, line 54: formatting is weird

src/kotlin_instance.cpp
```diff
@@ -1,81 +1,87 @@
...
 
-Object* KotlinInstance::get_owner() {
-    return owner;
+void KotlinInstance::get_property_list(List<PropertyInfo> *p_properties) const {
+	kt_class->get_property_list(p_properties);
 }
 
-void KotlinInstance::set_owner(Object *object) {
-    owner = object;
+Variant::Type KotlinInstance::get_property_type(const StringName &p_name, bool *r_is_valid) const {
```
piiertho, line 58: maybe add a todo to not forget to implement it

src/kotlin_instance.cpp
```diff
@@ -1,81 +1,87 @@
...
-    if (function) {
-        function->invoke(this->wrapped_object, p_args, p_argcount, ret_var);
-    } else {
-        r_error.error = Callable::CallError::CALL_ERROR_INVALID_METHOD;
-    }
-    return ret_var;
+	KtFunction *function{ kt_class->get_method(p_method) };
+	Variant ret_var;
+	if (function) {
+		function->invoke(binding->kt_object, p_args, p_argcount, ret_var);
```
piiertho, line 80: tab instead of spaces for indenting

src/kt_variant.h
```diff
@@ -196,37 +194,34 @@ namespace ktvariant {
...
+		KotlinBinding* binding = KotlinBindingManager::get_instance_binding(ptr);
+
+		//We only need to retrieve the constructor if the binding is not ready yet.
+		int classID = 0;
+		if(!binding->is_setup){
+			StringName class_name{ptr->get_class_name()};
+			if (!TypeManager::get_instance().JAVA_ENGINE_TYPES_CONSTRUCTORS.has(class_name)) {
+				class_name = ClassDB::get_parent_class(class_name);
+
+				while (class_name == String()) {
```
piiertho, line 214: isn't comparition with `StringName` more efficient ?

src/kt_variant.h
```diff
@@ -196,37 +194,34 @@ namespace ktvariant {
...
-            id = RefDB::get_instance().get_ref_id(ref);
-
-        } else {
-            id = ptr->get_instance_id();
-        }
+		//Create a binding if it doesn't exist yet.
+		KotlinBinding* binding = KotlinBindingManager::get_instance_binding(ptr);
+
+		//We only need to retrieve the constructor if the binding is not ready yet.
+		int classID = 0;
```
piiertho, line 208: I think `constructorId` would be more clear

src/memory/kotlin_binding_manager.cpp
```diff
@@ -0,0 +1,112 @@
...
+	binding->is_setup = true;
+}
+
+void KotlinBindingManager::delete_script_binding(KotlinInstance *instance) {
+	KotlinBindingManager &manager = get_instance();
+	//We avoid concurrent modification of the map. Before that it should be safe as the destructor of an object is not supposed to be called multiple times.
+	manager.spin.lock();
+	manager.binding_map.erase(instance->get_owner());
+	manager.spin.unlock();
+}
```
piiertho, line 102: doucle scope closing ?

src/memory/kotlin_binding_manager.cpp
```diff
@@ -0,0 +1,112 @@
...
+	&_instance_binding_free_callback,
+	&_instance_binding_reference_callback
+};
+
+void *KotlinBindingManager::_instance_binding_create_callback(void *p_token, void *p_instance) {
+	KotlinBindingManager &manager = get_instance();
+	Object *owner = reinterpret_cast<Object *>(p_instance);
+
+	manager.spin.lock();
+	KotlinBinding *binding = &manager.binding_map.insert(owner, KotlinBinding())->value;
```
piiertho, line 15: Why getting it as a ptr instead of a ref ?

kt/godot-library/src/main/kotlin/godot/core/StringName.kt
CedNaru: All native core times have to register themselves to the Garbage Collector.
I moved the garbage collector to a new package so the import is necessary now.

src/jni/types.h
CedNaru: Clion code style config being weird here

src/kt_variant.h
CedNaru: Probably yes

src/kt_variant.h
CedNaru: Will change it

src/memory/kotlin_binding_manager.cpp
CedNaru: The function signature requires to return a pointer.

src/java_instance_wrapper.h
CedNaru: I suggest we have a separate PR to set one format style once for good.

src/kotlin_instance.cpp
CedNaru: for a future PR

chippmann:
@CedNaru This PR will "close" #308 no?

piiertho:
> @CedNaru This PR will "close" #308 and #306 no?

I think we should test everything is ok before closing those issues.

chippmann:
> > @CedNaru This PR will "close" #308 and #306 no?
> 
> I think we should test everything is ok before closing those issues.

For sure but it should fix them once this is not a draft anymore and merged i guess

src/memory/bridges/packed_string_array_bridge.cpp
```diff
@@ -10,69 +10,69 @@ JNI_INIT_STATICS_FOR_CLASS(PackedStringArrayBridge)
...
-            (void*) PackedStringArrayBridge::engine_call_reverse
+        const_cast<char*>("engine_call_reverse"),
+        const_cast<char*>("(J)V"),
+        (void*) PackedStringArrayBridge::engine_call_reverse
     };
 //    jni::JNativeMethod engine_call_join_method{
 //            "engine_call_join",
 //            "(J)V",
 //            (void*) PackedStringArrayBridge::engine_call_join
 //    };
```
chippmann, line 51: dead code

src/memory/bridges_manager.cpp
```diff
@@ -1,89 +1,89 @@
...
+    initialize_bridge(env, class_loader, bridges::PACKED_INT_64_ARRAY_BRIDGE_CLASS_NAME, packed_int_64_array_bridge);
+    initialize_bridge(env, class_loader, bridges::PACKED_FLOAT_32_ARRAY_BRIDGE_CLASS_NAME, packed_float_32_array_bridge);
+    initialize_bridge(env, class_loader, bridges::PACKED_FLOAT_64_ARRAY_BRIDGE_CLASS_NAME, packed_float_64_array_bridge);
+    initialize_bridge(env, class_loader, bridges::PACKED_STRING_ARRAY_BRIDGE_CLASS_NAME, packed_string_array_bridge);
+    initialize_bridge(env, class_loader, bridges::PACKED_VECTOR2_ARRAY_BRIDGE_CLASS_NAME, packed_vector2_array_bridge);
+    initialize_bridge(env, class_loader, bridges::PACKED_VECTOR3_ARRAY_BRIDGE_CLASS_NAME, packed_vector3_array_bridge);
+}
+
+BridgesManager::BridgesManager() : memory_bridge(nullptr), gd_print_bridge(nullptr), variant_array_bridge(nullptr), dictionary_bridge(nullptr),
+                                   rid_bridge(nullptr), node_path_bridge(nullptr), string_name_bridge(nullptr),
```
chippmann, line 32: Not that it makes a difference but why are they swapped now?

src/memory/transfer_context.cpp
```diff
@@ -1,8 +1,8 @@
-#include <assert.h>
-#include <core/io/marshalls.h>
 #include "transfer_context.h"
-#include "gd_kotlin.h"
-#include "kotlin_instance.h"
+#include "core/io/marshalls.h"
+#include "modules/kotlin_jvm/src/gd_kotlin.h"
+#include "modules/kotlin_jvm/src/kotlin_instance.h"
+#include <assert.h>
```
chippmann, line 5: I don't see any actual usages of `assert` is it a leftover?

chippmann (changes requested):
There seems to still be an issue when trying to run the tests:
`Error: Exception in thread "main" java.lang.NoSuchMethodError: Method 'boolean godot.core.memory.GarbageCollector$MemoryBridge.destroyRef(long)' name or signature does not match`

Also we really need to define a formatting standard and do another PR with all formatting changes (as you suggested in this PR a few times). ATM our formatting is all over the place.

src/memory/bridges_manager.cpp
CedNaru: The compiler is raising warnings because it wants initialisation to be in the same order as declaration.

src/memory/transfer_context.cpp
CedNaru: probably a leftower yeah

CedNaru:
Yeah, next PR is planned to add a clang format file. I already got one prepared.

src/gd_kotlin.h
```diff
@@ -1,14 +1,13 @@
...
 #include "bootstrap.h"
-#include "kt_class.h"
 #include <core/string/ustring.h>
-#include "transfer_context.h"
-#include "modules/kotlin_jvm/src/bridges/memory_bridge.h"
 #include "kotlin_script.h"
 #include "gd_kotlin_configuration.h"
+#include "jni/wrapper.h"
+#include "kt_class.h"
+#include <modules/kotlin_jvm/src/memory/transfer_context.h>
```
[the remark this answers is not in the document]
CedNaru: Clion adding them automatically

kt/godot-library/src/main/kotlin/godot/core/memory/GarbageCollector.kt
```diff
@@ -174,64 +146,50 @@ internal object GarbageCollector {
...
+        bindingList.clear()
+        counter = 0
+
+        //We poll the reference that have been clear by the GC and then call c++ code to destroy the native object.
+        synchronized(ObjectDB) {
+            while (counter < CHECK_NUMBER) {
+                val ref = ((refReferenceQueue.poll() ?: break) as GodotWeakReference)
+                val index = ref.id.index
+                val otherRef = ObjectDB[index]
+                //Check if the ref in the DB hasn't been replaced by a new object before the GC could remove it.
```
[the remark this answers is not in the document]
CedNaru: Nope it should be fine.
Objects are manually released anyway, so it's not up to our module to do it.
When it comes to RefCounted, The ReferenceQueue automatically store the newly collected kotlin instances so we can call decrement on the C++ side and eventually delete.

src/kotlin_instance.cpp
piiertho: I don't think this should be for a future PR, all files are formatted with spaces.

src/memory/bridges/bridges_utils.h
```diff
@@ -1,14 +1,14 @@
...
-    static inline T* from_uint_to_ptr(jlong raw_ptr) {
-        return reinterpret_cast<T*>(static_cast<uintptr_t>(raw_ptr));
-    }
-}
-
-#endif //GODOT_JVM_BRIDGES_UTILS_H
+#ifndef GODOT_JVM_BRIDGES_UTILS_H
+#define GODOT_JVM_BRIDGES_UTILS_H
+
+#include "modules/kotlin_jvm/src/jni/wrapper.h"
```
piiertho, line 4: I would not change it to filepath include as it starts from `modules`

src/java_instance_wrapper.h
```diff
@@ -1,77 +1,113 @@
 #ifndef GODOT_JVM_JAVAINSTANCEWRAPPER_H
 #define GODOT_JVM_JAVAINSTANCEWRAPPER_H
 
-
 #include "jni/wrapper.h"
 
 #define JNI_INIT_STATICS_FOR_CLASS(C) \
-    C::JNIMethods C::jni_methods{};
+	C::JNIMethods C::jni_methods{};
```
piiertho, line 7: tab here too

kt/godot-library/src/main/kotlin/godot/core/TransferContext.kt
```diff
@@ -12,8 +12,8 @@ internal object TransferContext {
     private val bufferSize: Int
```
piiertho, line 12: Shouldn't we change package to reflect cpp change ?

src/memory/kotlin_binding.cpp
```diff
@@ -32,6 +32,18 @@ bool KotlinBinding::refcount_decremented_unsafe() {
     return refcount == 0;
 }
 
+void KotlinBinding::set_kt_object(KtObject* p_kt_object) {
+    //This function should only be called when we know the object is a RefCounter. We directly reinterpret the point to it
```
piiertho, line 36: typo `RefCounter` -> `RefCounted`

piiertho (approved):
Awesome work !

PR #362, opened 2022-10-09
feat: Use godot's actions to build from CI

piiertho:
This does not modify deploy editor action. Editor app is buggy on OSX, it fails to generate a configuration on project manager. I'll fix this bug before modifying editor app, so that I can test the generated app.  
This modifies check pr and deploy export template actions. It makes use of godot's actions to simplify workflow.  
This make android build in parallel, and add creation of OSX export template app.  

Modified deploy actions has been tested [here](https://github.com/piiertho/godot-kotlin-jvm/releases).


CedNaru (approved):

.github/actions/create-android-export-template/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+      with:
+        name: android-export-${{ inputs.target }}-binary-armv7
+    - name: Download android ${{ inputs.target }} export arm64v8 binary
+      uses: actions/download-artifact@v3
+      with:
+        name: android-export-${{ inputs.target }}-binary-arm64v8
+    - name: Build android ${{ inputs.target }} export template
+      shell: bash
+      run: |
+        cd platform/android/java/ && ./gradlew generateGodotTemplates
```
chippmann, line 28: Could we space the steps with a new line for readability (for all the actions) or is this considered best practice?

chippmann (approved):

.github/actions/create-android-export-template/action.yaml
piiertho: We can, I'll do this in next PR for editor app if it's ok for you ?

chippmann: Ok for me

PR #364, opened 2022-10-25
Add stack management for icall arguments.

CedNaru:
When several calls from Kotlin to Godot are made inside each other. (Kotlin calls Godot that calls Kotlin that call Godot etc…). The same address were reused to store the method arguments, which lead to memory corruption when signals were connected to several methods (The same arguments have to be used several times without being erased).

This PR adds a stack to handle such cases.
The main drawback is that we are limited in size and can end up in a Stack Overflow, which can be hard to debug for the users. In return, it's quite efficient.

The alternative would be to totally drop the shared array for storing arguments and make a new one dynamically for each call, which consume time, but it's safer.
It's required to use 2 different arrays, one to store the arguments values after being read from the buffer, one to store pointers to those values as required by Godot.


piiertho:
Can we imagine a hybrid system which goes with static array as you did right there, and if we are going to stack overflow pass on a dynamic array ?  
This way we would have efficience for most cases, and still be able to support deeper stack with extra computational cost ?

CedNaru:
That would require adding an extra branch in the function. I guess we could just set it as "unlikely" but I need to benchmark the actual cost.

CedNaru:
After benchmarking with and without a branch (the test is constantly making icalls from Kotlin to C++), there is no noticeable  difference in performance, so I'm adding it.

piiertho:
Thanks for this great job.
Maybe we can also add a test with one signal connected to two method to ensure the bug we spotted is totally behind us.

piiertho (approved):

piiertho (approved):

PR #366, opened 2022-10-30
Generate `.gdignore` files for the `build` and the `gradle` directories

### LINKED ISSUE #220: Godot editor filesystem search shows generated Kotlin classes
Is there a way to filter them out?

image

chippmann:
Resolves #220.

Still a draft PR, because i have issues with our class registration inside the godot editor in general atm and cannot properly test this Impl.
Will be resolved after our meeting from today i hope.

### DIFF
--- a/kt/plugins/godot-gradle-plugin/build.gradle.kts
+++ b/kt/plugins/godot-gradle-plugin/build.gradle.kts
@@ -42,6 +42,9 @@ dependencies {
     implementation("gradle.plugin.com.github.johnrengelman:shadow:${DependenciesVersions.shadowJarPluginVersion}")
     implementation("com.google.devtools.ksp:com.google.devtools.ksp.gradle.plugin:${DependenciesVersions.kspVersion}")
 
+    // used for idea sync gradle task integration (ex. generateGdIgnoreFilesTask)
+    implementation("org.jetbrains.gradle.plugin.idea-ext:org.jetbrains.gradle.plugin.idea-ext.gradle.plugin:1.1.6")
+
     implementation(project(":godot-build-props"))
 }
 
--- a/kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/configureThirdPartyPlugins.kt
+++ b/kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/configureThirdPartyPlugins.kt
@@ -4,6 +4,7 @@ import com.github.jengelman.gradle.plugins.shadow.ShadowPlugin
 import com.google.devtools.ksp.gradle.KspGradleSubplugin
 import org.gradle.api.Project
 import org.gradle.plugins.ide.idea.IdeaPlugin
+import org.jetbrains.gradle.ext.IdeaExtPlugin
 import java.io.File
 
 fun Project.configureThirdPartyPlugins() {
@@ -11,6 +12,7 @@ fun Project.configureThirdPartyPlugins() {
     repositories.google() //add google repository for ksp (kotlin symbol processing
     pluginManager.apply(KspGradleSubplugin::class.java)
     pluginManager.apply(IdeaPlugin::class.java) //needed so idea can find and index the generated sources from ksp
+    pluginManager.apply(IdeaExtPlugin::class.java) //needed so idea can find and index the generated sources from ksp
     pluginManager.apply(ShadowPlugin::class.java)
 
     addKspGeneratedSourcesToMainSourceSet()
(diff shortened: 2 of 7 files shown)


Frontrider:
After rethinking it: what if I want to generate classes to be used from godot? Is it possible to put those generated classes in say `build/generated/godot` instead of `build/generated/ksp`?

chippmann:
> 

You mean generating your own classes or using our generated classes?

Frontrider:
> You mean generating your own classes or using our generated classes?

Generating my own. Not sure how this behaves with kapt.

chippmann:
> > You mean generating your own classes or using our generated classes?
> 
> Generating my own. Not sure how this behaves with kapt.

You would need to define a custom output directory not in the `build` directory. I don't have a strong opinion on whether this is a problem or not. Generally i like the idea of hiding the whole build folder from godot and define a custom output directory for the code generation when i want to generate source files i intend to use with godot.
What's your take on that?

Frontrider:
> > > You mean generating your own classes or using our generated classes?
> > 
> > 
> > Generating my own. Not sure how this behaves with kapt.
> 
> You would need to define a custom output directory not in the `build` directory. I don't have a strong opinion on whether this is a problem or not. Generally i like the idea of hiding the whole build folder from godot and define a custom output directory for the code generation when i want to generate source files i intend to use with godot. What's your take on that?

My take is probably a little conflicted, but generated classes belong to the build folder (they are build artifacts) and not commented into the repository (as they will be regenerated).

Although,on the other side I'm noticing is that you may not even be able to register generated classes properly, because of some processing order magic. That leaves the whole build folder's exclusion reasonable.

chippmann:
> > > > You mean generating your own classes or using our generated classes?
> > > 
> > > 
> > > Generating my own. Not sure how this behaves with kapt.
> > 
> > 
> > You would need to define a custom output directory not in the `build` directory. I don't have a strong opinion on whether this is a problem or not. Generally i like the idea of hiding the whole build folder from godot and define a custom output directory for the code generation when i want to generate source files i intend to use with godot. What's your take on that?
> 
> My take is probably a little conflicted, but generated classes belong to the build folder (they are build artifacts) and not commented into the repository (as they will be regenerated).
> 
> Although,on the other side I'm noticing is that you may not even be able to register generated classes properly, because of some processing order magic. That leaves the whole build folder's exclusion reasonable.

Well I only partly agree. There are cases where generated file should not be treated as build artifacts IMO. One example are the generated godot api classes we commit in this project. Altough they are generaated, they should not be treated as build artefacts. Same goes for classes generated for godot IMO. Well it depends what you do with the generated class. If you add it to a node, it should be commited IMO to track it's api changes over time and check that no change you made, resulted in a api change which affects the function of the node you applied said script. If you however only call said script from let's say GDScript, then its fine IMO to treat it as a build artefact.

And you are correct. As we do source code processing and generation on our own, order matters. But thats a general code generation "problem" not specific to this project. Depending on what you generate, our module should come last in the code generation chain most of the time.

All of that said; i guess we could provide a more fine granular .gdignore config for the build folder but TBH i would only do that once such a usecase arises an one states that one needs this funtionality. Otherwise we implement stuff nobody uses which makes the config less readable and understandable.

Frontrider:
> Well I only partly agree. There are cases where generated files should not be treated as build artifacts IMO. One example are the generated godot api classes we commit in this project. Altough they are generated, they should not be treated as build artefacts. Same goes for classes generated for godot IMO. Well it depends what you do with the generated class. If you add it to a node, it should be commited IMO to track it's api changes over time and check that no change you made, resulted in a api change which affects the function of the node you applied said script. If you however only call said script from let's say GDScript, then its fine IMO to treat it as a build artefact.

On the back of this, it can stay as is and we can declare "don't put files needed by godot to the build folder" as a convention. I'M more or less me overthinking it, but I still think it was worth walking through it. :sweat_smile:

chippmann:
> On the back of this, it can stay as is and we can declare "don't put files needed by godot to the build folder" as a convention. I'M more or less me overthinking it, but I still think it was worth walking through it. sweat_smile

Yeah I think so as well. Once we need it we can still do it more granular.
No definitely worth discussing! :-) Didn't think of the code generation case in the first place!

CedNaru:
Note for when it will be time to put that in 4.0.
We won't need to keep any .kt files in Godot once the new extension system is in place. Right now in 3.x we need to keep those .kt files around as a "tag" to provide the actual Kotlin script stored in the .jar.  But they won't be scripts in 4.0, we won't need to export anything outside the .jar.

chippmann:
Tested the editor on linux and the exports (both debug and release) on linux and android
The .gdignore file do not seem to have any other effect than to hide the files within the editor. They do not affect the visibility of those files to godot internal tools like the export plugins.

piiertho (approved):

CedNaru (approved):

chippmann:
> > They do not affect the visibility of those files to godot internal tools like the export plugins.
> 
> Does that mean that you can still reference files that are in an ignored folder in your code?

Yes. But you might need to manually add it in exports (not speaking of individual scripts but binaries containing those scripts). We don't need that as every file we really need is added by our export plugin explicitly.

Frontrider:
> Yes. But you might need to manually add it in exports (not speaking of individual scripts but binaries containing those scripts). We don't need that as every file we really need is added by our export plugin explicitly.

I see. So the answer is yes, but with a config.

PR #368, opened 2022-10-30
Fix registration of empty abstract classes extending godot classes

### LINKED ISSUE #365: Abstract parent classes need to be registered to godot.
Do not know if it is a feature or a bug.

I have this class:
```
import godot.annotation.RegisterClass
import godot.annotation.RegisterFunction
import godot.global.GD

@RegisterClass("Player")
class Player: BaseEntity() {
    @RegisterFunction
    override fun _ready(){
        GD.print("I am the player!")
    }
}
```
If `BaseClass` is not registered to godot (should not be, as it is an abstract class), then the generated code will reference the non-existant `BaseEntityRegistrar`. 
This is also bad, as because that parent is abstract, the generated code is now wrong, as it tries to get its constructor.

chippmann:
Resolves #365.

### DIFF
--- a/harness/tests/src/main/kotlin/godot/tests/inheritance/AbstractClassInheritanceEmptyParent.kt
+++ b/harness/tests/src/main/kotlin/godot/tests/inheritance/AbstractClassInheritanceEmptyParent.kt
@@ -1,0 +1,6 @@
+package godot.tests.inheritance
+
+import godot.KinematicBody
+
+// issue: https://github.com/utopia-rise/godot-kotlin-jvm/issues/365
+abstract class AbstractClassInheritanceEmptyParent: KinematicBody()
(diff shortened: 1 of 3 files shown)


harness/tests/src/main/kotlin/godot/tests/inheritance/AbstractClassInheritanceEmptyParent.kt
```diff
@@ -0,0 +1,6 @@
+package godot.tests.inheritance
+
+import godot.KinematicBody
+
+// issue: https://github.com/utopia-rise/godot-kotlin-jvm/issues/365
+abstract class AbstractClassInheritanceEmptyParent: KinematicBody()
```
raniejade, line 6: is the missing `@RegisterClass` intentional?

raniejade (approved):

harness/tests/src/main/kotlin/godot/tests/inheritance/AbstractClassInheritanceEmptyParent.kt
chippmann: Yes. Abstract classes don't need to be registered. See: https://godot-kotl.in/en/stable/user-guide/advanced/abstract_classes/

CedNaru (approved):

piiertho (approved):

PR #341, opened 2022-07-15
Implement new rpc related functionality

chippmann:
This changes our rpc related code to support the changes made in godot.

Namely;
- Drop of rset and any other property related rpc calls and configuration
- New rpc config and config retrieval

This also updates the registration code, code generation and the IDE checks.

PR #369, opened 2022-10-31
Implement new rpc configs

chippmann:
This implements the new rpc configs of Godot 4

Replaces #341


src/kt_class.cpp
```diff
@@ -152,12 +161,18 @@ void KtClass::get_signal_list(List<MethodInfo>* p_list) {
     get_member_list(p_list, signal_infos);
 }
 
+const Dictionary KtClass::get_rpc_methods() {
```
piiertho, line 164: Is copy needed by api ?

src/kt_function.cpp
```diff
@@ -31,8 +31,8 @@ int KtFunction::get_parameter_count() const {
     return parameter_count;
 }
 
-Multiplayer::RPCMode KtFunction::get_rpc_mode() const {
-    return method_info->rpc_mode;
+Dictionary KtFunction::get_rpc_config() const {
```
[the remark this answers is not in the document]
CedNaru: Dictionary can be passed as a copy, it's implemented as a smart pointer.

src/kt_class.cpp
CedNaru: same as abovve

src/kt_function.h
```diff
@@ -39,7 +42,7 @@ class KtFunction : public JavaInstanceWrapper<KtFunction> {
 
     StringName get_name() const;
     int get_parameter_count() const;
-    Multiplayer::RPCMode get_rpc_mode() const;
+    Dictionary get_rpc_config() const;
```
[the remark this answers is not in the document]
chippmann: I thought about that as well.
But GDScript does it this way. Hence i followed suit. But I agree, this is not great. Will create a container.

src/kt_function.cpp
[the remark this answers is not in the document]
chippmann: Yeah GDScript does it like this as well. But you guys have more CPP exp. Just tell me what to do xD

src/kt_class.cpp
chippmann: Same comment as above i suppose? And yeah it's a copy by the api.

src/kt_function.h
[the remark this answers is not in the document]
CedNaru: A simple structure would be much better indeed

src/kt_function.cpp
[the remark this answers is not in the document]
piiertho: Forget that, pass by value as it's a reference wrapper.

src/kt_function.h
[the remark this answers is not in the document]
chippmann: Internally we can use a container and expose a dictionary where the api requires it i guess.
I agree with piertho; using a dictionary internally is neither typesafe nor nice to use imo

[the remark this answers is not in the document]
chippmann: @piiertho & @CedNaru I Impülemented an internal RpcConfig struct now. Not sure though if this is A correctly implemented (especially the HashMap in `KtClass` and B if it's really better than before. Please have a look :-)

src/kt_class.h
```diff
@@ -43,8 +43,10 @@ class KtClass : public JavaInstanceWrapper<KtClass> {
 
     bool is_assignable_from(KtClass* p_class) const;
 
+    const Dictionary get_rpc_config();
 private:
     HashMap<StringName, KtFunction*> methods;
+    HashMap<StringName, RpcConfig> rpc_method_configs;
```
CedNaru, line 49: Do we really need that HashMap when we already have the Rpc configs set in the Kfunction instances ? We have a duplicate of data here.

chippmann: True. Will remove the HashMap. Makes it simpler anyways.

CedNaru (approved):

kt/godot-library/src/main/kotlin/godot/registration/Registration.kt
```diff
@@ -284,23 +270,28 @@ class ClassBuilderDsl<T : KtObject>(
...
+                    returnVal = KtPropertyInfo(
+                        _type = returnType.type,
+                        name = "",
+                        className = returnType.className,
+                        _hint = PropertyHint.NONE,
+                        hintString = "",
+                        visibleInEditor = true, // always true. Only used for properties
                     ),
-                    rpcModeId
+                    rpcModeId = rpcModeId,
```
piiertho, line 291: I would expect to have a class matching the cpp struct

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
```diff
@@ -25,7 +25,6 @@ fun Type?.toKtVariantType(): ClassName = when {
         "godot.core.VariantType",
         fqName.substringAfterLast(".").camelToSnakeCase().uppercase(Locale.getDefault())
     )
-    fqName == "kotlin.Unit" -> ClassName("godot.core.VariantType", "NIL")
```
piiertho, line 28: Why do we remove this ?

piiertho (approved):

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
chippmann: because we already have that on line 10: https://github.com/utopia-rise/godot-kotlin-jvm/blob/34fa5fd647a49c96c7f6d6b047f45728c7e8b4c8/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt#L10

kt/godot-library/src/main/kotlin/godot/registration/Registration.kt
chippmann: Indeed. I I'mplemented this now by introducing a KtRpcConfig class. This also greatly simplifies the cpp side by using a JavaInstanceWrapper like we do for the Function/PropertyInfo classes

piiertho (approved):

PR #370, opened 2022-11-01
Android CI/CD tests

chippmann:
This adds CI/CD tests on an android emulator to test the android builds as well.

After this PR is merged, we should be able to do test reports for both desktop and android tests as this new GUT version is able to export the test results in junit format.


chippmann:
I close this as these changes have diverged to far.
After my current work on the CI/CD is done, I'll open a new PR for android export tests

Issue #135, opened 2021-04-16
Custom godot executable is not signed


raniejade:
In windows it will give this warning:

![sign](https://user-images.githubusercontent.com/1041919/115009170-54158180-9eef-11eb-8bb8-2a1b08cecd3f.PNG)

I think something similar will happen in OSX.

CedNaru:
With GDExtension, we no longer provide a custom Godot anymore

PR #374, opened 2022-11-14
feat: Create MacOs editor app on release

piiertho:
This creates macOs editor universal app on release instead of binaries per architecture.

### DIFF
--- a/.github/workflows/deploy-export-template.yaml
+++ b/.github/workflows/deploy-export-template.yaml
@@ -358,7 +358,6 @@ jobs:
           repository: godotengine/godot
           ref: 3.5.1-stable
 
-      # has to be the first step as the download url is only valid for 1 minute! See: https://stackoverflow.com/a/65049722
       - name: Get release export template binary
         uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/deploy-godot-editor-release.yaml
+++ b/.github/workflows/deploy-godot-editor-release.yaml
@@ -13,24 +13,37 @@ jobs:
   deploy-editor-release:
     strategy:
       matrix:
-        name: [ Linux ]
+        name: [ Linux, OSX-x64, OSX-arm64, Windows ]
         include:
           - name: Linux
             # always use the oldest still supported LTS version of github actions. See: https://github.com/utopia-rise/godot-kotlin-jvm/issues/224
-            os: ubuntu-18.04
+            os: ubuntu-20.04
             platform: x11
             binary: godot.x11.opt.tools.64
             cat_command: cat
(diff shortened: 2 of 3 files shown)


CedNaru (approved):

.github/workflows/deploy-export-template.yaml
```diff
@@ -358,7 +358,6 @@ jobs:
           repository: godotengine/godot
           ref: 3.5.1-stable
 
-      # has to be the first step as the download url is only valid for 1 minute! See: https://stackoverflow.com/a/65049722
```
chippmann, line 361: Isn't this the case anymore then?

.github/workflows/deploy-godot-editor-release.yaml
```diff
@@ -110,17 +129,138 @@ jobs:
...
+        name: [ MacOs ]
+        include:
+          - name: MacOs
+            os: macos-latest
+            java-version: 11
+            target: release_debug
+            binary-suffix: opt.tools
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Download release informations
```
chippmann, line 176: Same here i guess?

.github/workflows/deploy-godot-editor-release.yaml
```diff
@@ -110,17 +129,138 @@ jobs:
...
+      matrix:
+        os: [ macos-latest ]
+        include:
+          - os: macos-latest
+            java-version: 11
+            target: release_debug
+            binary-suffix: opt.tools
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Download release informations
```
chippmann, line 220: And here?

chippmann (approved):
Just some minor questions.

chippmann:
@piiertho Another question as I didn't use MacOS in a while; Our app will not be signed with this as far as i can see. Are the users still able to execute it even with the newest MacOS (whatever it's called now)? IIRC they increased the hurdle to execute non signed app's.

.github/workflows/deploy-export-template.yaml
piiertho: IIRC that's the case only between workflows

chippmann:
> > 
> 
> They can, but they have to go in security tag of mac parameters to authorize app :/

We really need to sign our builds somehow.
As said previously I'm happy if we use my apple developer account. I need it anyways and thus we don't need to pay apple any more.  That said: the downside is that then there stands that it's signed by me and not by utopia rise.
Anyways we already have an issue for that: https://github.com/utopia-rise/godot-kotlin-jvm/issues/135

PR #377, opened 2022-11-15
4.0/fix compilation errors

chippmann:
Make our code compile with the changes from the latest master ([245e6454a03a24bf074eb3715776125842e3b666](https://github.com/godotengine/godot/commit/245e6454a03a24bf074eb3715776125842e3b666))


chippmann:
The checks now pass as well. The only one not passing are the actual tests as the entry gen still produces wrong code which is fixed with https://github.com/utopia-rise/godot-kotlin-jvm/pull/369. Hence i won't fix it here.

piiertho (approved):
Few typo, otherwise ok for me, thx a lot !

PR #380, opened 2022-11-17
Export headless editor and server builds

chippmann:
This adds support for:
- headless editor builds (useful and needed for exporting games in CI/CD)
- server builds

This PR is a requirement for #370

### DIFF
--- a/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
+++ b/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
@@ -47,12 +47,17 @@ jobs:
   build-editor-debug:
     strategy:
       matrix:
-        name: [ Linux, OSX-x64, OSX-arm64, Windows ]
+        name: [ Linux, Linux-Editor-Headless, OSX-x64, OSX-arm64, Windows ]
         include:
           # always use the oldest still supported LTS version of github actions. See: https://github.com/utopia-rise/godot-kotlin-jvm/issues/224
           - name: Linux
             os: ubuntu-20.04
             platform: x11
+            arch: 64
+          # always use the oldest still supported LTS version of github actions. See: https://github.com/utopia-rise/godot-kotlin-jvm/issues/224
+          - name: Linux-Editor-Headless
+            os: ubuntu-20.04
+            platform: server
             arch: 64
           - name: OSX-x64
             os: macos-latest
@@ -79,7 +84,7 @@ jobs:
           path: modules/kotlin_jvm
           submodules: recursive
       - name: Linux dependencies
-        if: matrix.platform == 'x11'
+        if: matrix.platform == 'x11' || matrix.platform == 'server'
         shell: bash
         run: |
           # Azure repositories are not reliable, we need to prevent azure giving us packages.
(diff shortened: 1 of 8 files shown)


.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
```diff
@@ -100,19 +105,38 @@ jobs:
...
           target: debug
           tools: true
+
+      - name: Build with headless editor debug
+        if: matrix.platform == 'server'
+        uses: ./.github/actions/godot-build
+        with:
+          sconsflags: arch=${{ matrix.arch }} -j${{ steps.cpu-cores.outputs.count }}
+          platform: ${{ matrix.platform }}
+          target: release_debug
```
piiertho, line 124: Why `release_debug` ? this workflow is here to check target `debug`

.github/workflows/deploy-godot-editor-release.yaml
```diff
@@ -146,6 +154,17 @@ jobs:
...
 
+      - name: Upload zip release
+        if: matrix.platform == 'server'
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.save_release_info.outputs.upload_url }}
+          asset_path: godot-kotlin-jvm_editor_headless_${{ steps.save_release_info.outputs.godot_kotlin_jvm_version }}.zip
+          asset_name: godot-kotlin-jvm_editor_headless_${{ matrix.platform }}_${{ steps.save_release_info.outputs.godot_kotlin_jvm_version }}.zip
```
piiertho, line 165: I don't think creating an extra step only for adding headless is a good idea, as we already have `server` as platform, which already give information.

.github/workflows/deploy-export-template.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
   deploy-export-template:
```
piiertho, line 13: Did you check everything is working when deploying on a fork repo ?

piiertho (changes requested):

.github/workflows/deploy-godot-editor-release.yaml
chippmann: But how do we distinguish between server and headless editor builds then?
Because on the release page we would then have three builds named server hard to distinguish which one is the headless editor and which ones are the server builds

.github/workflows/deploy-export-template.yaml
chippmann: To be honest i didn't. Did not think of the ability to test releases on a fork. Will do that asap (same for the other PR)

chippmann: You were right to ask for testing on a fork. There was an issue with the zip naming. It's now fixed. See: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/untagged-ac089a3da05694148e49

.github/workflows/deploy-godot-editor-release.yaml
piiertho: Will sever have `editor` in its release name ? Because here name starts with `godot-kotlin-jvm_editor`

chippmann: Not the servers, but the headless editor as it's technically still the editor.
I could rename that into just `headless_editor`? Maybe it's more clear then that that one is the headless version of the editor?

.github/workflows/deploy-export-template.yaml
chippmann: Updated url: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.5.3-3.5.1-SNAPSHOT

.github/workflows/deploy-godot-editor-release.yaml
piiertho: What I was meaning is that you don't need to call it editor headless server, editor server is sufficient, this way you do not need two different steps, and make the CI file easier to maintain.

chippmann: Ah now i get it. Yeah true. I'll change it back.

chippmann: Updated url: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.5.3-3.5.1-SNAPSHOT

piiertho (approved):

CedNaru (dismissed):

CedNaru (approved):

piiertho (approved):

PR #381, opened 2022-11-17
Export android x86_64 target

chippmann:
The android x86_64 target is today mainly used for android emulators during development on x86 based host machines because of the increased emulation performance.

They are also a requirement for android apps on windows atm.

### DIFF
--- a/.github/workflows/deploy-export-template.yaml
+++ b/.github/workflows/deploy-export-template.yaml
@@ -13,7 +13,7 @@ jobs:
   deploy-export-template:
     strategy:
       matrix:
-        name: [ Linux-release, OSX-x64-release, OSX-arm64-release, Windows-release, Android-armv7-release, Android-armv8-release, Linux-debug, OSX-x64-debug, OSX-arm64-debug, Windows-debug, Android-armv7-debug, Android-armv8-debug ]
+        name: [ Linux-release, OSX-x64-release, OSX-arm64-release, Windows-release, Android-armv7-release, Android-armv8-release, Android-x86_64-release, Linux-debug, OSX-x64-debug, OSX-arm64-debug, Windows-debug, Android-armv7-debug, Android-armv8-debug, Android-x86_64-debug ]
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-export-template.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
   deploy-export-template:
     strategy:
       matrix:
-        name: [ Linux-release, OSX-x64-release, OSX-arm64-release, Windows-release, Android-armv7-release, Android-armv8-release, Linux-debug, OSX-x64-debug, OSX-arm64-debug, Windows-debug, Android-armv7-debug, Android-armv8-debug ]
+        name: [ Linux-release, OSX-x64-release, OSX-arm64-release, Windows-release, Android-armv7-release, Android-armv8-release, Android-x86_64-release, Linux-debug, OSX-x64-debug, OSX-arm64-debug, Windows-debug, Android-armv7-debug, Android-armv8-debug, Android-x86_64-debug ]
```
piiertho, line 16: Is that wanted to not add it to export apk ?  
Also did you check on a fork this workflow is working well, so that we don't have any surprise when releasing ?

chippmann: Sorry I did not catch what you mean with "not added to export apk"?
Regarding release testing: same as in the other PR :-)

piiertho: > Sorry I did not catch what you mean with "not added to export apk"? Regarding release testing: same as in the other PR :-)

nevermind, you added it to build android template action.

piiertho (approved):

.github/workflows/deploy-export-template.yaml
chippmann: Updated link: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.5.2-3.5.1-SNAPSHOT

PR #382, opened 2022-11-17
Fix api doc generation for parameters

chippmann:
The godot docs now reference member parameters with `[param <parameterName>]` similarly as they do for other references like method references: `[method remove_button]`.

This takes care of that by correctly parsing those references and translate them to kdoc references.

Example:
Before:
```kotlin
  /**
   * Adds a button with label [param text] and a custom [param action] to the dialog and returns the created button. [param action] will be passed to the [customAction] signal when pressed.
   *
   * If `true`, [param right] will place the button to the right of any sibling buttons.
   *
   * You can use [removeButton] method to remove a button created with this method from the dialog.
   */
  public fun addButton(
    text: String,
    right: Boolean = false,
    action: String = ""
  ): Button? {
    // omitted for this example
  }
```
After:
```kotlin
/**
   * Adds a button with label [text] and a custom [action] to the dialog and returns the created button. [action] will be passed to the [customAction] signal when pressed.
   *
   * If `true`, [right] will place the button to the right of any sibling buttons.
   *
   * You can use [removeButton] method to remove a button created with this method from the dialog.
   */
  public fun addButton(
    text: String,
    right: Boolean = false,
    action: String = ""
  ): Button? {
    // omitted for this example
  }
```


CedNaru (approved):

piiertho:
Why this does not build ?

piiertho:
> Why this does not build ?

Never mind, did not see master was not the target

chippmann:
> > Why this does not build ?
> 
> Never mind, did not see master was not the target

So it's fine then? Or do I need to change something?

piiertho:
I think this should build after rebasing on `4.0-adaptation`

kt/godot-library/src/main/kotlin/godot/gen/godot/AESContext.kt
```diff
@@ -104,6 +104,8 @@ import kotlin.Unit
  *
  *     public AESContext Aes = new AESContext();
  *
+ *
+ *
```
piiertho, line 108: Why does it adds two lines here ?

chippmann: That's a good question xD It's not present in the actual doc as far as i can tell. Will investigate that.

PR #390, opened 2022-11-25
Add Vector of properties in Tool Mode

### LINKED ISSUE #384: Exported properties should be sorted by declaration order
Version: `0.5.1-3.5.1`

Exported variables in Kotlin sometimes don't keep order like declared, especially if we use different types.

This non-deterministic behaviour is problematic because Godot Editor has issues with keeping old values in TSCN file when order has changed - values can be randomly nullified, even when class is written in GDScript. When "re-order" occurs, properties need to be manually assigned again in editor.

Code to reproduce:

```gdscript
extends Node

export var a: NodePath
export var b: Texture
export var c: Texture
```

```kotlin
@RegisterClass
class Test : Node() {

	@Export
	@RegisterProperty
	var a: NodePath? = null

	@Export
	@RegisterProperty
	var b: Texture? = null

	@Export
	@RegisterProperty
	var c: Texture? = null

}
```

Screenshot shows invalid order:

export_order_issue

Expected behaviour:

Kotlin properties order should be `A`, `B`, `C`.

CedNaru:
fix #384 
Properties are now set in a OrderedHashMap, so we can retrieve them in the order they were declared in script.

Wonder if signals and functions also require the same treatment. Properties matter for the inspector but I don't think there is an equivalent for the rest. 
Ordered hashmaps are also slower, switching functions to one would slow down calls to the JVM a bit (probably not noticeable but still slower)

### DIFF
--- a/src/kt_class.cpp
+++ b/src/kt_class.cpp
@@ -19,6 +19,10 @@ KtClass::~KtClass() {
     for (auto& constructor : constructors) {
         delete constructor;
     }
+
+#if TOOLS_ENABLED
+    property_list.clear();
+#endif
 }
 
 KtObject* KtClass::create_instance(jni::Env& env, const Variant** p_args, int p_arg_count, Object* p_owner) {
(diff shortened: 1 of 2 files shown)


src/kt_class.cpp
```diff
@@ -13,9 +13,18 @@ KtClass::KtClass(jni::JObject p_wrapped, jni::JObject& p_class_loader) :
 }
 
 KtClass::~KtClass() {
+    {
+        OrderedHashMap<StringName, KtProperty*>::Element element {properties.front()};
```
piiertho, line 17: Can't we use for each ?

piiertho (commented):
I'm really suspicious regarding performances, also it seems to be needed for tool only.

CedNaru:
Changed it for an extra Vector in ToolMode

piiertho (approved):

src/kt_class.cpp
CedNaru: not implemented for the 3.x HahMaps

CedNaru:
Tested and it works, the inspector displays properties in the right order.

chippmann (approved):

PR #393, opened 2022-11-29
Feature/4.0/implement vector4

piiertho:
This implements `Vector4` and `Vector4i` classes.


chippmann (approved):
LGTM, though hard to tell without working unit tests but if there is an issue, we can always revise it after the tests are working.

CedNaru (approved):

PR #395, opened 2022-11-29
Fix resource property type hint

### LINKED ISSUE #394: Exported property type is ignored in Inspector tab
There are too many options so you have to scroll a lot.

Equivalent GDScript code (e.g. `export var test: Texture`) shows only inherited types.

export_type_issue

chippmann:
Fixes #394

This fixes the type hints for resource types.
The issue we were having was two fold:
- For once we did not check the supertypes recursively to detect whether the property was a resource type (in fact the relevant type is `Reference`). We also had that problem for list types. I fixed it there too as well.
- Secondly, since we initially implemented this, two new type hints were added, which meant that the ordinal was not correct anymore. By adding the missing typehints, the ordinal we send to the cpp code is correct again. **Note:** We should investigate if we can either automatically generate these typehint enums, or automatically test whether they are still correct. But I think this is out of the scope for this fix.

### DIFF
--- a/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
+++ b/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
@@ -54,7 +54,7 @@ fun Type.toTypeName(): TypeName = ClassName(
 
 fun Type.isCompatibleList(): Boolean = when (fqName) {
     "godot.core.GodotArray", "godot.core.VariantArray" -> true
-    else -> supertypes.any { it.fqName == "godot.core.GodotArray" || it.fqName == "godot.core.VariantArray" }
+    else -> supertypes.any { it.isCompatibleList() }
 }
 
 fun Type.isReference(): Boolean = fqName == "godot.Reference" ||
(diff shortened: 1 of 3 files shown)


kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/ext/TypeExtensions.kt
```diff
@@ -54,14 +54,13 @@ fun Type.toTypeName(): TypeName = ClassName(
 
 fun Type.isCompatibleList(): Boolean = when (fqName) {
     "godot.core.GodotArray", "godot.core.VariantArray" -> true
```
CedNaru, line 56: I think "godot.core.GodotArray" is not even in the codebase anymore. There was a time when we had 2 different array classes, but it was removed a long time ago.

kt/godot-library/src/main/kotlin/godot/core/Functions.kt
```diff
@@ -17,8 +17,10 @@ enum class PropertyHint {
     FLAGS, ///< hint_text= "flag1,flag2,etc" (as bit flags)
     LAYERS_2D_RENDER,
     LAYERS_2D_PHYSICS,
+    LAYERS_2D_NAVIGATION,
```
CedNaru, line 20: Before we had to copy multiple enums in the godot library because the generated code was in a different module but now that everything is together, wouldn't that be better to directly use the one available in gen/ ?

chippmann: I agree. As discussed on Discord I will look into this and file a separate PR.

CedNaru (approved):

CedNaru (approved):

piiertho (approved):

piiertho:
Did not see but we seem to have a problem in our tests on array

chippmann:
> Did not see but we seem to have a problem in our tests on array

WDYM? I don't see anything in the actions in this PR?

PR #420, opened 2023-01-05
bugfix: fix invalid delete in MemoryBridge::decrement_ref_counter

piiertho:


CedNaru (approved):
Don't hit me.
I tend to forget Godot uses its own new/delete functions

PR #422, opened 2023-01-11
bugfix(4.0): Delete KtObject when destroying instance binding without script instance

piiertho:


CedNaru (approved):
When you find the issue but only fix half of it :p.

src/memory/kotlin_binding_manager.cpp
```diff
@@ -21,13 +21,17 @@ void KotlinBindingManager::_instance_binding_free_callback(void* p_token, void*
     // Called in the destructor of the Object.
     //  It's the very last action done in the destructor so assume variables local to the Object have been cleaned (including script and extension).
     // There are 2 cases, either an Object has been freed, and we have to release its reference OR it's a Refcounted and the JVM instance is already dead.
+    KtObject* binded_kt_object{reinterpret_cast<KotlinBinding*>(p_binding)->kt_object};
```
chippmann, line 24: nit: i'd suggest `bound_kt_object`. Is as far as i can tell gramatically correcter? ^^

chippmann (approved):

PR #426, opened 2023-01-15
bugfix(4.0): reset KtObject's initConfig after checking if need bind

piiertho:


CedNaru (changes requested):
Just to avoid any other mistakes in case we modify init() again, I would move the call to reset to the end of the instantiateWith() method.

CedNaru:
Okay forget what I said, I totally forgot why we reset the config here and not later. It's to deal with nested constructions of objects.
If your Kotlin script init{} method also creates other objects, the config has to be reset otherwise they will take the wrong path in the init of KtObject.

CedNaru (approved):

PR #430, opened 2023-01-20
chore(4.0): update array api with new godot 4 methods

piiertho:


kt/godot-library/src/main/kotlin/godot/core/IReadOnlyVariantArray.kt
```diff
@@ -0,0 +1,28 @@
+package godot.core
+
+interface IReadOnlyVariantArray<T> : Collection<T> {
+    fun isReadOnly(): Boolean
+    fun isTyped(): Boolean
+    fun hash(): Int
+
+    fun all(callable: Callable): Boolean
+    fun any(callable: Callable): Boolean
+    fun filter(callable: Callable) : VariantArray<T>
```
CedNaru, line 10: Should return a ReadOnlyVariant here. Then in the actual implemented, the return type is changed to a regularVariantArray.

kt/godot-library/src/main/kotlin/godot/core/IReadOnlyVariantArray.kt
```diff
@@ -0,0 +1,28 @@
+package godot.core
+
+interface IReadOnlyVariantArray<T> : Collection<T> {
```
CedNaru, line 3: I wouldn't use a "I" prefix here.  Because it means that users will have to manipulate the type IReadOnlyVariantArray or VariantArray side by side. This is not consistent.

kt/godot-library/src/main/kotlin/godot/core/VariantArray.kt
```diff
@@ -472,3 +688,10 @@ inline fun <reified T> Iterable<T>.toVariantArray() = VariantArray<T>().also { a
         arr.append(it)
     }
 }
+
+inline fun <reified T> VariantArray<T>.asReadOnly() = this.also { setReadOnly(true) } as IReadOnlyVariantArray<T>
+inline fun <reified T> IReadOnlyVariantArray<T>.asMutable(): VariantArray<T> {
+    require(this is VariantArray<T>)
+    this.setReadOnly(true)
```
CedNaru, line 695: should be false here.

CedNaru (changes requested):
Overall, not big on how that whole ReadOnly thing is handled.
If you have a regular mutable VariantArray and call `asReadyOnly()` to return it to some other parts of the code, you will set the readOnly mode on it, making changes disabled but the array still exist as a mutable VariantArray elsewhere but can't act as one anymore.

We are not in the same case with regular polymorphism where you return a mutable as an immutable so only a subset of the API is available to whatever receives it. Here you are actively making the mutable part of the API invalid to use.

Kotlin's polymorphism is already enough as a security measure here.  The` setReadOnly()` and `setTyped()` methods exist because there is no actual generic and inheritance for core types in GDScript ( and in the internal handling of variant.h indirectly) so readOnly and Typed are handled in this weird way because they also have to accommodate languages with dynamic types.
I think we don't need to implement any of those Type/ReadOnly api functions on the Kotlin side.

CedNaru (approved):

PR #431, opened 2023-01-25
feat(4.0/api-gen): Use meta from api json for int and float while generating kotlin api

piiertho:


kt/api-generator/src/main/kotlin/godot/codegen/constants/Classes.kt
```diff
@@ -136,6 +136,20 @@ object GodotTypes {
...
 
+object GodotMeta {
+    object Int {
+        const val uint32 = "uint32"
+        const val int32 = "int32"
+        const val uint64 = "uint64"
+        const val int64 = "int64"
+    }
+
+    object Float {
```
CedNaru, line 147: I only meant for Integers to use meta data, not for float.
The reasoning is that the "normal" types in Kotlin are actually Integer and Double. 
In Godot 3, we had no information about the Integer values being 32 or 64 bits which forced us to use Long everywhere, which is against the regular type in Kotlin and was forcing users to put `.toLong()` despite Godot API mostly use 32bits.
Now with meta that issue can finally be solved. But you applied the same treatment to Floating values, turning all Double values to Float. 
The final result is that we reversed the situation for Integers and Floats.

That change was not meant for any gain of performance or memory as the internal buffer and variant are always going to use 64bits values anyway. It was simply to avoid users to cast all the time.

kt/godot-library/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -642,6 +642,18 @@ enum class VariantType(
         }
     ),
 
+    JVM_ULONG(
```
CedNaru, line 645: I'm not for overloading Variant with even more fake types that are not part of the actual C++ Variant. We introduced JVM_INT in the first place so people can register methods with Int parameters instead of only Long. But IMO it was a poor way to implement it so I don't wish to make it worse.
 It's not like we are supporting registering unsigned integers in the first place so it's not necessary to provide a 2 way conversion.
 The casting shouldn't directly be in the implementation of the api methods when necessary.

kt/api-generator/src/main/kotlin/godot/codegen/constants/CoreTypeTraits.kt
```diff
@@ -5,5 +5,6 @@ import godot.codegen.traits.TypedTrait
 object CoreTypeTraits {
     val VECTOR3 = object : TypedTrait {
```
CedNaru, line 6: More of a question here because it was already there before, but why do we have only one for Vector3 and none of the other core types ?

kt/api-generator/src/main/kotlin/godot/codegen/traits/TypedTrait.kt
```diff
@@ -2,4 +2,5 @@ package godot.codegen.traits
 
 interface TypedTrait {
     val type: String?
+    val meta: String?
```
CedNaru, line 5: Should that really be part of the TypeTrait when it's used for a single class and forced to null for everything else ?

kt/godot-library/src/main/kotlin/godot/global/GDRandom.kt
```diff
@@ -8,20 +8,15 @@ internal interface GDRandom {
     var rng: RandomNumberGenerator?
 
     /** Random range, any floating point value between from and to. */
-    fun randRange(from: Float, to: Float) = randRange(from.toDouble(), to.toDouble()).toFloat()
-
-    /** Random range, any floating point value between from and to. */
-    fun randRange(from: Double, to: Double) = rng!!.randfRange(from, to).toDouble()
+    fun randRange(from: Float, to: Float) = rng!!.randfRange(from, to).toDouble()
```
CedNaru, line 11: I created 2 versions of those functions so people can freely use Float or Double for whatever is the most convenient at the moment. I don't think we should remove them. Kotlin's Math package is also full of doubles like this. 
We don't gain anything from removing more convenience for users.

kt/api-generator/src/main/kotlin/godot/codegen/constants/Classes.kt
```diff
@@ -136,6 +136,20 @@ object GodotTypes {
     )
 }
 
+object GodotMeta {
+    object Int {
+        const val uint32 = "uint32"
```
CedNaru, line 141: Same as Float, but I don't think that adding the unsigned integers is a good idea. Now we have many part of the Godot api that went from being regular integer to unsigned which will force many casting again.
If look at the usages of unsigned integers in the API, you'll see that it's not because they need this extra bit of data to get bigger values. It just means it can't be negative but in practice it has little value.
There is no benefit in forcing people to use an unsigned long for flags when connecting to a signal for example.

kt/godot-library/src/main/kotlin/godot/core/Variant.kt
piiertho: Exactly, JVM_INT is done the same way, adding cast instead would just break api gen flow.
If the reason for JVM variant being "poor" implementation is only "this does not exists in cpp" I think you should reconsider it. This implementation makes entry gen and api gen code much easier

kt/api-generator/src/main/kotlin/godot/codegen/traits/TypedTrait.kt
piiertho: Yes, that's a type information, and should be used with TypedTrait extensions

chippmann (changes requested):
After some discussions on Discord, I agree with @CedNaru : I would only use the metadata for convenience casting the "Int" types. Not the "Real" types.
Apart from that, LGTM

chippmann (approved):

CedNaru (approved):

PR #432, opened 2023-02-01
Extract constants from api-gen to separate project and use it in entry-gen and IDE-Plugin

chippmann:
This extracts the contents of the `constants` and parts of the `extensions` package of the api-gen into its own gradle project `tools-common`.

This is done to commonize all hardcoded occurences of fqNames from godot and our own.

This also made a slight restructuring of our gradle project setup necessary as outlined on Discord. Quick recap:
The api-gen is a gradle plugin. As such it is a separate gradle project included in our main project through the gradle composit builds feature. As such it has no direct access to any other modules in our main project. But with this PR it needs access to the new `tools-common` project as it depends on it. As a result, the `tools-common` cannot just be a module, it needs to be a separate gradle project included in the build with gradles composite build feature (just like the api-gen).
Furthermore as the `tools-common` is now a dependency of both the entry-gen and the IDE plugin, it needs to be published. But our gradle publish plugin resides in `buildSrc` which is not accessible by `tools-common` as it is a separate gradle project. Thus our publish plugin was made a separate gradle project as well.
Furthermore as the `tools-common` is published, it also needs the proper versioning. This was also done through `buildSrc` which, as said previously` is also not accessible from `tools-common` as it is a separate gradle project. As a consequence, this part was extracted into its own gradle plugin called `godot-dependenices`.

To summarize:
- `api-gen` is still a gradle plugin and as such a separate gradle project included through composite build
- `tools-common` is a new separate gradle project included through composite build
- our publish plugin is extracted from `buildSrc` into its own gradle plugin project included through composite build
- the version definition and depdenency version declaration is moved from `buildSrc` into its own gradle plugin project included through composite build
- constants are moved from `api-gen` to `tools-common`
- all hardcoded occurences of fqNames in the entry-gen and IDE plugin are replaced with the constants from `tools-common`

**Note:** This PR does not address anything in regards to the trait system introduced by @piiertho. This will be looked at in a separate PR after this one as it's already large enough.

PR #433, opened 2023-02-02
Make entry gen fully language independent

chippmann:
Depends on #432 . (I will change the target branch once that one is merged)

Makes entry gen (hopefully) fully language independent by removing all remaining kotlin fqName checks from it and replacing it by a fqName provider provided to the entry gen by the caller.

With this the entry gen retrieves the fqNames it needs for certain language specific type checks (like "is the fqName i got here an fqName of the Int type or not?") from the caller through the provider.

This should allow the language specific symbol processors to provide the entry gen with the language specific fqNames for the needed types.

### DIFF
--- a/kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/utils/JvmTypeProvider.kt
+++ b/kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/utils/JvmTypeProvider.kt
@@ -1,0 +1,38 @@
+package godot.annotation.processor.utils
+
+import godot.entrygenerator.model.JvmType
+
+class JvmTypeProvider: (JvmType) -> Set<String> {
+    override fun invoke(jvmType: JvmType): Set<String> {
+        return when(jvmType) {
+            JvmType.INT -> setOf(
+                requireNotNull(Int::class.qualifiedName),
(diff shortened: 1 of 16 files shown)


kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/utils/JvmTypeProvider.kt
```diff
@@ -0,0 +1,38 @@
+package godot.annotation.processor.utils
+
+import godot.entrygenerator.model.JvmType
+
+class JvmTypeProvider: (JvmType) -> Set<String> {
```
piiertho, line 5: Why not creating a top level lambda instead ?

piiertho (approved):
Only one comment, otherwise this is really nice !

kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/utils/JvmTypeProvider.kt
chippmann: Just because of lazy initialization of top level members. Once the class file is loaded, the lambda stays around forever even if not used anymore.
In this case really not relevant as the whole lifecycle of it is bound to the build process anyways so i would be fine if we make a top level lambda out of it.

PR #434, opened 2023-02-08
Feature/4.0/default values from instance

piiertho:
This cause troubles with exported `lateinit`.  
When it tries to get it, it throws an exception because it is not initialized.  
```
Godot-JVM: Try to create res://src/main/kotlin/godot/tests/Invocation.kt instance.
Hello Invocation!
Godot-JVM: Instantiated an object of type res://src/main/kotlin/godot/tests/Invocation.kt
kotlin.UninitializedPropertyAccessException: lateinit property lateinitString has not been initialized
        at godot.tests.Invocation.getLateinitString(Invocation.kt:77)
        at godot.godot.tests.InvocationRegistrar$register$1$1$87.get(InvocationRegistrar.kt:279)
        at godot.core.KtProperty.callGet(Properties.kt:33)
ERROR: Godot-JVM: An exception has occurred!
   at: (modules\kotlin_jvm\src\jni\env.cpp:68)
kotlin.UninitializedPropertyAccessException: lateinit property registerObject has not been initialized
        at godot.tests.Invocation.getRegisterObject(Invocation.kt:80)
        at godot.godot.tests.InvocationRegistrar$register$1$1$89.get(InvocationRegistrar.kt:280)
        at godot.core.KtProperty.callGet(Properties.kt:33)
ERROR: Godot-JVM: An exception has occurred!
   at: (modules\kotlin_jvm\src\jni\env.cpp:68)
Godot-JVM: Try to create res://src/main/kotlin/godot/tests/Invocation.kt instance.
Hello Invocation!
Godot-JVM: Instantiated an object of type res://src/main/kotlin/godot/tests/Invocation.kt
kotlin.UninitializedPropertyAccessException: lateinit property lateinitString has not been initialized
        at godot.tests.Invocation.getLateinitString(Invocation.kt:77)
        at godot.godot.tests.InvocationRegistrar$register$1$1$87.get(InvocationRegistrar.kt:279)
        at godot.core.KtProperty.callGet(Properties.kt:33)
ERROR: Godot-JVM: An exception has occurred!
   at: (modules\kotlin_jvm\src\jni\env.cpp:68)
```
The default value is set to null (as it is not initialized it results in a NIL variant), even if property type is not nullable.
I think this is misleading for user and can lead to big troubles.

When thinking with @CedNaru  we went to a point that's a non sense to export lateinit. As exports needs default values that should be resolved during construct time of Object.

----
Edit @chippmann:
**Documentation:** after some discussion on discord we decided to keep the lateinit case for exported properties. There are still some valid usecases where it makes sense to export a lateinit property. Like exporting nodepaths where one expects them to be set from the inspector.
To circumvent the problem described in the PR description, this PR now catches exceptions while creating the instance of the class and while retrieving the default values from it. If an exception arises, the types default value is used instead.
This also gives some safety when the user does something fishy during the constructing phase of the class which leads to an exception if the class is being constructed outside of its intended usage (like here in the editor)

### DIFF
--- a/src/jni/types.cpp
+++ b/src/jni/types.cpp
@@ -99,6 +99,10 @@ namespace jni {
         env.check_exceptions();
     }
 
+    void JObject::call_void_method_noexcept(Env& env, MethodId method, jvalue* args) const {
+        env.env->CallVoidMethodA((jclass) obj, method, args);
+    }
+
     MethodId JClass::get_method_id(Env& env, const char* name, const char* signature) {
         auto id = env.env->GetMethodID((jclass) obj, name, signature);
         JVM_CRASH_COND_MSG(id == nullptr, vformat("Method not found: %s with signature: %s", name, signature));
--- a/src/jni/types.h
+++ b/src/jni/types.h
@@ -76,6 +76,8 @@ namespace jni {
 
         void call_void_method(Env& env, MethodId method, jvalue* args = {}) const;
 
+        void call_void_method_noexcept(Env& env, MethodId method, jvalue* args = {}) const;
+
         bool is_null();
     };
 
--- a/src/kotlin_instance.cpp
+++ b/src/kotlin_instance.cpp
@@ -23,7 +23,7 @@ bool KotlinInstance::set(const StringName& p_name, const Variant& p_value) {
 
     KtProperty* ktProperty {kt_class->get_property(p_name)};
     if (ktProperty) {
-        ktProperty->setCall(binding->kt_object, p_value);
+        ktProperty->call_set(binding->kt_object, p_value);
         return true;
     } else {
         return false;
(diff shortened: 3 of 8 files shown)


CedNaru (approved):

CedNaru (approved):

src/jni/types.cpp
```diff
@@ -99,6 +99,10 @@ namespace jni {
         env.check_exceptions();
     }
 
+    void JObject::call_void_method_noexcept(Env& env, MethodId method, jvalue* args) const {
```
chippmann, line 102: Nit: I'd prefer no shortform here. To me it's not obvious by name what the method does. But I'm fine either way

chippmann (approved):

chippmann:
Documentation: after some discussion on discord we decided to keep the lateinit case for exported properties. There are still some valid usecases where it makes sense to export a lateinit property. Like exporting nodepaths where one expects them to be set from the inspector.
To circumvent the problem described in the PR description, this PR now catches exceptions while creating the instance of the class and while retrieving the default values from it. If an exception arises, the types default values is used instead. With this we also have safety if the user does something fishy in the constructing phase of the class (like dependency injection in the init block rather than in a notification function or reading files or whatever external action might be done).

PR #440, opened 2023-03-05
Avoid singletons gc registration when sent from cpp first

piiertho:
Avoid registration of singletons when those are sent from cpp first.
Set ObjectId when constructing singletons.


chippmann (approved):

kt/godot-library/src/main/kotlin/godot/gen/godot/AESContext.kt
```diff
@@ -178,7 +178,7 @@ public open class AESContext : RefCounted() {
   public fun start(
     mode: Mode,
     key: PackedByteArray,
-    iv: PackedByteArray = PackedByteArray()
```
CedNaru, line 181: This PR added a comma at the end of the last parameter of every function in the API. It should be fixed, it's a lot of noise for nothing.

kt/godot-library/src/main/kotlin/godot/core/KtObject.kt
```diff
@@ -48,11 +48,13 @@ abstract class KtObject {
             //Native object already exists, so we know the id and ptr without going back to the other side.
             rawPtr = config.ptr
             id = config.id
-            //Singletons are never initialized from here.
-            if (config.needBind) {
```
CedNaru, line 52: I am not fond of reintroducing the IsSingleton function after removing it (on top of adding a second layer of branching in the init of KtObject). Once we add the option for people to generate their own API based on the GDExtensions in their Godot projects, we won't be able to keep it with internal visibility.

CedNaru (changes requested):
Overall I do not like the approach here.
I don't want to transform singletons into an exception in the regular initialization code of the KtObjects.
Something I would do is to create a function initializing all singletons on the JVM side ,that would be called when our module starts, just like we do with the rest of the API when constructors and functions index are set.

kt/godot-library/src/main/kotlin/godot/gen/godot/AESContext.kt
piiertho: I don't know why this adds comma, seems that's a strange behaviour from Kotlin Poet.

PR #442, opened 2023-03-07
Just testing

chippmann:


chippmann:
Issue i had locally is fixed in https://github.com/utopia-rise/godot-kotlin-jvm/commit/c0612d52673c9a71ef2033b25671935362c76ea4

PR #448, opened 2023-03-07
Fix update export

chippmann:
This fixes a bug where after a property change in the inspector and saving, the resulting scene file would contain invalid default values for not exported properties.

For example a change on the property `nullable_long` in `Invocation` would also save (wrong) default values for `register_object_nullable_pre_init`, `register_object_non_nullable_pre_init`, `invocation` even though they are not even exported.

This only updates exports (and thus also default values) for properties which are actually exported.

### DIFF
--- a/src/kotlin_script.cpp
+++ b/src/kotlin_script.cpp
@@ -240,15 +240,23 @@ void KotlinScript::_update_exports(PlaceHolderScriptInstance* placeholder) {
         KotlinInstance* script_instance {
           dynamic_cast<KotlinInstance*>(_instance_create({}, 0, tmp_object))};
 
-        List<PropertyInfo> properties;
-        get_script_property_list(&properties);
-        for (int i = 0; i < properties.size(); ++i) {
+        List<PropertyInfo> all_properties;
+        get_script_property_list(&all_properties);
+
+        List<PropertyInfo> exported_properties;
+        for (const PropertyInfo& property_info : all_properties) {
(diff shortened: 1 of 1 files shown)


src/kotlin_script.cpp
```diff
@@ -240,15 +240,23 @@ void KotlinScript::_update_exports(PlaceHolderScriptInstance* placeholder) {
...
 
-        List<PropertyInfo> properties;
-        get_script_property_list(&properties);
-        for (int i = 0; i < properties.size(); ++i) {
+        List<PropertyInfo> all_properties;
+        get_script_property_list(&all_properties);
+
+        List<PropertyInfo> exported_properties;
+        for (const PropertyInfo& property_info : all_properties) {
+            if (property_info.usage != PropertyUsageFlags::PROPERTY_USAGE_NO_EDITOR) {
```
piiertho, line 248: I think this should be `!(property_info.usage & PropertyUsageFlags::PROPERTY_USAGE_NO_EDITOR)`

chippmann: Actually it should be `property_info.usage & PropertyUsageFlags::PROPERTY_USAGE_EDITOR` as `PROPERTY_USAGE_DEFAULT` is defined as `PROPERTY_USAGE_DEFAULT = PROPERTY_USAGE_STORAGE | PROPERTY_USAGE_EDITOR` and `PROPERTY_USAGE_NO_EDITOR` as `PROPERTY_USAGE_NO_EDITOR = PROPERTY_USAGE_STORAGE`

piiertho (approved):

PR #451, opened 2023-03-09
Fix reloading

chippmann:
This fixes multiple bugs with class reloading in the godot editor:

# Fixed Issues
## Reloading of code changes or new classes
We had one issue which was already present in our 3.x releases:
If a scripts resource file was reloaded before our the registration process finished, the editor would not update the displayed value until the scene was closed and reopened (and sometimes not even then).
This case could happen if the user either edited the script from within the godot editor rather than IDEA or returned to the godot editor before the build was done. Or even without building and then building from within the godot editor.
In these cases the resource file was loaded and the exports updated before the changes were registered in the engine. Leading to outdated data be displayed.

This is fixed by manually updating the exports after we're  finished with the registration process.
This fix requires the second fix listed below as we need the loaded `KotlinScript` instances to update them.

This fix could be back ported to the 3.x releases if we so desire

## Fix 4.0 scene corruption issue
With our current 4.0 adaptation branch we have the following issue: when the user edits a script and builds while said script is open in the godot editor (either via scene or in the script editor), the script instance would change upon reloading in the editor. This leads to the current script being marked as "not saved" and the new script treated as separate script. So we now have two different scripts in memory. When the user now saves the old (marked as "not saved") script, godot treats it as a scene local script as the former resource path was taken over by the new script.
As that's not possible, godot corrupts the scene file, by marking the the referenced script as scene local and deleting the resource path:
2023-03-09-163525_2560x1414_scrot

This is fixed by introducing a `KotlinScriptCache` like GDScript does. This cache keeps all loaded `KotlinScript` references and only updates them rather than replacing them.

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -2,6 +2,7 @@
 
 #include "jni/class_loader.h"
 #include "memory/bridges_manager.h"
+#include "kotlin_script_cache.h"
 
 #include <core/config/project_settings.h>
 #include <core/io/resource_loader.h>
@@ -392,6 +393,10 @@ void GDKotlin::finish() {
         garbage_collector_instance.call_void_method(env, clean_up_method_id);
     }
 
+#ifdef TOOLS_ENABLED
+    KotlinScriptCache::invalidate();
+#endif
+
     LongStringQueue::destroy();
     BridgesManager::get_instance().delete_bridges();
 
@@ -526,6 +531,22 @@ void GDKotlin::register_members(jni::Env& p_env) {
     for (const KeyValue<StringName, KtClass*>& item : classes) {
         item.value->fetch_members();
     }
+
+#ifdef TOOLS_ENABLED
+    // updates all exports for all classes loaded in the editor and makes sure the values are updated
+    //
+    // Reason: godot only does that for us when the resource file is updated, but not when the classes are reloaded but
+    // the resource files have not changed. Such a case is, when the user edits the script and goes back to the godot
+    // editor before building and presses build there, or goes back to the editor before the build is done. Then the
+    // last default value is loaded but the new ones are never loaded after registering.
+    //
+    // Downside: in other cases the exports might get updated twice. Case being, the user changes a script and the build
+    // finishes before returning to the godot editor
+    Vector<Ref<KotlinScript>> kotlin_scripts = KotlinScriptCache::get_cached_scripts();
+    for (Ref<KotlinScript>& kotlin_script : kotlin_scripts) {
+        kotlin_script->update_exports();
+    }
+#endif
 }
 
 bool GDKotlin::check_configuration() {
--- a/src/kotlin_script_cache.cpp
+++ b/src/kotlin_script_cache.cpp
@@ -1,0 +1,41 @@
+#ifdef TOOLS_ENABLED
+
+#include "kotlin_script_cache.h"
+
+KotlinScriptCache& KotlinScriptCache::get_instance() {
+    static KotlinScriptCache instance;
+    return instance;
+}
+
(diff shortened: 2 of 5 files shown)


CedNaru (approved):
LGTM but will that PR still make sense when DummyFiles are merged ?

chippmann:
> LGTM but will that PR still make sense when DummyFiles are merged ?

Yes. still the same issue there and the solution works there as well (I originally implemented it there and then changed the base of the branch to 4.0-adaptation)

src/gd_kotlin.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "jni/class_loader.h"
 #include "memory/bridges_manager.h"
+#include "kotlin_script_cache.h"
```
piiertho, line 5: Missing `#ifdef TOOLS_ENABLED`

src/kotlin_script_cache.cpp
```diff
@@ -0,0 +1,41 @@
+#ifdef TOOLS_ENABLED
+
+#include "kotlin_script_cache.h"
+
+KotlinScriptCache& KotlinScriptCache::get_instance() {
+    static KotlinScriptCache instance;
+    return instance;
+}
+
+Ref<KotlinScript> KotlinScriptCache::get_or_create_script(const String &p_path) {
```
piiertho, line 10: format `const String &p_path` => `const String& p_path`

src/kotlin_script_cache.cpp
```diff
@@ -0,0 +1,41 @@
...
+    MutexLock lock(KotlinScriptCache::get_instance().mutex);
```
piiertho, line 11: use uniform init.

src/kotlin_script_cache.cpp
```diff
@@ -0,0 +1,41 @@
...
+        Ref<KotlinScript> ref_new {memnew(KotlinScript) };
+        KotlinScriptCache::get_instance().script_cache[p_path] = ref_new;
+        ref = ref_new;
+    }
+
+    return ref;
+}
+
+void KotlinScriptCache::invalidate() {
+    MutexLock lock(KotlinScriptCache::get_instance().mutex);
```
piiertho, line 27: use uniform init

src/kotlin_script_cache.cpp
```diff
@@ -0,0 +1,41 @@
...
+    KotlinScriptCache::get_instance().script_cache.clear();
+}
+
+const Vector<Ref<KotlinScript>> KotlinScriptCache::get_cached_scripts() {
+    MutexLock lock(KotlinScriptCache::get_instance().mutex);
```
piiertho, line 32: use uniform init.

src/kotlin_script_cache.h
```diff
@@ -0,0 +1,33 @@
+#ifndef GODOT_KOTLIN_SCRIPT_CACHE_H
+#define GODOT_KOTLIN_SCRIPT_CACHE_H
+
+#include <core/string/ustring.h>
+#include "kotlin_script.h"
+
+#ifdef TOOLS_ENABLED
```
piiertho, line 7: place this ifdef at the top of file.

src/kt_resource_format_loader.h
```diff
@@ -1,6 +1,7 @@
 #ifndef GODOT_JVM_KT_RESOURCE_LOADER_H
 #define GODOT_JVM_KT_RESOURCE_LOADER_H
 #include <core/io/resource_loader.h>
+#include "kotlin_script.h"
```
piiertho, line 4: why is this include needed ?

piiertho (changes requested):

piiertho (approved):

PR #453, opened 2023-03-10
Add function args count check

chippmann:
This adds a similar check to function argument counts as we already have for constructors

### DIFF
--- a/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/FunctionArgCountCheck.kt
+++ b/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/FunctionArgCountCheck.kt
@@ -11,7 +11,7 @@ class FunctionArgCountCheck(logger: Logger, sourceFiles: List<SourceFile>): Base
             .flatMap { it.functions }
             .forEach { registeredFunction ->
                 if (registeredFunction.parameters.size > 5) {
-                    logger.error("RegisteredFunction ${registeredFunction.fqName} has more than 5 arguments. More than that is currently not supported as it's very inefficient to unpack that many values. If you need more, either wrap them in a wrapper object or pass a VariantArray containing your values.")
+                    logger.error("RegisteredFunction ${registeredFunction.fqName} has more than 5 arguments. More than that is currently not supported. If you need more, either wrap them in a wrapper object or pass a VariantArray containing your values.")
                     hasIssues = true
                 }
             }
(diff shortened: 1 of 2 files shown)


CedNaru (commented):
Wouldn't that be a good opportunity to put the maximum number of argument as a constant in the new common tool module?

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/FunctionArgCountCheck.kt
```diff
@@ -0,0 +1,17 @@
...
+import godot.entrygenerator.utils.Logger
+
+class FunctionArgCountCheck(logger: Logger, sourceFiles: List<SourceFile>): BaseCheck(logger, sourceFiles) {
+    override fun execute() {
+        sourceFiles
+            .flatMap { it.registeredClasses }
+            .flatMap { it.functions }
+            .forEach { registeredFunction ->
+                if (registeredFunction.parameters.size > 5) {
+                    logger.error("RegisteredFunction ${registeredFunction.fqName} has more than 5 arguments. More than that is currently not supported as it's very inefficient to unpack that many values. If you need more, either wrap them in a wrapper object or pass a VariantArray containing your values.")
```
CedNaru, line 13: I don't think it's accurate to say it's "very inefficient" to unpack many values. The cost is purely linear with the number of argument, it's not like jumping from 5 to 6 arguments would be a major cost increase.
I think the original reason why we limited to 5 arguments was because it was an actual limit in the Godot C++ code, but if I remember correctly that restriction is lifted in 4.0, you can basically use as many arguments as you want (Thanks to macro/template black magic).

piiertho: So maybe we should instead drop this limitation in our cpp code

chippmann:
> Wouldn't that be a good opportunity to put the maximum number of argument as a constant in the new common tool module?

I don't remember that we have any other usage of the args count in generated code? At least in the api-gen i cannot find anything

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/checks/FunctionArgCountCheck.kt
chippmann: Yeah that would be good. But I'd do that in a separate PR

chippmann:
> Wouldn't that be a good opportunity to put the maximum number of argument as a constant in the new common tool module?

With the IDE plugin update you're right. So if ok for you. I'll extract it there once more modules will use this number

CedNaru (approved):

piiertho (approved):

PR #455, opened 2023-03-18
Gradle 8 compatibility

chippmann:
This makes the project compatible with gradle version 8 and increases the supported kotlin version to `1.8.10`

### DIFF
--- a/kt/plugins/godot-gradle-plugin/build.gradle.kts
+++ b/kt/plugins/godot-gradle-plugin/build.gradle.kts
@@ -44,6 +44,20 @@ tasks {
     }
 }
 
+publishing {
+    publications {
+        val godotGradlePlugin by creating(MavenPublication::class) {
+            pom {
+                name.set(project.name)
+                description.set("Godot gradle plugin for kotlin language support.")
+            }
+            artifactId = project.name
+            description = "Godot gradle plugin for kotlin language support."
+            from(components.getByName("java"))
+        }
+    }
+}
+
 project.extra["artifacts"] = arrayOf("godotGradlePlugin")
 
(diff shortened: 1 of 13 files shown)


kt/plugins/godot-gradle-plugin/build.gradle.kts
```diff
@@ -2,34 +2,25 @@ import godot.dependencies.gradle.DependenciesVersions
...
-    }
-    isAutomatedPublishing = false
-}
-
-pluginBundle {
-    website = "https://github.com/utopia-rise/godot-kotlin-jvm"
-    vcsUrl = "https://github.com/utopia-rise/godot-kotlin-jvm.git"
-    tags = listOf("kotlin", "godot", "gamedev")
 
-    mavenCoordinates {
```
piiertho, line 29: Where do we set maven coordinates now ?

chippmann: The gradle publishing plugin does that automatically now for the gradle plugin portal release as far as i understood. But no way for me to test that yet without a release. But i think it's working correctly because for the publishToMavenLocal task i get the warning that those two publications will override each other. If the coordinates would not match, they wouldn't do that.

For our manual maven central release we define that still through the regular publishing block below

chippmann: In any case there seems to be no way anymore to define that manually as the `isAutomatedPublishing` flag (which was required to configure the coordinates manually) cannot be set to false anymore

piiertho (approved):

CedNaru (approved):

PR #462, opened 2023-04-12
feat(4.0): implement signal api and cpp-jvm protocol for them

piiertho:
This implements signal api on JVM side.
This implements protocol between cpp and JVM for them.

This also fixes `test_reflection.gd`, which has been imported with cherry pick but not adapted to new script resource file for godot 4.0.


chippmann (approved):
LGTM
thanks :-)

PR #463, opened 2023-05-02
4.0/graal jni configs

piiertho:
This updates packed array api.  
This updates graal jni configuration .


CedNaru (approved):
Looks fine to me. 
Way more methods than expected, thanks for the time used on it.

PR #465, opened 2023-05-24
Feature/4.0/update core types

piiertho:
This updates all core types for godot 4.0.


kt/godot-library/src/main/kotlin/godot/core/AABB.kt
```diff
@@ -318,6 +324,20 @@ class AABB(
         }
     }
 
+    /**
+     * Returns true if the AABB is empty.
+     */
+    fun hasSurface(): Boolean {
```
CedNaru, line 330: Commentaire is incorrect, it returns false if empty

kt/godot-library/src/main/kotlin/godot/core/AABB.kt
```diff
@@ -318,6 +324,20 @@ class AABB(
...
+        return (_size.x > CMP_EPSILON && _size.y > CMP_EPSILON && _size.z > CMP_EPSILON)
+    }
+
+    /**
+     * Returns true if the AABB is flat or empty.
+     */
+    fun hasVolume(): Boolean {
```
CedNaru, line 337: Same here

kt/godot-library/src/main/kotlin/godot/core/AABB.kt
```diff
@@ -396,6 +416,49 @@ class AABB(
         return under && over
     }
 
+    /**
+     * Returns `true` if the given ray intersects with this [AABB]. Ray length is infinite.
+     */
+    fun intersectsRay(from: Vector3, dir: Vector3): Boolean {
+        require(size.x >= 0 && size.y >= 0 && size.z >= 0) {
+            "AABB size is negative, this is not supported. Use AABB.abs() to get an AABB with a positive size."
+        }
```
chippmann, line 425: Is it a good idea to crash here? Shouldn't we log an error and return false instead?
I know godot "crashes" in such cases but don't we then have issues if such a call originated from the cpp side?

Same for other usages of `require` and the likes.

kt/godot-library/src/main/kotlin/godot/core/Color.kt
```diff
@@ -535,6 +587,237 @@ class Color(
             val r = (hex and 0xFFFF) / 65535.0
             return Color(r, g, b, a)
         }
+
+        private val namedColors = arrayOf(
```
CedNaru, line 591: Should add more consistency here. Those names already exist as variables, meaning we create them two times in two different ways (one with constructor and one with hex). Should be merged so we only create them one time.

kt/godot-library/src/main/kotlin/godot/core/Color.kt
```diff
@@ -535,6 +587,237 @@ class Color(
...
+            if (clr.isEmpty()) {
+                return false
+            }
+            if (clr[0] == '#') {
+                clr = clr.substring(1)
+            }
+
+            // Check if the amount of hex digits is valid.
+
+            // Check if the amount of hex digits is valid.
```
chippmann, line 804: Duplicate comment

kt/godot-library/src/main/kotlin/godot/core/Color.kt
```diff
@@ -535,6 +587,237 @@ class Color(
...
+            val len = clr.length
+            if (!(len == 3 || len == 4 || len == 6 || len == 8)) {
+                return false
+            }
+
+            // Check if each hex digit is valid.
+
+            // Check if each hex digit is valid.
```
chippmann, line 812: again

kt/godot-library/src/main/kotlin/godot/core/Quaternion.kt
```diff
@@ -302,12 +286,132 @@ class Quaternion(
...
+        var postQ = postB
+
+        // Align flip phases.
+        fromQ = Basis(fromQ).getRotationQuaternion()
+        preQ = Basis(preQ).getRotationQuaternion()
+        toQ = Basis(toQ).getRotationQuaternion()
+        postQ = Basis(postQ).getRotationQuaternion()
+
+        // Flip quaternions to shortest path if necessary.
+        // Flip quaternions to shortest path if necessary.
```
chippmann, line 312: duplicate

kt/godot-library/src/main/kotlin/godot/core/Quaternion.kt
```diff
@@ -302,12 +286,132 @@ class Quaternion(
...
+        val flip1 = fromQ.dot(preQ).signbit
+        preQ = if (flip1) -preQ else preQ
+        val flip2 = fromQ.dot(toQ).signbit
+        toQ = if (flip2) -toQ else toQ
+        val flip3 = if (flip2) toQ.dot(postQ) <= 0 else toQ.dot(postQ).signbit
+        postQ = if (flip3) -postQ else postQ
+
+        // Calc by Expmap in from_q space.
+        // Calc by Expmap in from_q space.
```
chippmann, line 321: duplicate

kt/godot-library/src/main/kotlin/godot/core/Quaternion.kt
```diff
@@ -302,12 +286,132 @@ class Quaternion(
...
+        var post_q = p_post_b
+
+        // Align flip phases.
+        from_q = Basis(from_q).getRotationQuaternion()
+        pre_q = Basis(pre_q).getRotationQuaternion()
+        to_q = Basis(to_q).getRotationQuaternion()
+        post_q = Basis(post_q).getRotationQuaternion()
+
+        // Flip quaternions to shortest path if necessary.
+        // Flip quaternions to shortest path if necessary.
```
chippmann, line 382: duplicate

kt/godot-library/src/main/kotlin/godot/core/Quaternion.kt
```diff
@@ -302,12 +286,132 @@ class Quaternion(
...
+        val flip1 = from_q.dot(pre_q).signbit
+        pre_q = if (flip1) -pre_q else pre_q
+        val flip2 = from_q.dot(to_q).signbit
+        to_q = if (flip2) -to_q else to_q
+        val flip3 = if (flip2) to_q.dot(post_q) <= 0 else to_q.dot(post_q).signbit
+        post_q = if (flip3) -post_q else post_q
+
+        // Calc by Expmap in from_q space.
+        // Calc by Expmap in from_q space.
```
chippmann, line 391: duplicate

kt/godot-library/src/main/kotlin/godot/core/StringName.kt
```diff
@@ -36,7 +36,7 @@ class StringName : NativeCoreType {
         return TransferContext.readReturnValue(VariantType.STRING, false) as String
     }
 
-    //TODO/4.0: Implement
+    operator fun <T> invoke(stringOperation: String.() -> T): T = toString().stringOperation()
```
chippmann, line 39: Not sure what I'm looking at here. What would a typical use case of this invoke operator look like?

kt/godot-library/src/main/kotlin/godot/core/Vector2i.kt
```diff
@@ -101,6 +96,29 @@ class Vector2i(
...
+
+    /**
+     * Returns a new vector with each component snapped to the closest multiple of the corresponding component in [step].
+     */
+    fun snapped(step: Vector2i) = Vector2i(
+        snapped(x, step.x),
+        snapped(y, step.y)
+    )
+
+    fun mod(other: Vector2i) = Vector2i(
```
CedNaru, line 112: It seems that VectorX got plenty new operators including the < operator. Should we add the Comparable interface to them ?

kt/godot-library/src/main/kotlin/godot/util/Utils.kt
```diff
@@ -40,6 +43,30 @@ internal inline fun Float.toGodotReal(): Float = this
...
+val Long.signbit: Boolean
+    get() = sign < 0
+
+val Float.signbit: Boolean
+    get() = sign < 0
+
+val Double.signbit: Boolean
+    get() = sign < 0
+
+fun cubicInterpolate(
```
CedNaru, line 58: Shouldn't that go in the MathFuncs.kt you created in that same PR ?

CedNaru (changes requested):
Seems good overall, didn't expect that many changes to core types since the Alpha. But I got a few nitpicks and questions

chippmann (changes requested):
And just as a note: once we redo the tests, we should probably create a test case for each of the reimplemented functions of the core types to make sure they match the implementation of the "real" core types.

kt/godot-library/src/main/kotlin/godot/core/StringName.kt
CedNaru: StringName has a LOOOOOOT of functions that would be cumbersome to reimplement, even as Bridge, and they happen to be the same as the regular Godot String. So instead of spending time implementing that, we decide to just add a utility function that can transform a StringName using the regular Kotlin String methods.

chippmann: hmm i see. but then i would a dd a documentation on the function with a link to [this page](https://docs.godotengine.org/en/latest/classes/class_stringname.html) and a hint that one can reimplement those functions with this operator, and probably also provide an example of such a reimplementation using this operator

kt/godot-library/src/main/kotlin/godot/core/AABB.kt
piiertho: Former comment that I did not checked. Will change it.

kt/godot-library/src/main/kotlin/godot/core/AABB.kt
piiertho: I think we should better explain to jetbrains they have a vision regarding preprocessors that is not compatible with game industry.  
Here I would expect a debug version of engine to have this require call and crash, to spot problems asap, and a release version without this call.  
Anyway, this should not be modified in this PR IMO, as we have a lot of already existing require in core types.

CedNaru (approved):

PR #468, opened 2023-06-09
Fix `3.x` release pipeline

### LINKED ISSUE #467: Fix `3.x` release pipeline
Our release pipeline for the `3.x` branch is broken and produces broken binaries.

The reason for this lies here:
https://github.com/utopia-rise/godot-kotlin-jvm/blob/5d03ee134757e0cde8daec7dc06d6815632f0be3/.github/workflows/deploy-export-template.yaml#L1
and here:
https://github.com/utopia-rise/godot-kotlin-jvm/blob/5d03ee134757e0cde8daec7dc06d6815632f0be3/.github/workflows/deploy-godot-editor-release.yaml#L1

These two workflows will be triggered on the default branch instead of the `3.x` branch once the `Create Draft Release` workflow on the `3.x` branch has finished.

After some discussion on discord, we'll merge the affected workflows into the `Create Draft Release` workflow for simplicity fo fixing this issue as we'll hardly ever provide more `3.x` releases.

chippmann:
Fixes #467 

Also drops the update docs workflow and fixes the version regex (now the editor assets should have the correct release version in their name again).
This also fixes the name inconsistencies between the editor builds per platform. They have now all the same prefix.

For a test release see:
https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.6.1-3.5.2

Note:
The following things need to also go into master:
- name inconsistency fix for editor builds
- regex fix

For these I will submit a separate PR in the near future.


chippmann:
The failing check is because of the gpg key not being accessible for forks.

piiertho (approved):

PR #474, opened 2023-07-09
Fix 4.x release pipeline and improve export templates

chippmann:
This fixes the release pipeline for the 4.x releases.

It also improves the export templates by packing them into a `tpz` file (godot extension for export template archive. Basically this is just a zip with a renamed extension).
With this the user can just import the export templates through the godot editor by choosing "Install from file":
2023-07-09-095322_720x291_scrot
2023-07-09-095414_700x500_scrot

Sample release can be found here: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.7.0-4.0.0-SNAPSHOT

**Note:** I only tested the resulting linux editor and export template. The others still need testing


chippmann:
Failing check is because of gpg key

CedNaru (approved):

.github/workflows/deploy-godot-editor-release.yaml
```diff
@@ -229,12 +229,12 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: godotengine/godot
-          ref: 3.5.1-stable
+          ref: 4.0-stable
```
CedNaru, line 232: Shouldn't that be 4.1 ?

chippmann: This PR was created before we were on 4.1.
But now yes.
But i will not be able to fix that until next sunday.
So if any of you two has time, feel free to just change it :-)

CedNaru (approved):

.github/workflows/deploy-export-template.yaml
```diff
@@ -221,19 +221,14 @@ jobs:
...
           name: android-export-${{ matrix.target }}-binary-${{ matrix.arch }}
-          path: |
-            **/${{ matrix.output_folder }}
+          path: ${{ matrix.output_folder }}
 
       - name: Upload export template ${{ matrix.target }}
         if: matrix.platform != 'android' && matrix.platform != 'macos'
-        uses: actions/upload-release-asset@v1.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
piiertho, line 231: Why is this removed ?

piiertho: oh version changed

piiertho (approved):

PR #482, opened 2023-07-21
Fix api gen for static methods

### LINKED ISSUE #477: Some static methods are not generated
`DirAccess` has a static function `open`. It is missing in the generated companion object

chippmann:
Resolves #477 

Static methods are called by passing a nullpointer in the method bind for the instance pointer.

This also fixes a generation issue if the default value is an int with meta uint_32. Previously this was packed in a Int class. But an unsigned 32 bit integer does not fit into a signed 32 bit integer. Hence I changed it to be of type long. (I could not use UInt because of: https://youtrack.jetbrains.com/issue/KT-43525. As UInt is a value class, we can not add the annotation `@JVMOverloads` to methods using it.)


kt/godot-library/src/main/kotlin/godot/core/memory/TransferContext.kt
```diff
@@ -52,9 +52,9 @@ internal object TransferContext {
         return ret
     }
 
-    fun callMethod(ptr: VoidPtr, methodIndex: Int, expectedReturnType: VariantType) {
+    fun callMethod(ptr: VoidPtr?, methodIndex: Int, expectedReturnType: VariantType) {
```
CedNaru: I would instead make the called directly input 0 here instead of adding a nullable + test.

src/memory/transfer_context.cpp
```diff
@@ -121,7 +121,10 @@ void TransferContext::icall(JNIEnv* rawEnv, jobject instance, jlong j_ptr, jint
     JVM_CRASH_COND_MSG(args_size > VARIANT_ARG_MAX, vformat("Cannot have more than %s arguments for method call.", VARIANT_ARG_MAX));
 #endif
 
-    auto* ptr {reinterpret_cast<Object*>(static_cast<uintptr_t>(j_ptr))};
+    Object* ptr;
```
CedNaru, line 124: Is the check necessary ? j_ptr is only going to be null in the case of a static method, which ignore the point argument anyway.

CedNaru (changes requested):

CedNaru (dismissed):

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/GenerationService.kt
```diff
@@ -150,6 +150,11 @@ class GenerationService(
             constantsTypeReceiver.addProperty(generateConstant(constant, name))
         }
 
+        for (method in enrichedClass.methods.filter { it.internal.isStatic }) {
+            //TODO/4.0: Change implementation for static method
```
piiertho, line 154: Should this comment stay here ?

chippmann: TBH i don't know. I did not add it and i don't know why it was here ^^

piiertho: IIRC, it was here to remember to do this PR

CedNaru (approved):

PR #483, opened 2023-07-22
Feature/use kotlin preprocessors plugin

piiertho:
This makes use of new [kotlin-preprocessors](https://github.com/utopia-rise/kotlin-preprocessors) gradle plugin.  
This creates a `debug` and `release` version of `godot-library`, which is now published under `godot-library-debug` and `godot-library-release`. Add `-Prelease` when building `godot-library` to build `release` version, default is `debug`.  
Godot Kotlin gradle plugins now check presence of `release` parameter. If it is present, it will fetch `godot-library-release`, otherwise it will fetch `godot-library-debug`.  

This does not build when exporting, this should be done in another PR.

### DIFF
--- a/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
+++ b/.github/workflows/check-pr-engine-editor-debug-and-tests.yaml
@@ -14,10 +14,16 @@ jobs:
   build-godot-bootstrap:
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
+        name: [ Debug, Release ]
         include:
-          - os: ubuntu-latest
+          - name: Debug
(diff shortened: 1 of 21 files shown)


chippmann (approved):

chippmann:
Could we merge that after #474?

chippmann (approved):

CedNaru:
Some documentation should be written so users know how they are supposed to build with the release library.

piiertho:
> Some documentation should be written so users know how they are supposed to build with the release library.

Added a little documentation about compiling.

CedNaru (approved):

PR #484, opened 2023-07-22
Steamlining: Remove JClass and ClassLoader from JavaInstanceWrapper

CedNaru:


harness/tests/test/unit/test_inheritance.gd
```diff
@@ -40,3 +40,12 @@ func test_call_parent_open_var_from_child() -> void:
 	child_script.open_var = 101
 	assert_eq(child_script.open_var, 101, "Open var inherited from parent should now be 101")
 	child_script.free()
+	
+func test_script_is_child_of() -> void:
+	var child = ClassInheritanceChild.new()
+	var chid_script: Script = child.get_script()
```
chippmann, line 46: Typo

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/filebuilder/ClassRegistrarFileBuilder.kt
```diff
@@ -1,12 +1,6 @@
 package godot.entrygenerator.filebuilder
 
-import com.squareup.kotlinpoet.AnnotationSpec
-import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.FileSpec
-import com.squareup.kotlinpoet.FunSpec
-import com.squareup.kotlinpoet.KModifier
-import com.squareup.kotlinpoet.TypeSpec
-import com.squareup.kotlinpoet.asClassName
+import com.squareup.kotlinpoet.*
```
chippmann, line 3: Please do single name exports (can be configured in the ide and yes we should make that a project setting at some point)

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/filebuilder/ClassRegistrarFileBuilder.kt
```diff
@@ -56,11 +50,14 @@ class ClassRegistrarFileBuilder(
         .beginControlFlow("with(registry)") //START: with registry
         .let { funSpecBuilder ->
             if (!registeredClass.isAbstract) {
+                val superClasses = registeredClass.supertypes.takeWhile {
+                    //Used to implement script inheritance methods, so we remove base types and abstract parents.
+                    it.name != registeredClass.godotBaseClass && !it.isAbstract
```
chippmann, line 55: I would be explicit whenever possible and write `supertype` instead of `it`

kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/filebuilder/ClassRegistrarFileBuilder.kt
```diff
@@ -56,11 +50,14 @@ class ClassRegistrarFileBuilder(
...
+                }.map { "\"${it.name}\"" }.reduceOrNull{ acc, string -> "$acc,$string" }?:""
```
chippmann, line 56: `acc`? The time you "gained" by not typing the remaining characters is time i waste every time trying to guess what that name means 🤣

chippmann, line 56: I see two problems here:
- this will probably break when the user configures fqName registration as you only use the name
- when you use the fqName the imports are not automatically added. Use poet templates so the imports are correctly added (`%T`)

kt/godot-library/src/main/kotlin/godot/registration/Registration.kt
```diff
@@ -53,7 +53,7 @@ data class KtFunctionArgument(
 class ClassBuilderDsl<T : KtObject>(
     @PublishedApi internal val localResourcePath: String,
     private val registeredName: String,
-    private val superClass: String,
+    private val superClass: List<String>,
```
chippmann, line 56: I would rename the parameter to `superClasses`

chippmann (changes requested):
A very welcome (and much needed) improvement! Amazing work!

kt/godot-library/src/main/kotlin/godot/core/KtClass.kt
```diff
@@ -4,13 +4,15 @@ package godot.core
...
     val registeredName: String,
-    val superClass: String,
+    private val _superClasses: List<String>,
     private val _constructors: List<KtConstructor<T>?>,
     private val _properties: Map<String, KtProperty<T, *>>,
     private val _functions: Map<String, KtFunction<T, *>>,
     private val _signalInfos: Map<String, KtSignalInfo>,
     val baseGodotClass: String
 ) {
+    val superClasses: Array<String>
```
chippmann, line 14: NIT: maybe it would make sense to name this registeredSuperTypes?

chippmann (dismissed):

src/gd_kotlin.cpp
```diff
@@ -547,3 +547,20 @@ bool GDKotlin::initialized() const {
 const Vector<Pair<String, String>>& GDKotlin::get_configuration_errors() const {
     return configuration_errors;
 }
+
+void GDKotlin::initialize_classes() {
```
piiertho, line 551: I'm not fan of this one.
Can't we have a way to make this happens in the concerned classes (still in a static context) ?
With this design we have to modify this place each time we add a new class.

piiertho (changes requested):

src/gd_kotlin.cpp
CedNaru: We can't really do that. Because we have to wait for the JVM to be started anyway before initialising the classes and doing some lazy init inside the constructor would be going back to the design I removed with that PR.

piiertho: For posterity I sum up discord discussions:  
We don't have better solution to put this code to concerned classes.  
Local static init in constructor with `const char*` as template argument for name of `JavaInstanceWrapper` would have done the job, but it involves a lock and so would be unefficient for `KtObject`.

piiertho (approved):

PR #485, opened 2023-07-23
Unbind Singletons before JVM closes

CedNaru:
Depends on #484

Godot does free its singletons when it terminates. The issue is that our module is terminated (and so the JVM) before the singletons are freed, which causes an error when we try to free the remaining JNI references inside singletons.
This PR makes sure that singletons are unbound (without KtBinding) before the closing the JVM.


CedNaru:
Adding a fix to a small error I just spotted from the previous PR

chippmann (approved):

piiertho (approved):

chippmann (approved):

PR #487, opened 2023-08-11
Improve gradle setup and update ide plugin

### LINKED ISSUE #446: Update checks and refactoring tools in the IDE plugin
With the introduction of dummy files in https://github.com/utopia-rise/godot-kotlin-jvm/pull/441 we lifted some restrictions.
We need to remove the corresponding IDE checks and also update the refactoring tools to work with the new `gdj` files.

Furthermore we need to update the checks in general for godot 4 as some package paths of godot classes which we check have changed.

Also the signal gutter icons need to be updated.

chippmann:
Resolves #446

This improves our gradle setup by making use of `jvmToolchain`, gradle version catalog and convention plugins.
This should yield in a consistent dev environment and reduces the places where versions need to be managed.

This also updates the ide plugin to support godot 4. All checks should work and basic refactoring support with the new registration files is working as well (only source files -> registration files not the other way around).

Note that this does not include proper java support. I migrated all of the class checks to also work with java classes to make sure that the only thing needed for fully supporting java in the ide checks are the annotators and their helper functions. All other checks still need to be adapted for java support.
That said everything else is already supporting java (refactoring, class name indexing, signal connection indexing and as said all class checks and their quick fixes).

There are still a lot of improvements necessary but this PR is already huge enough and everything else can be improved later on.
The goal of this PR in regards to the ide plugin was mainly to restore feature parity with what we had before godot 4.


kt/plugins/godot-intellij-plugin/build.gradle.kts
```diff
@@ -1,68 +1,53 @@
...
+            "gradle"
+        ) // kotlin plugin version no longer needed as it's now bundled with the IDE
     )
 )
 
+// needed as the intellij plugin does add its own repositories and thus the ones defined in the settings.gradle.kts are ignored. Hence, we need to redefine them here
 repositories {
+    mavenLocal()
     mavenCentral()
-    maven {
```
piiertho, line 50: Why are those removed ?

chippmann: These are now automatically added by https://github.com/utopia-rise/godot-kotlin-jvm/blob/0a7c78e3b2bd64e4a99cc64d302bc09ce9605236/kt/plugins/godot-intellij-plugin/build.gradle.kts#L14
Also the reason of: https://github.com/utopia-rise/godot-kotlin-jvm/blob/0a7c78e3b2bd64e4a99cc64d302bc09ce9605236/kt/plugins/godot-intellij-plugin/build.gradle.kts#L38

piiertho (approved):

CedNaru (approved):

Issue #491, opened 2023-08-27
ICall crash with 5 method args limit


chippmann:
We have a method args limit of 5 which we enforce with:
```cpp
#ifdef DEBUG_ENABLED
    JVM_CRASH_COND_MSG(args_size > VARIANT_ARG_MAX, vformat("Cannot have more than %s arguments for method call.", VARIANT_ARG_MAX));
#endif
```

This leads to api call to fail:
```
ERROR: Godot-JVM: Cannot have more than 5 arguments for method call but tried to call method "ENetMultiplayerPeer::create_client" with 6 args
```

chippmann:
It seems godot changed the max to 8 a while ago: https://github.com/godotengine/godot/pull/54188

PR #493, opened 2023-08-27
Fix max arg size

chippmann:
Fixes:  #491

~~**Note:** Depends on changes from #487 hence the base branch is not `master`. Once #487 is merged, I'll change the base branch (but it should change to master automatically anyways)~~


piiertho (approved):

chippmann:
Should we also bump the max allowed arg count of the constructors to 8?

piiertho:
> Should we also bump the max allowed arg count of the constructors to 8?

I think we should

CedNaru (changes requested):
To increase the number of arguments in constructor, we have to:
-  Increase the value of the contraint constant.
-  Add the new generic functions to the list of [KTConstructors](https://github.com/utopia-rise/godot-kotlin-jvm/blob/6540d73e8263f4aab9392243456990fb86016f93/kt/godot-library/src/main/kotlin/godot/core/Constructors.kt#L106)
-  We also have to add the new KTFunctions to the [registration file](https://github.com/utopia-rise/godot-kotlin-jvm/blob/638c3d4767f057389d736384c521105e6d2005a3/kt/godot-library/src/main/kotlin/godot/registration/Registration.kt#L377)

CedNaru (dismissed):
To the moon

piiertho (approved):

CedNaru (approved):

Issue #509, opened 2023-10-12
Moving buffer.rewind() at the beginning of calls instead of end.


CedNaru:
It's something that has been bothering me for a while, debugging the project and going back and forth between the Kotlin and C++. When, for some reason, the data in the shared buffer are not valid, the module is going to throw an error. The error is often things like the number of arguments not matching, or the type of the variable not being the expected one, etc....
The issue with that approach is that those errors are often thrown before `rewind()` is called, leaving the buffer cursor is an invalid position instead of going back to 0. 
The consequence of that is that once you have got a first sharedbuffer error, it's going to cascade into all the next calls throwing errors as well because the cursor was never reset and so the data is incorrectly read. It makes the logs full of irrelevant messages, making it harder to decipher. 

I propose to change all the TransferContext methods so `rewind()` is now called at the beginning of each call. It shouldn't alter the regular behavior of the ShareBuffer and would also allow the module to "recover" after an error. The only difference would be that the Sharedbuffer is always left at a non-zero position after being used. But I don't think there should be any issue, as the next method using the buffer should reset it anyway.

It's a simple change, only meant to make the module a bit easier to debug, by removing noises after a first error.

chippmann:
A much needed improvement indeed and a low hanging fruit. I see no issues with this approach.

piiertho:
Definitely a good idea. This will help a lot debugging.

PR #512, opened 2023-10-14
Rewind() called early in TransferContext

CedNaru:
Implement #509 

C++:
The idea is simple. Instead of having to call `rewind()` after any operation on the buffer, we simply rewind it as part of the `get_buffer()` method. (The buffer is stored directly inside that method, so it's the only way to get it).
The only manual `rewind()` left is the one in `icall()` as I need to reset it to write the return value.

Kotlin:
I decided to made `TransferContext.buffer` private (doesn't change its visibility from C++ through JNI). I don't think it should be used directly outside TransferContext, so I did a small rework of its methods.
The goal was to remove any manual call to `rewind()` in the rest of the code. Rewinding should be the responsibility of TransferContext alone.

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/core/memory/TransferContext.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/memory/TransferContext.kt
@@ -2,6 +2,7 @@ package godot.core.memory
 
 import godot.core.KtObject
 import godot.core.LongStringQueue
+import godot.core.ObjectID
 import godot.core.VariantType
 import godot.tools.common.constants.Constraints
 import godot.util.VoidPtr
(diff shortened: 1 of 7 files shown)


src/memory/transfer_context.cpp
```diff
@@ -43,7 +43,7 @@ SharedBuffer* TransferContext::get_buffer(jni::Env& p_env) {
         shared_buffer = SharedBuffer {address, 0};
 #endif
     }
-
+    shared_buffer.rewind();
```
chippmann, line 46: This part i find a bit dangerous. If one calls `get` it has the side effect of rewinding the buffer. I do agree that this is the correct way of doing it, but i would rename the method from `get` to `get_and_rewind` or `use` ore something like that.

chippmann (changes requested):

src/memory/transfer_context.cpp
CedNaru: I can do that yes, not a big deal.
I just don't think it dangerous to use as it is, `get_buffer` is a private method so we are not exposing a dangerous method anyway.

chippmann: True

chippmann (approved):

kt/godot-library/src/main/kotlin/godot/core/memory/TransferContext.kt
```diff
@@ -22,31 +23,51 @@ internal object TransferContext {
...
     fun writeReturnValue(value: Any?, type: VariantType) {
-        type.toGodot(buffer, value)
         buffer.rewind()
+        type.toGodot(buffer, value)
     }
 
     fun readReturnValue(type: VariantType, isNullable: Boolean = false): Any? {
-        val ret = type.toKotlin(buffer, isNullable)
         buffer.rewind()
+        val ret = type.toKotlin(buffer, isNullable)
```
piiertho, line 70: We can remove variable here and directly return

src/memory/transfer_context.h
```diff
@@ -46,13 +46,10 @@ class TransferContext : public JavaInstanceWrapper {
         for (uint32_t i = 0; i < args_size; ++i) {
             ktvariant::get_variant_from_buffer(buffer, p_args[i]);
         }
-
-        buffer->rewind();
     }
 
     _FORCE_INLINE_ static void write_return_value(SharedBuffer* buffer, const Variant& r_ret) {
         ktvariant::send_variant_to_buffer(r_ret, buffer);
-        buffer->rewind();
```
piiertho, line 55: Why this method does not rewind ?
This would avoid to rewind before calling it in icall method.

CedNaru: It's simply to keep all the rewind() calls in a single files, so it's easier to follow. I'd rather just erasing this function than putting rewind back in it.

piiertho: I'm ok with deleting it. This way all is in cpp and is still consistent

piiertho (approved):

PR #515, opened 2023-10-15
Increase MAX_FUNCTION_ARG_COUNT to 16.

CedNaru:
Some API methods couldn't be called because they exceeded the maximal amount of arguments we support.
I added a sanity check in the api generator to detect when such cases happen.
The theorical maximal number of arguments in Godot in unlimited since version 4.
The practical maximal number to date in 4.1.2 is 14 arguments.
I increased the limit to 16, to keep it a power of 2 and give us a bit of leeway.


chippmann (approved):
We should really think about generating those functions at some point

piiertho (changes requested):
This misses change of constant in constraints.h

piiertho:
> This misses change of constant in constraints.h

Ah never mind, did not see it

piiertho (approved):

PR #518, opened 2023-10-17
Update BunnyMark to Godot Kotlin 4

CedNaru:
- Kotlin scripts updated for Godot 4
- Gdscripts updated for Godot 4 and now typed to get the best performance out of them.
- I didn't test C#. I have never used that language and don't want to spend 4h trying to set it on my computer. If someone is willing they can do it but I don't think it's much relevant for us anyway.
- Also updated the benchmark script, so it can reach stability faster.

Kotlin JVM is incredibly faster than before, compared to GDScript (even with the extra performance from types) Our worst case is still the BunnymarkV3. That one is supposed to stress JNI calls. Despite this, we are still holding decently well.
Native images are a disappointment, I expected less performance overall because of the absence of JIT, but it seems that even BunnymarkV3 is way slower too. It doesn't seem native compilation is helping with JNI calls.

### DIFF
--- a/harness/bunnymark/Benchmarker.gd
+++ b/harness/bunnymark/Benchmarker.gd
@@ -1,25 +1,29 @@
 extends Control
 
-var fps_update_interval = 1.0
-var elapsed_time = 0.0
-var fps_label = null
-var benchmark_container = null
-var benchmark_node = null
-var output_path = "user://benchmark_results.json"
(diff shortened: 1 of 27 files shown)


chippmann (approved):

piiertho (approved):
One comment. We should do it in another PR.

PR #522, opened 2023-10-22
Add unit tests and fixes for core types.

CedNaru:
Unit tests are straight conversions from the C++ ones to Kotlin. I only made changes when it was necessary for syntax reasons and removed tests checking methods not part of the user API.

Also took the opportunity to move a few files in the core package, so it's less bloated.

There is now a "math" directory for all the reimplemented core types and a "bridge" for the native ones (replicating the way it's done in C++).

The unit tests revealed a lot of issues in the math core types, so I fixed everything I could find, mostly converting the  updated C++ code to Kotlin. The issues were a mix of errors from when we did with the initial implementation, as well as mathematical mistakes in the original Godot source code, fixed since Godot 4.


kt/godot-library/src/test/kotlin/godot/core/GodotTest.kt
```diff
@@ -0,0 +1,22 @@
...
+    assertFalse(value)
+}
+
+inline fun checkMessage(value: Boolean, lazyMessage: () -> String) {
+    assertTrue(lazyMessage(), value)
+}
+
+inline fun checkFalseMessage(value: Boolean, lazyMessage: () -> String) {
+    assertFalse(lazyMessage(), value)
+}
```
chippmann, line 22: Not sure if I'm fond of these abstractions. For sure IMO assert is the better keyword. What is the reason for these abstractions?

kt/godot-library/build.gradle.kts
```diff
@@ -38,8 +38,10 @@ dependencies {
     // we need to add it here so reflection is available where the code is loaded (Bootstrap.kt) otherwise it will not work
     api(kotlin("reflect", version = libs.versions.kotlin.get()))
     implementation("com.utopia-rise:tools-common:$fullGodotKotlinJvmVersion")
+    testImplementation("junit", "junit", "4.12")
```
chippmann, line 41: Why junit 4 and not 5?

chippmann (changes requested):
Nice work! I'd also add a little readme somewhere where you note the commit hash and a date or something like this when you last synced the core type tests with the godot core type tests. With that, someone other than you has an easier time to check what has changed in those tests in the meantime.

kt/godot-library/src/test/kotlin/godot/core/GodotTest.kt
CedNaru: Because I wanted it to be easy to change whenever we switch to Junit 5

kt/godot-library/build.gradle.kts
CedNaru: Because I don't know how to use Junit 5 :p

kt/godot-library/src/main/kotlin/godot/core/math/AABB.kt
```diff
@@ -224,17 +225,17 @@ class AABB(
     /**
      *  Returns the scalar length of the longest axis of the AABB.
```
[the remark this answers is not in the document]
chippmann: as discussed on discord: this will be done in a separate PR.

kt/godot-library/src/main/kotlin/godot/core/math/AABB.kt
```diff
@@ -243,20 +244,20 @@ class AABB(
     /**
      * Gets the position of the 8 endpoints of the AABB in space.
```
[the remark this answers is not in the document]
chippmann: as discussed on discord: this will be done in a separate PR.

kt/godot-library/src/test/kotlin/godot/core/GodotTest.kt
chippmann: ok for me

kt/godot-library/build.gradle.kts
chippmann: As discussed on discord. Fine for now.

chippmann (approved):

piiertho (approved):

Issue #497, opened 2023-09-25
Allow Godot enums to work as bitfields.


CedNaru:
In the godot Json API, every enum has a `is_bitfield` property.

If it's false, it means that the enum should be used as a unique value:
`node.process_node = PROCESS_MODE_INHERIT `

If it's true, each enum represent one of several bits to be set in a mask.
`object.connect(callable, CONNECT_DEFERRED or CONNECT_PERSIST)`

Right now, we can't do like the second case directly. 
Generated enum don't have any bitwise operator, so you would need to use their "value" directly: 
`CONNECT_DEFERRED.id or CONNECT_PERSIST.id` 
For convenience right now, we discarded the use of the enum type if it's set as a bitfield, so we can directly use a long.
What we generate:
```kotlin
  public fun save(
    resource: Resource,
    path: String = "",
    flags: Long = 0,
  ): GodotError
```
What we have in the json:
```json
					"name": "save",
					"is_const": false,
					"is_vararg": false,
					"is_static": false,
					"is_virtual": false,
					"hash": 2303056517,
					"return_value": {
						"type": "enum::Error"
					},
					"arguments": [
						{
							"name": "resource",
							"type": "Resource"
						},
						{
							"name": "path",
							"type": "String",
							"default_value": "\"\""
						},
						{
							"name": "flags",
							"type": "bitfield::ResourceSaver.SaverFlags",
							"default_value": "0"
						}
					]
```

I suggest we change the api gen to get the enum type back as a parameter and add basic boolean operation on them.
Of course the result of using a OR operation on 2 enums won't be another existing enum, so the solution would be to change from enum to sealed class. It would have the following layout (simplified for sake of example, using ResourceSaver.SaverFlags as example):

```kotlin
sealed class SaverFlags(val id: Int){
    object FLAG_NONE: SaverFlags(0)
    object FLAG_RELATIVE_PATHS: SaverFlags(1)
    object FLAG_BUNDLE_RESOURCES: SaverFlags(2)
    object FLAG_CHANGE_PATH: SaverFlags(4)
    class CustomFlag private constructor (id: Int): SaverFlags(id)
    
    fun or(other: SaverFlags) = custom(id or other.id)
}
```
All original enum values are set as singleton object, but a custom regular class is added to allow the storing of other values.
That way we can write:
```kotlin
[...]

PR #524, opened 2023-10-23
feat(api-gen): Generate bitfields using interface, value class and statics

piiertho:
This resolves https://github.com/utopia-rise/godot-kotlin-jvm/issues/497


piiertho:
> Do we really need those XTrait interfaces ? It seems everything could be directly moved inside the sealed classes (which are just abstract classes already) and would bloat the generated code less.
> 
> Something like this, taking KeyModifiedMask as example:
> 
> ```kotlin
> public sealed class KeyModifierMask(
>     public val flag: Int,
> ){
>     public data object KEY_CODE_MASK : KeyModifierMask(8388607)
>     public data object KEY_MODIFIER_MASK : KeyModifierMask(532676608)
>     public data object KEY_MASK_CMD_OR_CTRL : KeyModifierMask(16777216)
>     public data object KEY_MASK_SHIFT : KeyModifierMask(33554432)
>     public data object KEY_MASK_ALT : KeyModifierMask(67108864)
>     public data object KEY_MASK_META : KeyModifierMask(134217728)
>     public data object KEY_MASK_CTRL : KeyModifierMask(268435456)
>     public data object KEY_MASK_KPAD : KeyModifierMask(536870912)
>     public data object KEY_MASK_GROUP_SWITCH : KeyModifierMask(1073741824)
>     public class CustomFlag internal constructor(flag: Int) : KeyModifierMask(flag)
> 
>     public infix fun or(other: KeyModifierMask): KeyModifierMask = CustomFlag(flag or other.flag)
>     public infix fun or(other: Int): KeyModifierMask = CustomFlag(flag or other)
>     public infix fun xor(other: KeyModifierMask): KeyModifierMask = CustomFlag(flag xor other.flag)
>     public infix fun xor(other: Int): KeyModifierMask = CustomFlag(flag xor other)
>     public infix fun and(other: KeyModifierMask): KeyModifierMask = CustomFlag(flag and other.flag)
>     public infix fun and(other: Int): KeyModifierMask = CustomFlag(flag and other)
> }
> ```
> 
> We lose the value class for Custom flag with this implementation but the "value" was already useless anyway. The moment you send it as a `KeyModifierMask` parameter and not a `CustomFlag` parameter, it's going to be autoboxed. Any polymorphic context removes the value and transforms it into a regular object, and there is no case where it's not going to happen in the context of Godot bitfields.
> 
> Also careful about the use of either Long or Int to carry the value. Right now the Int values are sent alongside a VariantType.Long when written to the buffer. Regular enums are using Long to store their values if I remember. I don't know if Godot ever go above the limits of 32 bits for enums. Depending on that, we could just move everything to Int. But regardless, we need consistency on that front.

After testing with bytecode viewer it seems it's sometime inline, sometime not.  
Interface has the counterpart to enable user implem, which we don't want.  
So we'll lose value class but not a big deal.

CedNaru:
The issue with the way it's implemented is that each enum value is a kotlin singleton, which ends up as its own class. The solution would be to discard the use of sealed class containing objects and use a sealed interface instead:

```kotlin
public sealed interface KeyModifierMask {
    public val flag: Int
    
   @JvmInline
   value class CustomFlag internal constructor(override val flag: Int) : KeyModifierMask 
}

enum class KeyModifierMaskConstants(override val flag: Int) : KeyModifierMask {
    KEY_CODE_MASK(8388607),
    KEY_MODIFIER_MASK(532676608),
    KEY_MASK_CMD_OR_CTRL(16777216),
    KEY_MASK_SHIFT(33554432),
    KEY_MASK_ALT(67108864),
    KEY_MASK_META(134217728),
    KEY_MASK_CTRL(268435456),
    KEY_MASK_KPAD(536870912),
    KEY_MASK_GROUP_SWITCH(1073741824);
}
```

With this we are back to enums with the exception of a custom implementation in the parent sealed interface. But we still introduce 2 different names. All enum users will manipulate are in the "Contant" enum, but api methods will require an instance of the parent interface, which can lead to confusions. 
If we want to go even further, we could entirely get rid of the enum and only use values inside companion object for the sealed interface.

```kotlin

public sealed interface KeyModifierMask {
    public val flag: Int

    @JvmInline
    value class BitFlagValue internal constructor(override val flag: Int) : KeyModifierMask

    companion object {
        val KEY_CODE_MASK: KeyModifierMask = BitFlagValue(8388607)
        val KEY_MODIFIER_MASK: KeyModifierMask = BitFlagValue(532676608)
        val KEY_MASK_CMD_OR_CTRL: KeyModifierMask = BitFlagValue(16777216)
        val KEY_MASK_SHIFT: KeyModifierMask = BitFlagValue(33554432)
        val KEY_MASK_ALT: KeyModifierMask = BitFlagValue(67108864)
        val KEY_MASK_META: KeyModifierMask = BitFlagValue(134217728)
        val KEY_MASK_CTRL: KeyModifierMask = BitFlagValue(268435456)
        val KEY_MASK_KPAD: KeyModifierMask = BitFlagValue(536870912)
        val KEY_MASK_GROUP_SWITCH: KeyModifierMask = BitFlagValue(1073741824)
    }
}
```

CedNaru (approved):

PR #531, opened 2023-11-01
Fix comilation of unit tests in modules with out gradle plugin

### LINKED ISSUE #519: Can't build the project if there is a test source set
The entry generation happens on the test source set as well, not just the main one. The issue is that the test source set also does not have the godot library, so if you try to make tests then you can't compile it anymore.

image
image

The easy fix is to add the library to the implementation config so it transitions to the test source set as well.
```kotlin
dependencies {
// https://mvnrepository.com/artifact/com.utopia-rise/godot-library
    implementation("com.utopia-rise:godot-library:0.7.2-4.1.2")
}
```

It does not really matter that the generator processes that too, only that you can't compile it anymore.

chippmann:
Fixes #519

Defines our godot library as implementation so it can be used by test source sets as well.
Wile the library get's added to the main jar as well, there are no conflicts with the one added by bootstrap jar.


chippmann:
Closing as this will get resolved once we switch to ksp 2.0.0

PR #535, opened 2023-11-08
Implement an LRU cache for StringName objects.

apkelly:
A test of my forked repo being able to send PRs to main repo.

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/core/bridge/StringName.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/bridge/StringName.kt
@@ -64,4 +64,6 @@ class StringName : NativeCoreType {
     }
 }
 
-fun String.asStringName() = StringName(this)
+fun String.asStringName(): StringName {
+    return MemoryManager.getOrCreateStringName(this)
+}
--- a/kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
@@ -3,6 +3,7 @@ package godot.core.memory
 import godot.core.KtObject
 import godot.core.NativeCoreType
 import godot.core.ObjectID
+import godot.core.StringName
 import godot.core.VariantType
 import godot.util.VoidPtr
 import godot.util.info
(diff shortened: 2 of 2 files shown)


kt/godot-library/src/main/kotlin/godot/core/bridge/StringName.kt
```diff
@@ -64,4 +64,6 @@ class StringName : NativeCoreType {
     }
 }
 
-fun String.asStringName() = StringName(this)
+fun String.asStringName(): StringName {
```
chippmann, line 67: As mentioned on discord, we'd need this for nodepath as well

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
```diff
@@ -67,6 +68,24 @@ internal object MemoryManager {
...
 
+    // A basic LRU cache.
+    private class LRUCache<K, V>(private val capacity: Int) : LinkedHashMap<K, V>(capacity, 0.75f, true) {
+        override fun removeEldestEntry(eldest: MutableMap.MutableEntry<K, V>?): Boolean {
+            return size > capacity
+        }
+    }
+
+    // Create an LRU cache for StringName objects based on a String key.
+    private val stringNameCache = LRUCache<String, StringName>(100)
```
chippmann, line 79: the size needs to be definable through commandline args. but as this requires cpp changes and jni calls it's fine for me if we do this in another pr if you don't want to do this.
Just tell me and I'll create a separate issue for this.

chippmann (changes requested):
Some unit test would be nice as well. But here again; Personally I'm fine with adding them later as well. I don't want to force work onto volunteers.

That said; thanks a lot for your first contribution!

kt/godot-library/src/main/kotlin/godot/core/bridge/StringName.kt
apkelly: I've added NodePath and moved the extension function for StringName into StringExt.

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
apkelly: I've added a TODO comment for now, let me take a look at the example you referenced and see if I can copy the implementation.

piiertho (approved):

PR #567, opened 2024-01-25
fix: Kill existing user threads except main when closing engine

### LINKED ISSUE #566: Game does not exit when non-daemon thread is running in the JVM
If we create a regular thread in Kotlin which keeps running, like so:

```kotlin
@RegisterClass
class MyButton : Button() {

    @RegisterFunction
    override fun _pressed() {
        thread {
            GD.print("Sleeping")
            Thread.sleep(TimeUnit.HOURS.toMillis(1))
            GD.print("Wake up!")
        }

    }
}
```

... then the game window will not exit properly when clicking on the window close ("x") button. Hitting the "stop" button in the editor still works.

The workaround is to declare the thread as a `daemon` thread:

```kotlin
    thread(daemon = true){
        // ...
    }
```

... as this will not prevent the JVM from exiting.

I don't think that this needs to have super high priority, but if a library would spawn threads (e.g. an async HTTP client might do that) and doesn't permit control over the creation of the thread, then this can become an issue.

Maybe this can be handled on the C++ side? It works when stopping the game via the godot editor "stop" button after all.

piiertho:
This resolves #566

### DIFF
--- a/harness/tests/src/main/kotlin/godot/tests/Invocation.kt
+++ b/harness/tests/src/main/kotlin/godot/tests/Invocation.kt
@@ -41,11 +41,13 @@ import godot.core.asStringName
 import godot.core.dictionaryOf
 import godot.core.variantArrayOf
 import godot.extensions.getNodeAs
+import godot.global.GD
 import godot.registration.Range
 import godot.signals.signal
 import godot.tests.subpackage.OtherScript
 import godot.util.RealT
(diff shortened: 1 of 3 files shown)


CedNaru (approved):

chippmann (approved):

CedNaru:
I think we can close that PR now. That issue is going to be addressed by a new feature added in this #661

PR #572, opened 2024-02-07
Fix reflection usage in exported projects

### LINKED ISSUE #571: Kotlin reflection not always applied to bootstrap jar in exported projects
In editor everything works as expected. The bootstrap jar we ship with the editor build does include kotlin-reflect (`full` and `jvm`) due to: #452 .

As a reminder for the following; currently the exported bootstrap jar is not the same as the one we ship with the editor. The one that get's actually exported is built together with the users project. Whether this still makes sense or not is not part of this issue and should be looked at individually later on in a follow up issue!

In simple reflection use cases, exported projects do work.  The exported bootstrap jar does contain the necessary reflect classes.
If one however defines the following jackson dependency in the project (i strongly assume there are many other such dependencies out there) `implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.16.1")`, the reflect classes are suddenly missing from the exported bootstrap jar and upon reflection usage, a crash occurs stating that `kotlin-reflect` is missing.

If a user manually replaces the built bootstrap jar in the `build` directory before exporting the game, the code works as the reflection classes are present. Some other classes are stripped as well but these are expected (such as tool classes only used for editor builds).

chippmann:
Fixes #571 

Note: this is intended as a quick fix for #571 so we can publish a hotfix.
We still need to further investigate if a separate bootstrap jar for exported projects is still a good idea and still needed or if we should just export the bootstrap jar we ship with the editor.

@piiertho the graalvm config for the newly added tests are missing here. As i cannot remember how to check which configs are missing, could you give me a quick heads up again on discord?

This was tested on linux x64 on editor builds, debug and release exports.


chippmann:
@MartinHaeusler I'd appreciate if you could test this with your project as well if possible and you have the time.

MartinHaeusler:
@chippmann I've just tested it locally; I checked out this PR from github, built the godot plugin, published to maven local, loaded it from there in my project, and did a clean build of my project. The `godot-bootstrap.jar` which was created this way now contains the full `kotlin/reflect/jvm` and `kotlin/reflect/full` packages. I then used the godot editor to export the project and started the standalone executable; everything worked like a charm!

Thank you very much for your efforts :)

CedNaru (approved):

PR #574, opened 2024-02-08
[Bugfix] Fix typed array registration

### LINKED ISSUE #573: Exported VariantArrays are missing type information in Godot Editor
Exporting an array in GDScript like this:

```gdscript
@export var nodes: Array[Node3D]
```

... produces a **typed** array editor like this:

image

Exporting the corresponding array property in Kotlin:

```kotlin
@Export
@RegisterProperty
var nodes: VariantArray<Node3D>? = null
```

... produces an **untyped** array editor like this:

image

Note that in the exported Kotlin property, the editor just says "Array (Size 2)" in the title, whereas in the GDScript case, it says "Array[Node 3D] (size 2)".

This issue is not merely cosmetical; in the editor resulting from kotlin code, the user can assign **any** resource type in the Godot editor, which will fail immediately with a type error at runtime.

@chippmann assumes that it has something to do with the way Godot 4 exports type hints for arrays.

chippmann:
Fixes #573 

With godot 4 the hint string for the inspector changed it's format and now also Node and Resource based classes are allowed for Array's.
Our implementation now covers those cases as well.

For reference, the code which handles this in the inspector can be found here: https://github.com/godotengine/godot/blob/36e943b6b20cb7a8a89bc30489c4a81c3e149d74/editor/editor_properties_array_dict.cpp#L616-L635

And the GDScript parser handles this here: https://github.com/godotengine/godot/blob/36e943b6b20cb7a8a89bc30489c4a81c3e149d74/modules/gdscript/gdscript_parser.cpp#L4077-L4170

Custom classes are also supported now:


piiertho (approved):

piiertho (approved):

harness/tests/scripts/ScriptInOtherSourceDir.gdj
```diff
@@ -6,17 +6,17 @@ fqName = ScriptInOtherSourceDir
 relativeSourcePath = otherSourceDir/ScriptInOtherSourceDir.kt
 baseType = Node
 supertypes = [
-    godot.Node,
+	godot.Node,
```
CedNaru, line 9: Lot of .gdj files with indentation suddenly changed to tab instead of space.

kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/ext/ksFunctionDeclarationExt.kt
```diff
@@ -8,14 +8,19 @@ import godot.entrygenerator.model.ConstructorAnnotation
 import godot.entrygenerator.model.FunctionAnnotation
 import godot.entrygenerator.model.RegisteredConstructor
 import godot.entrygenerator.model.RegisteredFunction
+import java.io.File
 
-internal fun KSFunctionDeclaration.mapToRegisteredConstructor(): RegisteredConstructor {
+internal fun KSFunctionDeclaration.mapToRegisteredConstructor(
+    isFqNameRegistrationEnabled: Boolean,
```
CedNaru, line 14: You are passing the same 3 arguments around everywhere, I would gather them in a single data class and pass a single parameter to make things less bloated.

harness/tests/scripts/ScriptInOtherSourceDir.gdj
chippmann: Just the first line. They were commited wrongly and i just committed then correctly.
Either the godot editor or intellij breaks the formatting of the first line. I need to investigate this at some point but IMO this is very low priority

kt/entry-generation/godot-kotlin-symbol-processor/src/main/kotlin/godot/annotation/processor/ext/ksFunctionDeclarationExt.kt
chippmann: Agreed. Will change that.

CedNaru (approved):

PR #575, opened 2024-02-08
Show system alert dialog if system is misconfigured

### LINKED ISSUE #562: Game Window closes itself without any error message when GDJ files are outdated
I've experienced the following several times:
- change some kotlin files which have `@RegisterClass` on them
- forget to rebuild gradle
- switch to godot editor
- launch the game
- the game window appears...
- ... and closes itself immediately again

The issue here is that we get **no error message** or any indication of what went wrong. Perhaps we could improve upon this by showing some logs?

chippmann:
This shows system alert dialogs if the system or project is misconfigured (JAVA_HOME missing, no embedded jvm on exported projects).

**NOTE:** This only handles cases where we would otherwise quit the application directly or would not initialize our module. This also adds a hard exit of the editor where we otherwise just did not initialize our module. But currently our module misbehaves if there are gdj files present but our module is not properly initialized. Until this is fixed, we just exit the editor after displaying the error message.

This should help in detecting issues without the need of looking into the editor (or game) logs
2024-02-08-151450_1029x641_scrot

This now also creates a alert dialog upon fatal errors which will close the program and thus should resolve #562 :
2024-02-08-154400_895x426_scrot


src/logging.h
```diff
@@ -20,8 +21,14 @@
 #define JVM_ERR_FAIL_COND_V_MSG(condition, ret_var, message) \
     ERR_FAIL_COND_V_MSG(condition, ret_var, vformat("Godot-JVM: %s", message))
 
-#define JVM_CRASH_COND_MSG(condition, message) CRASH_COND_MSG(condition, vformat("Godot-JVM: %s", message))
-
-#define JVM_CRASH_NOW_MSG(message) CRASH_NOW_MSG(vformat("Godot-JVM: %s", message))
+#define JVM_CRASH_COND_MSG(condition, message) \
+    if (unlikely(condition)) { OS::get_singleton()->alert(message, "Fatal error"); } \
```
piiertho, line 25: Not a big fan of adding another if for crash cond.  
In this case this will crash, so crash message should be enough.  
Or this if can be debug only.

src/logging.h
```diff
@@ -20,8 +21,14 @@
...
+    CRASH_COND_MSG(condition, vformat("Godot-JVM: %s", message))
 
+#define JVM_CRASH_NOW_MSG(message) \
+    OS::get_singleton()->alert(                \
```
piiertho, line 29: Same: Debug only. Otherwise we add behaviour to release game.

piiertho (changes requested):

src/logging.h
chippmann: the issue is that the crash message is only logged to the terminal. If you did not start the game from the terminal you'll never see it. But i agree, a ifdef for debug only would probably be best

piiertho (approved):

CedNaru (approved):

PR #580, opened 2024-02-18
Fix android export templates

chippmann:
This fixes our android export templates.

What's actually the issue:
Our normal android export templates are fine and they work. 

But in godot, you can enable gradle build for an android export. For simple games one usually does not need this. But if one needs to edit the android manifest or otherwise needs to alter how the android build is built, then this is needed. If enabled, godot pulls a whole android gradle project out of `android_sources.zip` into the `res://` folder and uses this to build the actual android exports rather than the export templates.

Our problem was that, because of this line: https://github.com/utopia-rise/godot-kotlin-jvm/blob/1dfa11d5c43fc9f85de2260b8e88f4911afddcb9/.github/workflows/deploy-export-template.yaml#L298, we only have the needed native libraries for release builds. Hence in this mode, debug builds are broken as the `android_sources.zip` does not contain the needed binaries for the debug target (namely `libs/debug/godot-lib.template_debug.aar`)

This means we no longer can build debug and release separately, they need to be built in one go, or at least the `android_sources.zip` needs to be built for both at once and thus needs access to debug and release binaries.

## This is just a hotfix!
We really need to rework our CI/CD pipeline and this issue shows that we should do this sooner than later. Hence I plan to tackle this next. But this will hardly be finished before the next release so this is a dirty hotfix to fix this issue in time.

Once we build per platform rather than per target, and thus align ourselves more with how godot builds their binaries, this issue should basically "solve" itself.

### DIFF
--- a/.github/workflows/deploy-export-template.yaml
+++ b/.github/workflows/deploy-export-template.yaml
@@ -240,23 +240,7 @@ jobs:
 
   build-android-export-template:
     needs: [deploy-export-template]
-    strategy:
-      matrix:
-        name: [ Android-debug, Android-release ]
-        include:
-          - name: Android-debug
-            os: ubuntu-20.04
-            java-version: 11
-            target: template_debug
-            base-path: platform/android/java/lib/libs/debug
-            apk-name: android_debug.apk
-          - name: Android-release
-            os: ubuntu-20.04
-            java-version: 11
-            target: template_release
-            base-path: platform/android/java/lib/libs/release
-            apk-name: android_release.apk
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


piiertho:
Can you try this modification on a fork repo to check release is fine ?

chippmann:
> Can you try this modification on a fork repo to check release is fine ?

Hard to do with a broken pipeline ^^. I just tested the release templates locally. I try to find the commit of godot which fixed the pipeline and try to cherry pick that one. Otherwise the test release on the fork has to wait until the next patch version of godot is released

chippmann:
Superseeded by #581

PR #584, opened 2024-03-10
Add option to pass arguments to scene JVM

LukasVykuka:
This add new parameter to `godot_kotlin_configuration.json` to pass argument to the JVM, which runs in the scene regardless of whether the scene was launch from editor or from command line.
My use case is possibility of writing Kotlin code and connecting debugger in IntelliJ and at the same time running scene from Godot editor. For example
```
{"max_string_size":512,"vm_type":"jvm","scene_jvm_args":"-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"}
```
Also it can be used for any JVM arguments. I will add docs after we agree on parameter name.

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -154,6 +154,10 @@ void GDKotlin::init() {
         }
     }
 
+    if (!Engine::get_singleton()->is_editor_hint()) {
+        args.option(configuration.get_scene_jvm_args().utf8());
+    }
+
     if (!jvm_debug_port.is_empty() || !jvm_debug_address.is_empty()) {
         if (jvm_debug_address.is_empty()) {
             jvm_debug_address = "*";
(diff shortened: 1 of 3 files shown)


CedNaru:
So the goal is that you can use those JVM argument even when starting the scene from the editor itself instead of having to launch it from command line only with the argument.
So, the arguments written in the configuration JSON are guaranteed to be used for the non editor usages of the project. Given the usage, wouldn't that be better to call it `debug_jvm_args` or similar ?

LukasVykuka:
I think `debug` is misleading because this parameter can be used to pass any args to JVM (not only to enable debugging) and also can be used in release build of game. So just `jvm_args` can be enough?

piiertho (changes requested):
Few cpp changes.  
@CedNaru and @chippmann do you remember why we did not use project parameters instead of a json file ?

CedNaru:
I think it's mostly for convenience. We can easily edit the json before starting the editor, meanwhile using project parameters require to already start the engine (so after reading said parameters) or to edit a much more complex godot file.

chippmann (changes requested):
LGTM (apart from the arg name. Once that's resolved, I'll happily approve :-) )
Thx a lot!

CedNaru:
I'm fine `jvm_args`

CedNaru:
As part of the overhaul of the C++ code, the argument parsing part of the module is currently being completly changed.
To avoid conflict, I am going to add the feature of this PR into mine. #601

LukasVykuka:
Ok, cool. This PR can be closed, I guess.

PR #592, opened 2024-03-30
Add support for Java source files

CedNaru:
Like for Kotlin files, the module can now handle Java files. They can be attached to Nodes and open in the editor with minimal support.
I did a heavy lifting of the Script and Language classes so they can handle each language individually.
The documentation and logs have been updated to mention Java files and use more generic words like "Jvm" or "Source file" when it applies to both languages.


chippmann (approved):
Works like a charm. Nice work!

Still unhappy about the `JavaScript` naming situation but if we don't find a better name, it should not block merging...

src/gd_kotlin.cpp
```diff
@@ -60,13 +60,6 @@ void register_engine_types_hook(JNIEnv* p_env, jobject p_this, jobjectArray p_en
 }
 
 void GDKotlin::init() {
-    if (Engine::get_singleton()->is_project_manager_hint()) {
```
piiertho, line 63: Why this has been removed ?

src/script/jvm_instance.cpp
```diff
@@ -57,7 +57,7 @@ bool KotlinInstance::get(const StringName& p_name, Variant& r_ret) const {
         return true;
     }
 
-    if (KtFunction* function {kt_class->get_method(SNAME("_get"))}) {
+    if (KtFunction * function {kt_class->get_method(SNAME("_get"))}) {
```
piiertho, line 60: Weird formatting

CedNaru: Clang not dealing properly with {} initializer again.

src/gd_kotlin.cpp
CedNaru: Not removed but moved.
It is in GdjLanguage now, I just don't call GDKotlin::init() at all if in project manager.

piiertho (approved):

Issue #591, opened 2024-03-29
Source files should be empty after export.


CedNaru:
With #579, source files are coming back to Godot Kotin as useable scripts. It also means they will get exported alongside the game. 
To avoid exposing source code that is not even necessary to run the code (the files are only markers at this point), we should update the export plugin so it removes all the content inside Kotlin and Java files.

PR #593, opened 2024-03-31
Source files are now exported without the code.

CedNaru:
Implements #591
Pretty straightforward to implement.
It's not easy to check if the exported files are actually empty because Godot embeds all of them in its own file format full of meta-data and with the content encoded (even without encryption)

So I compared the size of a simple script exported by the editor before/after this commit.
Before: 767 Bytes
After: 241 Bytes

I think it's safe to say that it's working.

### DIFF
--- a/src/kotlin_editor_export_plugin.cpp
+++ b/src/kotlin_editor_export_plugin.cpp
@@ -3,7 +3,11 @@
 #include "kotlin_editor_export_plugin.h"
 
 #include "gd_kotlin.h"
+#include "godotkotlin_defs.h"
 #include "jni/jni_constants.h"
+#include "script/gdj_script.h"
+#include "script/java_script.h"
+#include "script/kotlin_script.h"
 
 #include <core/config/project_settings.h>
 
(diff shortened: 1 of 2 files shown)


chippmann (approved):
Nice addition!

src/kotlin_editor_export_plugin.cpp
```diff
@@ -143,4 +146,36 @@ String KotlinEditorExportPlugin::get_name() const {
...
+    // We simply return a constant as source files are going to be cleared regardless of the configuration.
+    return 0;
+}
+
+Ref<Resource> KotlinEditorExportPlugin::_customize_resource(const Ref<Resource>& p_resource, const String& p_path) {
+    String ext = p_path.get_extension();
+
+    // We create a new resource, otherwise it overwrites the one in the project, not just in the export.
+    // After export, source files are only used to find the right script in the .jar using its path or name.
+    if (ext == GODOT_JVM_REGISTRATION_FILE_EXTENSION) {
```
piiertho, line 164: Do we really need to clean those as this is not source code ?

piiertho (approved):

src/kotlin_editor_export_plugin.cpp
CedNaru: We don't really need to, they only contain the list of properties, method and signal, which is information easily obtainable from the Script API. But its content also has no utility at all for running the game, so I just thought I would be thorough and clean all 3 kinds of scripts equally.

PR #594, opened 2024-04-01
Rework CI/CD Pipeline

chippmann:
This reworks our CI/CD pipeline.

A release example can be found here: https://github.com/chippmann/godot-kotlin-jvm/releases/tag/0.8.3-4.2.1-SNAPSHOT
A Release Run Example can be found here: https://github.com/chippmann/godot-kotlin-jvm/actions/runs/8501035075
A PR Run Example can be found here: https://github.com/chippmann/godot-kotlin-jvm/actions/runs/8501035188

The whole setup closely resembles the godot pipeline but we have some more steps to do.

I suggest to only merge this PR AFTER the next release. As the next release is already quite large and contains important fixes and improvements, we should merge this one after that release in case there should be any issues with this PR.

**Note:** In this PR i commented out one of our tests as it fails on windows.
This needs to be fixed but i think not as part of this PR as it has nothing to do with the changes i made here


chippmann:
Changed base to the gd_kotlin rework branch as per request of @CedNaru . Like this we can test it for a bit while doing the rework and can still do deployment as befor for intermediate releases

CedNaru (approved):

piiertho:
Is it possible to add PDB files to debug release ? This would help a lot to debug.

chippmann:
Here the checks fail (probably because of changes in the rework base branch). The same checks work though in the fork: https://github.com/chippmann/godot-kotlin-jvm/actions/runs/8647770940

I think the issue lies in the rework branch and not in the test execution of this PR

PR #598, opened 2024-04-03
Fix export

CedNaru:
Two issues were solved :
- I forgot to fill the map that link source files to names in release mode.
- Source files were exported with a different extension and not picked up by our custom loader.

I tested:
- Windows Editor (with reloading)
- Windows Export
- Android Export

### DIFF
--- a/src/kotlin_editor_export_plugin.cpp
+++ b/src/kotlin_editor_export_plugin.cpp
@@ -146,35 +146,12 @@ String KotlinEditorExportPlugin::get_name() const {
     return "Godot Kotlin/Jvm";
 }
 
-bool KotlinEditorExportPlugin::_begin_customize_resources(const Ref<EditorExportPlatform>& p_platform, const Vector<String>& p_features) {
-    return true;
-}
-
-uint64_t KotlinEditorExportPlugin::_get_customization_configuration_hash() const {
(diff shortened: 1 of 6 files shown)


src/kotlin_editor_export_plugin.cpp
```diff
@@ -146,35 +146,10 @@ String KotlinEditorExportPlugin::get_name() const {
...
-        exported_script.instantiate();
-        return exported_script;
-    } else if (ext == GODOT_JAVA_SCRIPT_EXTENSION) {
-        Ref<JavaScript> exported_script;
-        exported_script.instantiate();
-        return exported_script;
-    } else {
-        return {};
+    if (ext == GODOT_JVM_REGISTRATION_FILE_EXTENSION || ext == GODOT_KOTLIN_SCRIPT_EXTENSION || ext == GODOT_JAVA_SCRIPT_EXTENSION) {
+        add_file(p_path, Vector<uint8_t>(), true);
```
chippmann, line 152: I would add a comment on what we do here and why

chippmann (changes requested):
Will test exports and editor on linux later today

chippmann (approved):
Editor and exports work on Linux

piiertho (approved):
Tested on iOS

Issue #590, opened 2024-03-14
Add missing print functions to GD singleton.


CedNaru:
It seems we never implement some print methods from the Godot API like push_warning and push_error.
Should be a simple fix requiring to add a few more bridges.

Issue #505, opened 2023-10-10
Add new rotate_toward methods to core types


CedNaru:
Issue used as a memo to remember to update core types when this PR is merged: https://github.com/godotengine/godot/pull/82926

PR #600, opened 2024-04-05
Add new core methods

CedNaru:
Implement #505 

Also sneaking in a small change I wanted for VariantArray and Dictionary. It doesn't change anything for calling code, but it removes the part that complains about a function using the same name as a constructor.

I could have implemented #590 too, but I want to avoid touching bridges in the master with the GDKotlin rework branch around

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/core/math/Basis.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/math/Basis.kt
@@ -678,6 +678,19 @@ class Basis() : CoreType {
     }
 
     /**
+     * Just as before, there's no need for the param inside square brackets.
+     * Use the parameter names directly. Also, keep in mind that the description inside the brackets should match the actual parameter names used in the function signature.
+     */
+    fun rotateToward(to: Basis, delta: RealT): Basis {
+        return Basis(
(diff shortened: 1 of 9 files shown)


kt/godot-library/src/main/kotlin/godot/core/math/Basis.kt
```diff
@@ -677,6 +677,19 @@ class Basis() : CoreType {
         this._z = ret._z
     }
 
+    /**
+     * Just as before, there's no need for the param inside square brackets.
+     * Use the parameter names directly. Also, keep in mind that the description inside the brackets should match the actual parameter names used in the function signature.
```
chippmann, line 682: This java doc is odd. even if it's copied from godot. TBH I'd change that.

chippmann (changes requested):
Also the tests fail atm

kt/godot-library/src/main/kotlin/godot/core/math/Basis.kt
CedNaru: My bad a copied pasted the wrong line

chippmann (approved):

piiertho (approved):

PR #601, opened 2024-04-06
GDKotlin Rework: Jvm configuration

CedNaru:
I split the whole configuration process into smaller functions, each with their responsibilities.
All those functions are static and part of JvmConfiguration. They are then used together by GDKotlin to create a final version of the configuration.

I first parse the JSON file, then the command lines, making sure in both cases that the data provided is correct.
If the JSON ends up being invalid or missing, I write a new one on disk.
Then command line arguments replace the value set by the JSON if they exist.
Last I added a bunch of sanity check to make sure the file is ready to be used for the platform.

Note: I didn't touch the JVM initialization itself. Aside of a few renames because of the changes in JvmConfiguration, it remains the same. The rework of that part is going to be in the next PR.


src/jni/platforms/jvm_default.cpp
```diff
@@ -16,7 +16,7 @@ namespace jni {
 
     JavaVM* Jvm::vm = nullptr;
     jint Jvm::version = 0;
-    Jvm::Type Jvm::vm_type {Jvm::JVM};
+    Jvm::Type Jvm::vm_type {Jvm::NONE};
```
piiertho, line 19: Why is it NONE here ?

CedNaru: If it was a bitmask, it would mean that a VM can be ART and Graal at the same time (for example).
The idea behind none is because I'm using VARIANT_ENUM_CAST so godot can read the enum as Variant value, which make the code easier in some parts, and the 0 value for Variant enum is basically the default for Godot enums (most but not all enums and Godot have a NONE value as the 0).
I know use for the json and command line as the "auto" tag, so the VM will automatically switch to ART on Android and graal on IOS.
Otherwise, with the way I know export the json, it would always be generated with JVM by default. And even if it's not incorrect for desktop anyway, i prefer the automatique mode to be explicit.

piiertho (approved):
Neat !

chippmann (approved):
Nice work!

PR #609, opened 2024-04-15
GDKotlin Rework: Module initialization

CedNaru:
Make the module more resilient to invalid states. Many of the previous explicit crashes got replaced by errors that won't stop the engine execution.
If something goes wrong somewhere in the initialization of the module, a helpful message will appear on-screen before falling back to a Godot with only JVM placeholders.
In debug/tool mode, it's even possible to start the game using JVM scripts that are not built yet. In such case, the script is just not created and only the naked native object will remain (Errors are still thrown in such cases).

GDKotlin got explicit initialization states and a custom error message exists for almost each of them (when the cause is likely a basic one) in case an issue was encountered.
If the final state is invalid, GDkotlin will be immediately freed and unloaded. The languages and resource loader are still registered to Godot, even in the case of an invalid state, so users can manipulate JVM scripts in the editor without a JAVA_HOME, embedded JRE, boostrap.jar or main.jar (They will have an alert appearing on-screen when the engine starts).

image

Note that the last part of the initialization still has to be done in GdjLanguage. This final step fetches all the Jvm Scripts from Kotlin to C++ as well as sending all method pointers from C++ to Kotlin.
This latter detail matters because this operation requires all classes to be initialized in Godot, so we can't do it as part of the module initialization itself, even using the latest initialization enum value in it. Sadly Godot doesn't have a "afterAll" phase for module initialization.


src/binding/kotlin_binding.cpp
```diff
@@ -37,7 +37,7 @@ bool KotlinBinding::refcount_decremented_unsafe() {
 }
 
 void KotlinBinding::set_kt_binding(jni::JObject j_object) {
-    JVM_CRASH_COND_MSG(kt_binding, "Trying to set a KtBinding on an already bound Object");
+    JVM_DEV_ASSERT(!kt_binding, "Trying to set a KtBinding on an already bound Object");
```
chippmann, line 40: What would this mean if this happens in a non dev build?

src/gd_kotlin.cpp
```diff
@@ -325,7 +341,65 @@ String GDKotlin::get_path_to_native_image() {
...
+            cause = "Couldn't open Graal Native Image.";
+            hint = "Make sure you have built your JVM project with Graal native image enabled in your gradle build.";
+        }
+#endif
+    }
+
+    if (state == State::JVM_LIBRARY_LOADED) {
+        invalid = true;
+        cause = "Couldn't start the JVM.";
+        hint = "Check your configuration file and command-lines argument for any invalid setting, including your "
```
chippmann, line 377: command-line arguments

src/jvm_wrapper/bootstrap.cpp
```diff
@@ -36,15 +45,11 @@ void Bootstrap::register_engine_type(JNIEnv* p_env, jobject p_this, jobjectArray
...
-
 void Bootstrap::init(jni::Env& p_env, const String& p_project_path, const String& p_jar_file, const jni::JObject& p_class_loader) {
     LOCAL_FRAME(2);
     jni::JObject project_path = p_env.new_string(p_project_path.utf8().get_data());
     jni::JObject jar_file {p_env.new_string(p_jar_file.utf8().get_data())};
-    jvalue args[5] = {
-      jni::to_jni_arg(project_path),
-      jni::to_jni_arg(jar_file),
-      jni::to_jni_arg(p_class_loader)};
+    jvalue args[5] = {jni::to_jni_arg(project_path), jni::to_jni_arg(jar_file), jni::to_jni_arg(p_class_loader)};
```
chippmann, line 52: 5? shouldn't it be 3?

(not from your PR but just noticed)

src/lifecycle/jvm_manager.cpp
```diff
@@ -70,17 +71,19 @@ void JvmManager::initialize_or_get_jvm(void* lib_handle, JvmUserConfiguration& u
...
-    jint result {get_create_jvm_function(lib_handle)(&java_vm, reinterpret_cast<void**>(&jni_env), &args)};
+    CreateJavaVM func {get_create_jvm_function(lib_handle)};
+    JVM_ERR_FAIL_COND_V_MSG(func == nullptr, false,"Failed to obtain JNI_CreateJavaVM symbol from dynamic library!");
+    jint result {func(&java_vm, reinterpret_cast<void**>(&jni_env), &args)};
 
     // Set std::local::global to value it was before creating JVM.
     // See https://github.com/utopia-rise/godot-kotlin-jvm/issues/166
     // and https://github.com/utopia-rise/godot-kotlin-jvm/issues/170
-#ifndef NO_USE_STDLIB
+#ifndef NO_USE_STDLIBss
```
chippmann, line 81: `NO_USE_STDLIBss` ? why `ss`?

src/binding/kotlin_binding.cpp
CedNaru: Hard to say, this check is really just for debugging purpose in case we did something wrong when synchronizing bindings with their respective JVM instance.
Having this Assert triggers means that we either created 2 bindings for the same KtObject, or that the binding got set on the wrong native object and something is wrong in our memory manager.
It's not supposed to be anything the user can influence.

src/jvm_wrapper/bootstrap.cpp
CedNaru: True, I forgot to reduce the size of the array in the JVM loading PR after changing the signature of the method.

src/lifecycle/jvm_manager.cpp
CedNaru: Because I'm a grammar nazi obviously.
But just a mistake I probably had the wrong window selected when trying to type something and didn't pay attention (having to often switch between Clion and Intellij).

src/gd_kotlin.cpp
```diff
@@ -325,7 +340,65 @@ String GDKotlin::get_path_to_native_image() {
...
 void GDKotlin::unload_dynamic_lib() {
     if (OS::get_singleton()->close_dynamic_library(jvm_dynamic_library_handle) != OK) {
-        JVM_CRASH_NOW_MSG("Failed to close the jvm dynamic library!");
+        JVM_ERR_FAIL_MSG("Failed to close the jvm dynamic library!");
+    }
+}
+#endif
+
+#ifdef DEBUG_ENABLED
+void GDKotlin::validate_state() {
```
chippmann, line 349: I would probably move this to a separate class no? we could centralize the checks once we add more. Also for once we add back the editor plugin we could do some checks there.

But I'm also fine if we think about this later

chippmann (approved):
Excellent job. Not much to complain.

Just briefly tested build, startup and reloading on the macos arm editor.
A simple error case was also tested.

src/lifecycle/jvm_manager.cpp
```diff
@@ -92,33 +89,35 @@ void JvmManager::initialize_or_get_jvm(void* lib_handle, JvmUserConfiguration& u
 #endif
 
     jni::Jvm::initialize(java_vm, user_configuration.vm_type, jvm_options.version);
+    return true;
 }
 
-void JvmManager::initialize_jni_classes(jni::Env& p_env, ClassLoader* class_loader) {
+bool JvmManager::initialize_jni_classes(jni::Env& p_env, ClassLoader* class_loader) {
+    bool loaded_properly;
```
piiertho, line 96: Simplification:
instead of ```cpp
bool loaded_properly;
loaded_properly = TransferContext::initialize(p_env, class_loader);
loaded_properly = loaded_properly && TypeManager::initialize(p_env, class_loader);
...
```

We can write:
```cpp
bool loaded_properly {
    TransferContext::initialize(p_env, class_loader) &&
    TypeManager::initialize(p_env, class_loader) &&
    ...
};

piiertho (approved):

PR #612, opened 2024-04-17
Update to Godot 4.2.2

CedNaru:
I removed the cherry pick in the CI.

Note: The fix regarding exporting static lib for iOS is part of the 4.2.2 release


kt/api-generator/src/main/resources/api.json
```diff
@@ -2,10 +2,10 @@
 	"header": {
 		"version_major": 4,
 		"version_minor": 2,
-		"version_patch": 1,
+		"version_patch": 2,
 		"version_status": "stable",
-		"version_build": "custom_build",
-		"version_full_name": "Godot Engine v4.2.1.stable.custom_build"
+		"version_build": "official",
```
chippmann, line 7: Why is it suddenly `official`?

CedNaru: I think it's because I downloaded the api.json they provide instead of generating it from our own binary.

chippmann: I would probably generate it locally

piiertho (approved):

chippmann (approved):

PR #614, opened 2024-04-18
Improve ci cd workflow

chippmann:
This PR contains the following changes:
- hopefully improves our cache usage by only using the cache name and no refs for cache creation
- always prints the testoutput now (including the error out)
- Adds a workflow which builds and tests on every commit for linux only but does not (fully) run on pull requests


CedNaru:
I don't think this is the right approach to save on cache. I have read the documentation in details. 
Caches are created per key AND per branch and can only be accessed by that branch or a child branch.

Let's take this PR as en example. This branch is a child of `gd_kotlin_rework`, itself a child of `master`
If you try to fetch a cache here, it's going to check for the key in this branch's cache, then in `gd_kotlin_rework` then in `master`.
Here is the relevant part in the doc:
image


But afterward, a new cache will be saved automatically by the action, a cache that will only be accessible from this PR. 
It means that none of the work cached in that branch will be useable by the others.
The only way I see to properly share cache between branches is to only save the cache when running on the master and prevent saving on other branches. Or at least to only save cache on merge actions, that way we only generate caches for "parent branches" whatever they are (in this case, that would be cached for gd_kotlin_rework).
Preventing generating cache on "leaf" branches is important so we don't bloat the cache with each PR:
image


So you should keep your different key for every kind of build (release, debug and deb should all be a separate key).
And instead only save when the conditions are met (either you are on master or triggered by a merge action if that's possible) so the amount of generated cache is kept to a minimum.
To be more granular with restoring and saving caches, you can use the `actions/cache/restore` and `actions/cache/save` instead of the more generic `actions/cache` that both restore and save automatically.

PR #621, opened 2024-04-27
CPP reloading

CedNaru:
The user code is now reloaded from the C++ code.
This doesn't get rid of the boostrap yet. For now it's just moving the new jar detection and usercode loading from Kotlin to C++.


harness/tests/.gitignore
```diff
@@ -5,11 +5,12 @@
 # Ignore Gradle build output directory
 build
 
-jre-arm64-*
-jre-amd64-*
+jvm/jre-arm64-*
```
chippmann, line 8: These are not needed anymore if you add the jvm folder below

kt/godot-library/src/main/kotlin/godot/runtime/Bootstrap.kt
```diff
@@ -36,67 +26,18 @@ internal class Bootstrap {
...
-                }, 3, 3, TimeUnit.SECONDS)
-            }
+            doInit(userCodeFile.toURI().toURL(), loader)
         }
     }
 
     fun finish() {
-        executor?.shutdown()
-        watchService?.close()
+        Thread.currentThread().contextClassLoader = null
```
chippmann, line 34: Why do we set the context classloader to null here explicitly?

kt/godot-library/src/main/kotlin/godot/runtime/Bootstrap.kt
```diff
@@ -10,22 +10,12 @@ import godot.util.JVMLogging
...
-import java.nio.file.WatchService
 import java.util.*
-import java.util.concurrent.Executors
-import java.util.concurrent.ScheduledExecutorService
-import java.util.concurrent.TimeUnit
 
 
 internal class Bootstrap {
-    private val classRegistries: MutableList<ClassRegistry> = mutableListOf()
     private lateinit var classloader: ClassLoader
```
chippmann, line 17: Saving these as explicit class variables is no longer necessary isn't it if you call `doInit` each time anyway?

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupTasks.kt
```diff
@@ -76,28 +73,30 @@ fun Project.setupTasks() {
             )
             // END: graal native image specific tasks
 
+            val copyJarTask = createCopyJarsTask(
+                packageBootstrapJarTask = packageBootstrapJarTask,
+                packageMainJarTask = packageMainJarTask,
+            )
+
             @Suppress("UNUSED_VARIABLE")
             val buildTask = setupBuildTask(
```
chippmann, line 82: I know i was the one who wrote the gradle changes; but the variable here is no longer necessary as we now directly configure the jar tasks instead of the too generic `build` task which does way more than we need anyway.

Regarding this; it would probably also make sense to change the task which get's invoked by the editor plugin upon build. to use a task which produces the jar's rather than `build` which not only produces chars but also runs linter's and tests.
So probably something like `assemble` or `shadowJar`. But that can also be done in a separate PR

src/jvm_wrapper/bridge/packed_color_array_bridge.h
```diff
@@ -6,6 +6,13 @@
 namespace bridges {
     PACKED_ARRAY_BRIDGE(PackedColorArrayBridge, Color, "godot.core.PackedColorArray$Bridge") {
         PACKED_ARRAY_BRIDGE_CLASS(PackedColorArrayBridge, Color)
+
+          // clang-format off
```
chippmann, line 10: Why format off and on here?
would make sense to format it here to be multiline no?

src/jvm_wrapper/bridge/packed_string_array_bridge.h
```diff
@@ -7,6 +7,12 @@ namespace bridges {
 
     PACKED_ARRAY_BRIDGE(PackedStringArrayBridge, String, "godot.core.PackedStringArray$Bridge") {
         PACKED_ARRAY_BRIDGE_CLASS(PackedStringArrayBridge, String)
+
+        // clang-format off
```
chippmann, line 11: same here (and some more later)

src/lifecycle/paths.h
```diff
@@ -1,36 +1,33 @@
 #ifndef GODOT_JVM_PATHS_H
 #define GODOT_JVM_PATHS_H
 
-static constexpr const char* DESKTOP_BOOTSTRAP_FILE {"godot-bootstrap.jar"};
-static constexpr const char* DESKTOP_USER_CODE_FILE {"main.jar"};
+static constexpr const char* DESKTOP_BOOTSTRAP_FILE {"jvm/godot-bootstrap.jar"};
```
chippmann, line 4: We have the folder `jvm` as a separate constant for the kotlin code already. Can we do this for the cpp code as well please? Then it would be consistent and easy to change.

src/resource_format/java_archive.h
```diff
@@ -0,0 +1,13 @@
+#ifndef GODOT_JVM_JAVA_ARCHIVE_H
+#define GODOT_JVM_JAVA_ARCHIVE_H
+
+#include <core/io/resource.h>
+
+class JavaArchive : public Resource {
```
chippmann, line 6: Isn't a jar technically a `JvmArchive` and not a `JavaArchive`? I'd rather name it as such

src/resource_format/java_archive_resource_format_loader.cpp
```diff
@@ -0,0 +1,43 @@
+
+#include "java_archive_resource_format_loader.h"
+
+#include "gd_kotlin.h"
+#include "java_archive.h"
+#include "lifecycle/paths.h"
+#include "logging.h"
+
+void JavaArchiveFormatLoader::get_recognized_extensions(List<String>* p_extensions) const {
```
chippmann, line 9: Same here

src/resource_format/java_archive_resource_format_loader.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+Ref<Resource> JavaArchiveFormatLoader::load(
+  const String& p_path,
+  const String& p_original_path,
+  Error* r_error,
+  bool p_use_sub_threads,
+  float* r_progress,
+  ResourceFormatLoader::CacheMode p_cache_mode
+) {
+    JVM_LOG_VERBOSE(vformat("Loading usercode file at: %s", p_path));
```
chippmann, line 32: This technically no longer has to be "usercode". I'd just state that we load a JvmArchive

src/script/jvm_script_manager.cpp
```diff
@@ -206,13 +206,17 @@ int64_t JvmScriptManager::get_last_reload() {
...
+    JvmScriptManager* singleton = get_instance();
+
+    singleton->named_scripts.clear();
+    singleton->named_scripts_map.clear();
+    singleton->name_to_filepath_map.clear();
+    singleton->filepath_to_name_map.clear();
+    singleton->path_scripts.clear();
+    singleton->path_scripts_map.clear();
+
+    //memdelete(singleton);
```
chippmann, line 219: Dead code

src/gd_kotlin.cpp
```diff
@@ -83,8 +176,9 @@ void GDKotlin::set_jvm_options() {
     if (user_configuration.jvm_jmx_port >= 0) { jvm_options.add_jmx_option(user_configuration.jvm_jmx_port); }
 
     if (!Engine::get_singleton()->is_editor_hint() && !user_configuration.jvm_args.is_empty()) {
-        JVM_LOG_WARNING("You are using custom arguments for the JVM. Make sure they are valid or you risk the JVM to not "
-                    "launch properly");
+        JVM_LOG_WARNING("You are using custom arguments for the JVM. Make sure they are valid or you risk the JVM to "
+                        "not "
```
chippmann, line 180: I don't know what this formatting is, but it's even worse than before xD

src/gd_kotlin.cpp
```diff
@@ -98,27 +192,22 @@ void GDKotlin::set_jvm_options() {
...
-
-    Error err;
-    Ref<DirAccess> dir_access {DirAccess::open(BUILD_DIRECTORY, &err)};
-
-    JVM_ERR_FAIL_COND_V_MSG(err != OK, "", "Cannot open %s file in res://.", file_name);
-
-    dir_access->copy(file_res_path, file_user_path);
-
+        Ref<DirAccess> dir_access {DirAccess::open(USER_DIRECTORY)};
+        dir_access->make_dir("jvm");
```
chippmann, line 209: Please use the cpp constant here as well once you created it

src/gd_kotlin.cpp
```diff
@@ -98,27 +192,22 @@ void GDKotlin::set_jvm_options() {
...
     if (!FileAccess::exists(file_user_path) || FileAccess::get_md5(file_user_path) != FileAccess::get_md5(file_res_path)) {
-        JVM_LOG_VERBOSE("%s file has changed. Copying it from res:// to user://.", file_name);
+        JVM_LOG_VERBOSE("%s file has changed. Copying it from %s to %s.", file_name, file_res_path, file_user_path);
 #else
     // as per suggestion of https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading, we first delete existing files and then copy them again
     // if we don't do this, subsequent app starts where the files already exist, error out
 
     String file_user_path_global {ProjectSettings::get_singleton()->globalize_path(file_user_path)};
-    unlink(file_user_path_global.utf8().get_data()); // we do not really care about errors here
+    unlink(file_user_path_global.utf8().get_data());// we do not really care about errors here
```
chippmann, line 206: Space should be kept

src/kotlin_editor_export_plugin.cpp
```diff
@@ -129,11 +130,11 @@ void KotlinEditorExportPlugin::_export_begin(const HashSet<String>& p_features,
...
         }
     } else if (is_android_export) {
-        files_to_add.push_back(String(BUILD_DIRECTORY).path_join(ANDROID_BOOTSTRAP_FILE));
-        files_to_add.push_back(String(BUILD_DIRECTORY).path_join(ANDROID_USER_CODE_FILE));
+        files_to_add.push_back(String(RES_DIRECTORY).path_join(ANDROID_BOOTSTRAP_FILE));
+        files_to_add.push_back(String(RES_DIRECTORY).path_join(ANDROID_USER_CODE_FILE));
         _generate_export_configuration_file(jni::JvmType::ART);
     } else if (is_ios_export) {
-        String base_ios_build_dir {String(BUILD_DIRECTORY).path_join("ios") };
+        String base_ios_build_dir {String(RES_DIRECTORY).path_join("jvm/ios")};
```
chippmann, line 137: Use the cpp constant once created

chippmann (changes requested):

kt/godot-library/src/main/kotlin/godot/runtime/Bootstrap.kt
CedNaru: It was already there before. We first set it from the C++ side, but somehow remove it from the Kotlin side.

src/jvm_wrapper/bridge/packed_color_array_bridge.h
CedNaru: because clang-format is a terrible tool

src/script/jvm_script_manager.cpp
CedNaru: Not dead, I just forgot to uncomment it

src/resource_format/java_archive.h
CedNaru: JAR stands for Java Archive

src/gd_kotlin.cpp
chippmann: as discussed on discord; not worth the effort

src/kotlin_editor_export_plugin.cpp
chippmann: as discussed on discord; not worth the effort

src/lifecycle/paths.h
chippmann: as discussed on discord; not worth the effort

chippmann (dismissed):

src/editor/build/build_manager.cpp
```diff
@@ -49,14 +49,19 @@ bool BuildManager::build_project_blocking() {
         // When in blocking mode, only make the window appears when it fails
         GodotKotlinJvmEditor::get_instance()->update_build_dialog(build_log);
         return false;
+    } else {
```
piiertho, line 52: Simplification: reduce nesting
```cpp
if (build_project() != Error::OK) {
      // When in blocking mode, only make the window appears when it fails
      GodotKotlinJvmEditor::get_instance()->update_build_dialog(build_log);
      return false;
}
GodotKotlinJvmEditor::get_instance()->on_build_finished();
return true;
  ```

PR #623, opened 2024-05-07
fix(doc): place update ios jdk contribution guide to the right place

piiertho:


chippmann (dismissed):

CedNaru (changes requested):
You created a new file, the one at the wrong place was not deleted/moved so it means we would have 2 of them on the master if we merge like this.
Also this new files misses the part mentioning you have to update the download link to GraalVM when updating.

chippmann (approved):

CedNaru (approved):

PR #625, opened 2024-05-09
Support multidex and fix export for android 14 devices

chippmann:
This implements multidex support for both usercode jar and bootstrap jar:
2024-05-09T22:02:35,106757814+02:00

This also fixes an issue regarding dynamic code loading on android 14 when built with the corresponding target sdk version: https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading

Android export tested on our tests project.
Will test the 3D demo project tomorrow as well

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -99,12 +99,16 @@ String GDKotlin::copy_new_file_to_user_dir(const String& file_name) {
     String file_res_path {String(BUILD_DIRECTORY) + file_name};
     String file_user_path {String(USER_DIRECTORY) + file_name};
 
-#ifdef __ANDROID__
+#ifndef __ANDROID__
+    if (!FileAccess::exists(file_user_path) || FileAccess::get_md5(file_user_path) != FileAccess::get_md5(file_res_path)) {
+        LOG_VERBOSE(vformat("%s file has changed. Copying it from res:// to user://.", file_name));
+#else
     // as per suggestion of https://developer.android.com/about/versions/14/behavior-changes-14#safer-dynamic-code-loading, we first delete existing files and then copy them again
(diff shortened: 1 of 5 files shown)


chippmann:
Tested on the template project as well: 
Screenshot_20240510_100822

On the 3D demo i still encounter some issues but I'm not yet sure if they really stem from this PR given that that demo was never smooth sailing on mobile

chippmann:
Ok the error in the 3D demo is also present on desktop but there it just does not crash the application which it does no android it seems. But i guess it's safe to say that this is unrelated to this PR

src/gd_kotlin.cpp
```diff
@@ -91,10 +91,28 @@ void GDKotlin::set_jvm_options() {
...
 
+#ifdef __ANDROID__
+#include <unistd.h>
+#endif
+
 String GDKotlin::copy_new_file_to_user_dir(const String& file_name) {
     String file_res_path {String(BUILD_DIRECTORY) + file_name};
     String file_user_path {String(USER_DIRECTORY) + file_name};
 
+#ifdef __ANDROID__
```
CedNaru, line 102: More code can be shared between the 2 preprocessors blocks. Also we don't always copy those files, only when we detect a change in the file, is that a requirement that Android makes a fresh copy each time now ?

CedNaru (changes requested):

src/gd_kotlin.cpp
chippmann: As i tried to convey in the comment one line below, android recommends copying dynamically loaded content each time the app starts: https://developer.android.com/about/versions/14/behavior-changes-14#handle-existing-files

piiertho (dismissed):

CedNaru (approved):

piiertho (approved):

Issue #617, opened 2024-04-21
Fatal Error in native method: bad global or local ref passed to JNI


MartinHaeusler:
The following happened randomly to me today while test playing my game:

```
FATAL ERROR in native method: Bad global or local ref passed to JNI
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x9b28bf]  ReportJNIFatalError+0x2f
V  [libjvm.so+0x9bbc6a]
V  [libjvm.so+0x9be5e6]  checked_jni_DeleteGlobalRef+0x96
C  [godot.linuxbsd.editor.x86_64+0xca05da]
```

Unfortunately that's all I have, there were no more logs available from this particular crash. I don't know which JNI method it was, nor which reference was problematic. I thought I should report this anyway, perhaps we can at least improve the logging to include the called method name or other diagnostic information.

piiertho:
We will add debug symbols for debug build in a next release.  
Which version are you using ? We can still disassemble binary and look where it crashes using adresse

MartinHaeusler:
@piiertho I'm currently on `0.8.1-4.2.0`. Hoping for a new version that targets Godot 4.2.2 soon :)

CedNaru:
It at least tells us it's related to deleting a global ref from the c++ side, and if it rarely happens and randomly, it means it's a race condition.
Most likely, it's the issue I talked about here:
https://github.com/utopia-rise/godot-kotlin-jvm/issues/616

CedNaru:
It's very unlikely to be your mistake, users code is not supposed to have any influence on how the module handle the JNI ref.
It's random because it's probably a race condition between Godot main thread and the MemoryManager thread we use. 
I have a few ideas of the origin of it.

MartinHaeusler:
@CedNaru I saw your proposal in #618 and for what my opinion is worth, the proposal makes sense to me. Maybe it could also be worth a try to see how it's done for the Godot C# module? After all, in the grand scheme of things, the CLR and the JVM are similar beasts and the CLR has its own GC as well.

Any attempt at fixing this issue would be highly appreciated, as it ultimately is a showstopper. Individual issues I can work around; random crashes are a different story.

CedNaru:
I already checked how C# does and actually, the second point of my proposal is to move away from the way C# is doing things. 

[...]

Issue #616, opened 2024-04-18
Current Memory Manager can cause RefCounted instances to be freed incorrectly.


CedNaru:
This is quite complex to explain so hang in there.
The following example is extreme, but its purpose is to show that incorrect behaviours are possible with the way we handle memory within the binding.
Take the following Kotlin Script:
```kt
@RegisterClass
class MyScript {
 
  private var ref: RefCounted? = RefCounted()
  
  @RegisterFunction
  fun getRef(): RefCounted{
    val ret = ref
    ref = null
    return ret
  }
}
```

In this configuration, `ref` is the only reference to the RefCounted instance in both the JVM and C++. Let's say that some time happens between the creation of this script and the call to `getRef()`so we are sure that `ref` is bound by the MemoryManage

Now we have this GDScript (but it could be the internals of any Godot native type as well).
```gdscript

var ref: Refcounted
var my_script := MyScript.new()

func _enter_tree() -> void
    ref = my_script.get_ref()

```

We prepare an instance of this gdscript and add it later to the scene tree (so MyScript is created, and its RefCounted as well).
When `getRef() `get called, we set `ref` to null and return. The native pointer of the RefCounted is going to be placed in the SharedBuffer but once we return from the JVM, there will be no reference left of the RefCounted (`ref` is null and the local variable `ret` is no longer in the scope)
It means that between the moment we return from the JVM and the moment the pointer of the RefCounted instance is parsed from the buffer and converted back to a Variant or Ref<>, there will be nothing keeping it alive. Its counter won't immediately reach 0 because it can only go from 1 to 0 when the MemoryManager calls CPP, which will happen.
When you return from the JVM, all reference to `ref` are gone, and so the instance will be collected and the MemoryManager will use the GC callback to call CPP and decrement the counter.

At this moment, 2 things can happen depending on the timing of the 2 threads (the one running the scripts and the one running the MemoryManager):
- If the script thread is faster, then the Refcounted will be sent to C++ and incremented again, which means the native instance won't die but the JVM wrapper will. If the RefCounted is sent to the JVM again by some other mean afterward, it will try to use the dead weak reference stored in the MemoryManager (because the MemoryManager is the slower one in this situation, the entry in its ObjectDB hasn't been cleared yet).
[...]

PR #626, opened 2024-05-09
Rework Memory manager to make it more threadsafe.

### LINKED ISSUE #615: Memory leaks in the TransferContext class
Right now we have the following code in TransferContext
````cpp
const int MAX_STACK_SIZE = MAX_FUNCTION_ARG_COUNT * 8;

thread_local static Variant variant_args[MAX_STACK_SIZE];// NOLINT(cert-err58-cpp)
thread_local static const Variant* variant_args_ptr[MAX_STACK_SIZE];
thread_local static int stack_offset = -1;
```

The reason for that is that we need to store the variables somewhere once they got parsed from the shared buffer. Godot calls take an array of Variant pointers as an argument, so we use the second array to store the pointer and send it directly as an argument.
Now, the reason for the stack size is that a lot of JVM <=> C++ exchanges can happen in a single call. (Jvm calling cpp calling Jvm calling cpp caling etc...). Because the buffer is overwritten each time, we can't keep the data from the first layer of call, hence the long stack to store them.
We have this dynamic for every thread. This design has leaks because data are kept in those Variant array until overwritten. 
Let's say you make a icall with a lot of parameters (let's 10), when most icalls are not going to go above 4. It means that the 6 Variant afer that will remain in memory. That's not an issue when storing value based Variant (like Vector2) but it is when storing RefCounted or even pointer based Variant like Dictionary (with their own atomic counters). They will never be freed.
Even worse, they can be the cause of leaks when terminating the program, the TransferContext destructor looks like this:

````cpp
TransferContext::~TransferContext() {
    for (auto& variant_arg : variant_args) {
        variant_arg = Variant();
    }
}
```
 We free everything in the array but it can only happen for the Array specific to the main thread. All the array related to other thread will not be freed and cause memory leaks that the ObjectDB will report.

We have 2 ways to fix that:
- Manually clear the whole array after every icall (performance cost because we have to erase previous values by replacing them with newly constructed Nil Variant, to be benchmarked).
- Entirely rework the way we do icalls.

The first option is self-explanatory.
The second is quite more complex. So far we have been using `MethodBind::call()` which require a Variant**. Godot has another a second method `MethodBind::ptrcall()` which require a `void**`. This method is said to be more efficient, MethodBind being typed with templates, it can directly reinterpret the pointer to their correct type. 
Doing so would avoid using an intermediary Variant. So far the data conversions has been the following:
SharedBuffer -> Core Type -> Variant -> `MethodBind::call()` -> Core.
By using `MethodBind::ptrcall()`; it could become:
SharedBuffer -> Core Type -> `MethodBind::call()` -> No Conversion needed in the function.

We would still keep a stack like today, except it would be not be Variant any more but a generic typeless stack. The main difference would be that we will only store the value based core type in it. For the other pointer based types, we allocated a native memory slot when they were created on the JVM side (which is something we also have to optimize with PoolAllocators) so we can use the same pointer for the MethodBind. The consequence would be to not keep those refcounted type in the stack, instead they would be managed by the Kotlin Memory Manager (like they already are). 
I won't lie, this rework would be complicated to implement and test.

But our 2 choices are either :
- Simple but a bit less efficient than today
- Complex but more efficient than today.
The gain and loss of each are of course hard to know before actually trying.

CedNaru:
fixes #615 #616 #617

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/core/callable/KtCallable.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/callable/KtCallable.kt
@@ -1,8 +1,9 @@
 package godot.core.callable
 
 import godot.core.KtObject
-import godot.core.memory.TransferContext
 import godot.core.VariantType
+import godot.core.memory.TransferContext
+import godot.global.GD
 import godot.tools.common.constants.Constraints
 import godot.util.threadLocal
 
@@ -17,10 +18,27 @@ abstract class KtCallable<T : KtObject, R : Any?>(
 
     fun invoke(instance: T) {
         TransferContext.readArguments(types, isNullables, paramsArray)
-        val ret = invokeKt(instance)
+        try {
+            invokeKt(instance)
+        } catch (e: Exception) {
+            GD.printErr("Error calling a JVM method from Godot:", e.stackTrace)
+        }
         resetParamsArray()
+    }
 
-        TransferContext.writeReturnValue(ret, variantType)
+    fun invokeWithReturn(instance: T): Any? {
+        TransferContext.readArguments(types, isNullables, paramsArray)
+
+        var ret: Any? = Unit
+        try {
+            ret = invokeKt(instance)
+            TransferContext.writeReturnValue(ret, variantType)
+        } catch (e: Exception) {
+            GD.printErr("Error calling a JVM method from Godot:", e.stackTrace)
+            TransferContext.writeReturnValue(null, VariantType.NIL)
+        }
+        resetParamsArray()
+        return ret
     }
 
     companion object {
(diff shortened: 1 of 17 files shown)


kt/godot-library/src/main/kotlin/godot/core/callable/KtCallable.kt
```diff
@@ -17,10 +18,27 @@ abstract class KtCallable<T : KtObject, R : Any?>(
 
     fun invoke(instance: T) {
         TransferContext.readArguments(types, isNullables, paramsArray)
-        val ret = invokeKt(instance)
+        try {
+            invokeKt(instance)
+        } catch (e: Exception) {
```
chippmann, line 23: Shouldn't we catch `Throwable` in these instead as these would also catch `RuntimeExceptions` and `Error` instances?

chippmann (dismissed):
Just one question. Apart from that LGTM

kt/godot-library/src/main/kotlin/godot/core/callable/KtCallable.kt
CedNaru: RuntimeExceptions are already exception. And regarding Errors, I forgot about them, it's not usually something you recover from, but I can do the change regardless.

chippmann (dismissed):

piiertho (approved):

chippmann (approved):

PR #627, opened 2024-05-10
Add better jni error reporting inside the godot editor

chippmann:
This improves the jni error reporting inside the godot editor by grabbing the stacktrace from the jvm and printing them using the godot print commands. Only applies to tool mode. In all other cases, we still print the error using the jni provided method which just prints to the console.

2024-05-10T12:05:06,018362263+02:00
2024-05-10T11:57:53,693923600+02:00
2024-05-10T12:04:09,201022038+02:00
2024-05-10T12:00:49,301472094+02:00

### DIFF
--- a/src/jni/env.cpp
+++ b/src/jni/env.cpp
@@ -59,10 +59,8 @@ namespace jni {
                 jmethodID toStringMethod = env->GetMethodID(stringWriterClass, "toString", "()Ljava/lang/String;");
                 auto jStackTraceString = (jstring) env->CallObjectMethod(stringWriter, toStringMethod);
 
-                // Convert Java String to C++ string
                 const char* cStackTrace = env->GetStringUTFChars(jStackTraceString, nullptr);
                 String stackTrace {cStackTrace};
-                // Release memory
                 env->ReleaseStringUTFChars(jStackTraceString, cStackTrace);
 
                 return stackTrace;
--- a/src/jni/env.cpp
+++ b/src/jni/env.cpp
@@ -39,7 +39,7 @@ namespace jni {
     }
 
     String Env::exception_describe() {
-#ifdef TOOLS_ENABLED
+#ifdef DEBUG_ENABLED
         jthrowable e = env->ExceptionOccurred();
 
         if (e != nullptr) {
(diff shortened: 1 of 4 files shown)


CedNaru:
There is a partial conflict with my new PR.
I made the JNI check optional for method calls and instead use a try catch directly on the Kotlin side. It's more efficient to catch exception when possible directly from the JVM instead of doing it form the C++ side. 
See: 
https://github.com/utopia-rise/godot-kotlin-jvm/pull/626/commits/fd5fbb2a3eeb43dec6ffc6240da3c0ca7545a93b

chippmann:
> There is a partial conflict with my new PR. I made the JNI check optional for method calls and instead use a try catch directly on the Kotlin side. It's more efficient to catch exception when possible directly from the JVM instead of doing it from the C++ side. See: [fd5fbb2](https://github.com/utopia-rise/godot-kotlin-jvm/commit/fd5fbb2a3eeb43dec6ffc6240da3c0ca7545a93b)

Ok for me. I'll just wait with this PR until yours is merged and then fix and rebase

src/jni/env.cpp
```diff
@@ -38,8 +38,36 @@ namespace jni {
         return env->ExceptionCheck();
     }
 
-    void Env::exception_describe() {
+    String Env::exception_describe() {
+#ifdef TOOLS_ENABLED
```
CedNaru, line 42: Why is that Tool only and not debug ?

chippmann: Because I'm a dum dum and somehow had in mind that games run from the editor are tools_enabled. Will fix this

CedNaru (dismissed):

src/jni/env.cpp
```diff
@@ -38,8 +38,36 @@ namespace jni {
         return env->ExceptionCheck();
     }
 
-    void Env::exception_describe() {
+    String Env::exception_describe() {
+#ifdef DEBUG_ENABLED
+        jthrowable e = env->ExceptionOccurred();
+
+        if (e != nullptr) {
```
piiertho, line 45: `if (ptr != nullptr)` is redundant

piiertho (changes requested):

piiertho (approved):

PR #629, opened 2024-05-10
Add callable convenience functions

chippmann:
This adds convenience function to convert a kotlin function to a callable.
It's main use is typesafe conversion if one needs method binds.
As with godot 4, method binds are no longer in the corresponding call functions, but on a callable.
So with our binding, one needed to create a callable with string names instead of being able to use typesafe conversion functions like we provide for other things like signals.
2024-05-10T13:24:45,409946107+02:00

Now one can write code like this:


chippmann:
After brief discussion on discord i changed the convenience extension functions from being extension functions to invoke operator functions on the callable companion object. Like this, we're more in line with the regular godot constructors for callable while retaining type safety.

The new example looks like this: `Callable(this, CallableMethodBindTest::readySignalMethodBindTest).bind(1, 2, 3).call()`

CedNaru (approved):

piiertho (approved):

PR #630, opened 2024-05-11
enh: Simplify cpp core type buffer protocol read and write

piiertho:
This simplify cpp read and write buffer protocol for core types.  
This reinterprets byte buffer to core type to avoid multiple calls to get each primitives.  
This needs test on all platforms to be sure we don't have memory layout troubles, even if we shouldn't as all core types are perfectly aligned.


CedNaru (approved):

CedNaru (dismissed):

chippmann (approved):
Tested on Linux editor and linux export templates

CedNaru:
Test on Android

CedNaru (approved):

chippmann (approved):

PR #632, opened 2024-05-12
Make jre folder platform dependent

chippmann:
This makes the jre folder platform dependent.

This makes it possible for a project to have multiple jre folder simultaneously on the file system. The module then automatically picks the correct one for the current platform.
The main advantage of this is that on a project with multiple people working on it, the embedded jre can be committed to the repository for each platform together with the built jar and a project member does not need to have java locally installed and JAVA_HOME set. Which is especially useful if one is working with non programmers.

**Note:** Breaking change. Needs special notes in release notes!

### DIFF
--- a/src/kotlin_editor_export_plugin.cpp
+++ b/src/kotlin_editor_export_plugin.cpp
@@ -40,21 +40,55 @@ void KotlinEditorExportPlugin::_export_begin(const HashSet<String>& p_features, 
                 bool is_x64 {p_features.has("x86_64")};
 
                 if (!is_arm64 && !is_x64) {
-                    add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_DIRECTORY));
+                    add_macos_plugin_file(String(RES_DIRECTORY).path_join(HOST_EMBEDDED_JRE_DIRECTORY));
                 } else {
-                    if (is_arm64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_ARM_DIRECTORY)); }
-                    if (is_x64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_AMD_DIRECTORY)); }
+                    if (is_arm64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(MACOS_EMBEDDED_JRE_ARM_DIRECTORY)); }
+                    if (is_x64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(MACOS_EMBEDDED_JRE_AMD_DIRECTORY)); }
                 }
             } else {
-                String jre_dir {String(RES_DIRECTORY).path_join(EMBEDDED_JRE_AMD_DIRECTORY)};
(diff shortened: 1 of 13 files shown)


piiertho (dismissed):
I think we need to test in exports to be sure we don't break them.

chippmann:
I tested linux editor and export templates. Could you test windows and mac? (I don't think testing on mobile is necessary but you can if you want to :-) )

CedNaru (changes requested):

chippmann:
Tested the exports and editors for linux amd64 with the changes @CedNaru requested.
Will test on macos later today

chippmann:
Tested editor and export on macOS as well.

If you @CedNaru or @piiertho could test on windows, that would be appreciated.

chippmann:
Tested it on mac and windows as well in the meantime. Just graalvm is missing testing on those platforms

src/kotlin_editor_export_plugin.cpp
```diff
@@ -40,21 +40,55 @@ void KotlinEditorExportPlugin::_export_begin(const HashSet<String>& p_features,
...
                 if (!is_arm64 && !is_x64) {
-                    add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_DIRECTORY));
+                    add_macos_plugin_file(String(RES_DIRECTORY).path_join(HOST_EMBEDDED_JRE_DIRECTORY));
                 } else {
-                    if (is_arm64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_ARM_DIRECTORY)); }
-                    if (is_x64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(EMBEDDED_JRE_AMD_DIRECTORY)); }
+                    if (is_arm64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(MACOS_EMBEDDED_JRE_ARM_DIRECTORY)); }
+                    if (is_x64) { add_macos_plugin_file(String(RES_DIRECTORY).path_join(MACOS_EMBEDDED_JRE_AMD_DIRECTORY)); }
                 }
             } else {
```
CedNaru, line 48: We talked about it last time. We don't support all system and architecture that godot does, so we can't assume that the generic "else" works in all cases. We should make the if extra explicit and only leave a final else to raise an error, "Platform not supported"). We can also simplify its structure a bit compared to what it is now.

CedNaru (changes requested):

src/kotlin_editor_export_plugin.cpp
chippmann: replaced the generic else with a proper check for universal for macos and failure in all other cases
restructured the export a bit. it should now represent what we've discussed last time IIRC

CedNaru (approved):

piiertho:
I don't remember: did we test it on all platforms ?

chippmann:
I did on all platforms except mobile and graalvm for windows and macos
Although it might be worth to test again after my last change even though i did not change much about the logic of the export

piiertho:
I will test it on graalVM native-image for iOS.

piiertho:
Tested on iOS

piiertho (approved):

PR #635, opened 2024-05-14
Fix PathScript caching.

CedNaru:
We never properly cached the Kotlin and Java Scripts.
Each time you reload the .jar, new ones are created instead of updating the ones already here which triggers a nasty side effect.

The original Script is replaced inside the ResourceLoader by the new one. And its path taken as well.
The consequence is that if you got a Godot scene opened when reloading and then save, the script is no longer recognized as an external resource (because no path) and is stored as an internal script (which we don't support).

I also move the content of update_export to another method. The reason is that Godot calls that method whenever it detects a change in the source files, which is totally useless in our case because only the .jar matters.

This doesn't require testing on each OS, but I welcome if you can monkey proof the reloading behaviour with gdj and Kotlin scripts.

### DIFF
--- a/src/script/jvm_script_manager.cpp
+++ b/src/script/jvm_script_manager.cpp
@@ -1,122 +1,144 @@
 
 #include "jvm_script_manager.h"
 
+#include "lifecycle/paths.h"
 #include "script/language/gdj_script.h"
 
 void JvmScriptManager::create_and_update_scripts(Vector<KtClass*>& classes) {
-    LOG_DEV("Loading JVM Scripts...");
-
 #if defined(DEBUG_ENABLED) && !defined(TOOLS_ENABLED)
-    JVM_ERR_FAIL_COND_MSG(named_user_scripts.size() != 0, "JVM scripts are being initialized more than once.");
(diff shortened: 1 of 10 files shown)


chippmann (dismissed):

src/script/jvm_script_manager.cpp
```diff
@@ -1,122 +1,148 @@
...
+void JvmScriptManager::update_all_scripts() {
+    for (const Ref<NamedScript>& script : named_scripts) {
+        // We have to delay the update_export. The engine is not fully initialized and scripts can cause undefined behaviors.
+        JvmScript* ptr = script.ptr();
+        MessageQueue::get_singleton()->push_callable(callable_mp(ptr, &JvmScript::update_script));
     }
-    for (const Ref<PathScript>& script : path_user_scripts) {
-        script->update_exports();
+    for (const Ref<PathScript>& script : path_scripts) {
+        // We have to delay the update_export. The engine is not fully initialized and scripts can cause undefined behaviors.
```
piiertho, line 130: Comment should be changed as it is not `update_export` method used

piiertho (dismissed):
Only one comment otherwise that's LGTM

PR #637, opened 2024-05-16
Fix void method call

chippmann:
This fixes the onDestroy method call which caused a crash on android debug builds


chippmann:
Did not see #636

PR #641, opened 2024-05-25
Update to kotlin 2.0

chippmann:
This updates our project to Kotlin 2.0

I also took the opportunity to fix (new and existing) compiler warnings along the way and fix/add missing files in the various harness projects as well as update to the latest gradle version (8.7) on all gradle based projects so they are all on the same version again.

### DIFF
--- a/kt/godot-library/src/main/kotlin/godot/core/Variant.kt
+++ b/kt/godot-library/src/main/kotlin/godot/core/Variant.kt
@@ -745,7 +745,8 @@ enum class VariantType(
 
     internal val toGodot = { buffer: ByteBuffer, any: Any? ->
         if (any == null) {
-            NIL.toGodotWithoutNullCheck(buffer, Unit)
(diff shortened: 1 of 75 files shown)


kt/godot-library/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -745,7 +745,8 @@ enum class VariantType(
 
     internal val toGodot = { buffer: ByteBuffer, any: Any? ->
         if (any == null) {
-            NIL.toGodotWithoutNullCheck(buffer, Unit)
+            // TODO: replace with NIL.toGodotWithoutNullCheck(buffer, Unit) once https://youtrack.jetbrains.com/issue/KT-68339 is fixed!
```
piiertho, line 748: Can we create an issue for this one so we don't forget ?

piiertho (approved):

kt/godot-library/src/main/kotlin/godot/core/Variant.kt
chippmann: done: https://github.com/utopia-rise/godot-kotlin-jvm/issues/643

CedNaru (approved):

PR #649, opened 2024-06-03
fix function moveToward

datouzhu125:
The moveToward function in GDMath class under godot.global package has wrong behavior when it goes from 0 to negative value.
I query realized the Godot source code,as follows.
`static _ALWAYS_INLINE_ float move_toward(float p_from, float p_to, float p_delta) { return abs(p_to - p_from) <= p_delta ? p_to : p_from + SIGN(p_to - p_from) * p_delta; }`
The implementation in Kotlin should be as follows.
`fun moveToward(from: Float, to: Float, delta: Float) = if (abs(to - from) <= delta) to else from + (to - from).sign * delta`


CedNaru (approved):

CedNaru:
Thank you for the fix!

PR #657, opened 2024-08-08
feat: Implement KotlinCallableCustom to enable lambda callables

piiertho:
This adds kotlin lambda callables.  
We can now connect a signal directly to a kotlin lambda:  
```kotlin
signalWithParams.connect { p0, p1, p2 ->
    signalString = p0
    signalLong = p1
    signalNode = p2
}
```
You can also register a kotlin lambda as property like if it was a classic callable:  
```kotlin
    @RegisterProperty
    var ktCallable = { str: String -> ktCallableString = str }.asCallable()
```

This makes signal and callables generics generated.


kt/godot-library/src/main/kotlin/godot/core/bridge/NativeCallable.kt
```diff
@@ -0,0 +1,230 @@
...
+import godot.core.callable.KtCallable
+import godot.core.memory.MemoryManager
+import godot.core.memory.TransferContext
+import godot.util.VoidPtr
+import godot.util.camelToSnakeCase
+import kotlin.reflect.KCallable
+
+class NativeCallable : NativeCoreType, Callable {
+
+    constructor() {
```
CedNaru, line 16: I think all those constructors should be internal. The primary way of creating a Callable (Native or Kt) should be done through the invoke methods of the parent Callable so the users don't have to care about which type it is.

kt/godot-library/src/main/kotlin/godot/core/bridge/VariantArray.kt
```diff
@@ -207,15 +207,15 @@ class VariantArray<T> : NativeCoreType, MutableCollection<T> {
      *
      * **Note**: For an empty array, this method always returns `true`.
      */
-    fun all(callable: Callable): Boolean {
+    fun all(callable: NativeCallable): Boolean {
```
CedNaru, line 210: Should still be Callable here, otherwise users won't be able to use a Kotlin lambda alongside a VariantArray.

kt/godot-library/src/main/kotlin/godot/signals/Signals.kt
```diff
@@ -3,12 +3,10 @@ package godot.signals
...
 import godot.core.Callable
 import godot.core.CoreType
-import godot.core.GodotError
 import godot.core.StringName
 import godot.core.asStringName
 import godot.global.GD
 import godot.util.camelToSnakeCase
-import kotlin.reflect.KCallable
 
 open class Signal internal constructor(
```
CedNaru, line 11: File should be renamed Signal given that only the base class is left here.

kt/godot-library/src/main/kotlin/godot/signals/SignalProvider.kt
```diff
@@ -3,38 +3,6 @@ package godot.signals
 import godot.Object
 import kotlin.reflect.KProperty
 
-/**
- * For java convenience usage
- */
-object SignalProvider {
```
CedNaru, line 9: I don't see replacements for all those methods in the generated code. If I recall they exist because Java can't use the same syntax as Kotlin `val signal by signal1<Int>` so Cedric created those to help.

src/kt_variant.h
```diff
@@ -232,6 +232,20 @@ namespace ktvariant {
         return Variant(*to_native_core_type<T>(byte_buffer));
     }
 
+    static Variant from_kvariant_tKCallableValue(SharedBuffer* byte_buffer) {
+        bool is_custom {static_cast<bool>(decode_uint32(byte_buffer->get_cursor()))};
+        byte_buffer->increment_position(BOOL_SIZE);
+
+        if (is_custom) {
+            LOG_INFO("received custom callable");
```
CedNaru, line 240: You forgot to remove the logs

kt/godot-library/src/main/kotlin/godot/core/Variant.kt
```diff
@@ -81,6 +82,8 @@ internal val variantMapper = mutableMapOf(
     Vector4i::class to VECTOR4I,
     Projection::class to PROJECTION,
     Callable::class to CALLABLE,
```
CedNaru, line 84: Callable is now abstract so it can be removed.

CedNaru (changes requested):

chippmann (approved):
Some very minor nitpicks.

Nice work. I've yet to test it but as the tests checks are ok, you don't have to wait for that IMO

CedNaru:
Piertho and I talked about it. For now, we will just focus on making custom callables work + coroutines. Once we have something function with a satisfying API, we will go back to both to add properly disconnect/cancellation behavior.

Also note that even GDScript doesn't provide a way to disconnect callables made from GDScript lambda. Right now the only way is to query the list of callables subscribed to the signal and disconnect them one by one.

Still, we want to go above GDScript provides because we heavily rely on connecting/disconnecting a lot of callable in our project (which is still stuck on 3.x for now)

CedNaru (approved):

Issue #618, opened 2024-04-23
Proposal: Change the way the MemoryManager runs.


CedNaru:
**Reminder:**
The MemoryManager is our central database that holds a weak reference to all Godot related object created on the JVM side, being core types, wrappers for native types or even scripts. 
When created, all those objects are registered to a ReferenceQueue that allows us to know when they have been collected by the GC.
This callback is necessary so we can decrement the counter of RefCounted objects on the Godot side. That operation is currently done in a separate thread.

Whenever that thread is woken up by the OS's scheduler, it's checking the queue to see if there is any RefCounted instance to decrement. This design has 3 consequences:
- This thread can be active at random times and is totally at the mercy of the operating system. Not all schedulers are born equals, it can lead to quite the different behaviour depending on the platform.
- Objects are freed in big batches, with timings correlated to the GC (When the GC runs + the time for the thread to wake up). The JVM and GC might be optimized for this kind of purpose, but Godot is not. We run the risk of interfering with the main thread because of the multiple locks necessary to manage RefCounted (our owns and the one already in the native Object).
- All Refcounted instances with a JVM binding can no longer be freed by Godot itself. Because our binding is permanently increasing the counter by 1, MemoryManager always has the last word. We are bypassing how Godot normally operates with what I consider to be a hacky solution.

**My proposal holds in 2 points:**
- Make the MemoryManager run in the main thread. The Language class has a `frame()` method. Like the name suggests, it's called at the end of every frame once the Scene (and so scripts) have been fully processed, which also include the secondary threads running in the new multithreaded scene system. Instead of relying on a JVM thread calling C++ once for every collected reference. We can instead have the module directly query the memory manager for the instances that have been GCs during that frame (if any) and get all the relevant data in a single JNI call (made even faster now that we can convert a whole Kotlin collection in a single call). It technically means more work for the main thread, but the pacing will be stable (at least as stable as the game itself) and more granular. This will greatly reduce the amount of data race, the call happening after the scene system ran. It won't stop all of them, because Godot also has its ThreadPool and users can create their own threads as well. But it's safe to say that the scene threads are the ones that will interact the most with the Godot Script system anyway, and so with our module.
[...]

Issue #508, opened 2023-10-11
Remove GodotStatic, Singleton and Leak management from MemoryManager.


CedNaru:
Previously Godot didn't have any way to remove existing instance_bindings from Object, leading to some additional code in the MemoryManager to handle singletons, objects stored in static instances and potential leaks as cyclical references between C++ and Kotlin were possible.

Godot 4.2 introduces a new method `Object::free_binding_instance`. With that method, we can easily unbind any C++ object and Kotlin wrapper remaining in the memory manager when the JVM is closing.
The singletons needed to be a special case, as the native objects were freed by Godot only after the JVM module was finished, leaving their references in the memory managers.
Refcounted references stored in static objects will no longer be stuck in a cyclical reference either, with the instance binding removed, and regular Object are supposed to be manually managed by design anyway, so it was kind of a bad practice to give them an autofreed feature.

With that out of the way, there is no longer a need to keep track of potential leaks on the JVM side. If leaks remain, the user is going to be notified by the regular Godot ObjectDB anyway and the potential causes would just be for regular Godot reasons not related to the JVM (like not freeing an Object).

Let's take the 4 potential cases:

1. Object not leaking in both C++ and JVM: Everything is fine
2. Object leaking in both C++ and JVM: The C++ side already has the responsibility of reporting those leaks, there is no need for the JVM to do the same work a second time.
3. Object leaking in C++ but not in the JVM: Not our job.
4. Object leaking in the JVM but not in C++: In such a case, it means that the native object was already freed but that the reference is still alive in the JVM. By design this case is impossible for Refcounted instances because if a JVM instance is still alive, then the C++ side is as well, and we are in case 2. The only possibility left then are regular Objects that have been manually freed earlier in the code but somehow still used in the Kotlin code. I think this last case is not our responsability either. Manual Object management has to be properly done in Godot regardless of the script language, this is up to the user to design a code that won't make use of any dead pointer. We also don't need to report on that, because if a dead instance was improperly used in the Kotlin code, then the user would have received an error from Godot way before the game closing.

Taking everything into account and with the new memory design allowed by 4.2. We have no reason to check for leaks on the Kotlin side (and currently making the game crash when it happens instead of just a warning like Godot does) so I would just remove that part.

Can anyone think of a possible case I didn't mention ?

PR #661, opened 2024-08-17
MemoryManager rework

CedNaru:
Solve #508, #618, #566, #666 and #617 

It's honestly hard to explain in details all the changes without turning it into an essay. Most of the changes and explanations were already written in #618.

Here's a recap of the main changes:
- The MemoryManager is no longer its own thread, instead it's called directly by the C++ code at the end of each frame.
- Bindings no longer keep JNI references to their JVM wrappers. Objects are directly stored as strong references in the JVM ObjectDB and RefCounted as weak references. 
- The "binding phase" that was made in a delayed manner is now gone. Instead, it's replaced by a "dead object" phase. Each frame, the list of dead objects is sent to the JVM to remove them from the ObjectDB.
- Refcounted wrappers are no longer permanent, they die as soon as they are no longer used by the JVM, decrementing the counter in the process. The JVM is no longer the one with the final words on RefCounted wrappers' life.
- The only exception are RefCounted Scripts. The ping pong of weak/strong JNI ref was kept but moved from KoltinBinding to JvmInstance. The ping pong is also less frequent. References are transformed into strong ones immediately, but the switch to weak is delayed to the end of the frame. 
- ForceGC was removed, it's not a reliable method to make sure the GC ran.
- DisplayLeaks was removed, Godot should be the only one responsible for such reports.
- Singleton management has been removed because no longer relevant in the new model.
- Static management has been replaced by a broader "cleanup callback" system. It can be used to close third party threads, for example.


kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
```diff
@@ -293,95 +199,44 @@ internal object MemoryManager {
...
-                break
+    fun checkCleanUp(): Boolean {
+        if (ObjectDB.any { it != null } && nativeCoreTypeMap.isNotEmpty()) {
+            // Still work to do so we force the gc
+            var any: Any? = Any()
+            val wkRef = WeakReference(any)
+            @Suppress("UNUSED_VALUE")
+            any = null
+            while (wkRef.get() != null) {
+                System.gc()
```
chippmann, line 221: Didn't we want to get rid of this?

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/GenerationService.kt
```diff
@@ -754,14 +754,10 @@ class GenerationService(
             FunSpec.builder("new")
                 .addModifiers(KModifier.OVERRIDE)
                 .addParameter("scriptIndex", Int::class)
-                .returns(Boolean::class)
```
chippmann, line 757: Please explicitly return `Unit` in these cases. It will be required once we upgrade kotlin poet

src/language/gdj_language.cpp
```diff
@@ -35,6 +36,13 @@ void GdjLanguage::init() {
 #endif
 }
 
+void GdjLanguage::frame() {
+    if (!GDKotlin::get_instance().user_configuration.disable_gc) {
```
chippmann, line 40: wouldn't an unlikely here be beneficial?

src/script/jvm_instance.cpp
```diff
@@ -151,10 +161,52 @@ String JvmInstance::to_string(bool* r_valid) {
     return ScriptInstance::to_string(r_valid);
 }
 
-void JvmInstance::refcount_incremented() {}
+void JvmInstance::refcount_incremented() {
+    // This function should only be called when we know the object is a RefCounted. We directly reinterpret the pointer to it
+    RefCounted* ref = reinterpret_cast<RefCounted*>(owner);
+    int refcount = ref->get_reference_count();
+
+    if (refcount == 2) {
```
chippmann, line 169: shouldn't this be a >=? then?

chippmann (changes requested):
Very nice work!
Will follow up with proper testing later this week

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
CedNaru: I'm keeping it for the debugging phase right now. Depending on how it goes, I might have to keep it, which I would prefer not.

src/language/gdj_language.cpp
CedNaru: Not really, it's meant to be called once per frame, it has no performance implications

src/script/jvm_instance.cpp
CedNaru: I switched to 2 because I want the function to only be called when we increment from 1 to 2. Bumping it to higher numbers is irrelevant. The pingpong only happens in the 1 <=> 2 transitions.

kt/godot-library/src/main/kotlin/godot/core/KtObject.kt
```diff
@@ -95,10 +95,14 @@ abstract class KtObject {
     protected fun <T : KtObject> T.godotNotification(block: T.(Int) -> Unit): GodotNotification = GodotNotification(block as Any.(Int) -> Unit)
 
     @Suppress("FunctionName")
+    /**
+     * Called automatically when the Object is destroyed. Not that this method is not available for RefCounted or any of its child class.
```
chippmann, line 99: Typo: Not -> Note

kt/godot-library/src/main/kotlin/godot/global/GD.kt
```diff
@@ -78,6 +78,14 @@ object GD : GDMath, GDCore, GDRandom, GDPrint {
         return ResourceLoader.loadAs(path, typeHint, cacheMode)
     }
 
+    /**
+     * The MemoryManager runs automatically at the end of each frame, but it can happen that the engine freeze for a long period of time (a loading screen).
+     * You can use that method to manually trigger it if you want memory to be quickly freed.
+     * */
+    fun syncMemory() {
```
chippmann, line 85: why not manage memory like in the `MemoryManager`? IMHO this is more fitting

src/jvm_wrapper/memory/memory_manager.cpp
```diff
@@ -88,6 +90,53 @@ bool MemoryManager::unref_native_core_type(JNIEnv* p_raw_env, jobject p_instance
     return has_free;
 }
 
+void MemoryManager::manage_memory(JNIEnv* p_raw_env, jobject p_instance) {
+    jni::Env env {p_raw_env};
+    MemoryManager::get_instance().sync_memory(env);
```
chippmann, line 95: Shouldn't we align the method names?

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
```diff
@@ -6,68 +8,48 @@ import godot.core.NodePath
...
-    /** Queue of objects that need to be bound to native objects*/
-    private val bindingQueue = ArrayDeque<GodotBinding>()
-
-    /**
-     * Queue of objects which will to be bound to native objects in a given iteration.
-     *
-     * At the start of a garbage collection iteration. The objects from the [bindingQueue] are copied into this list in order to spend as little time as possible in a `synchonized` block. Later we loop through this list to actually bind the objects.
-     */
-    private val bindingList = ArrayList<GodotBinding>(CHECK_NUMBER)
+    private val ObjectDB = Array<GodotNativeEntry?>(OBJECTDB_SIZE) { null }
```
piiertho, line 34: Can be changed to `arrayOfNulls`

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
```diff
@@ -6,68 +8,48 @@ import godot.core.NodePath
...
     /** Queues so we are notified when the GC runs on References.*/
     private val refReferenceQueue = ReferenceQueue<GodotBinding>()
 
-    /** List of element to remove from ObjectDB*/
-    private val deleteList = ArrayList<GodotWeakReference>(CHECK_NUMBER)
+    /** List of references to decrement.*/
+    private val deadReferences = mutableListOf<GodotRefCountedEntry>()
+
+    /** Pointers to NativeCoreType.*/
+    private val nativeCoreTypeMap = ConcurrentHashMap<VoidPtr, NativeCoreWeakReference>(INITIAL_CAPACITY)
```
piiertho, line 43: Can be changed to `concurrentMapOf`

register_types.cpp
```diff
@@ -70,7 +70,7 @@ void uninitialize_kotlin_jvm_module(ModuleInitializationLevel p_level) {
     if (Engine::get_singleton()->is_project_manager_hint()) { return; }
 #endif
 
-    if (p_level != MODULE_INITIALIZATION_LEVEL_SERVERS) { return; }
+    if (p_level != MODULE_INITIALIZATION_LEVEL_SCENE) { return; }
```
piiertho, line 73: Why this has changed ?

src/jvm_wrapper/memory/memory_manager.cpp
```diff
@@ -1,27 +1,19 @@
...
-    KotlinBindingManager::bind_object(id, j_object);
-}
-
-void MemoryManager::unbind_instance(JNIEnv* p_raw_env, jobject p_instance, jlong instance_id) {
-    ObjectID id {static_cast<uint64_t>(instance_id)};
-    KotlinBindingManager::unbind_object(id);
-}
-
 void MemoryManager::decrement_ref_counter(JNIEnv* p_raw_env, jobject p_instance, jlong instance_id) {
     Object* obj = ObjectDB::get_instance(static_cast<ObjectID>(static_cast<uint64_t>(instance_id)));
```
piiertho, line 15: Simplification:
```cpp
if (Object* obj { ObjectDB::get_instance(static_cast<ObjectID>(static_cast<uint64_t>(instance_id))) }) { 
    KotlinBindingManager::decrement_counter(reinterpret_cast<RefCounted*>(obj));
}
```

register_types.cpp
CedNaru: It was a mistake, the init was set to the scene level but the uninit was set to server. People had bug when closing the game because the destructor of some Godot objects needs singletons to be available, which is not guaranted if you do it later than the Scene level.

src/jvm_wrapper/memory/memory_manager.cpp
CedNaru: It's a mistake, I added the if during debug but I don't even need it. I know the RefCounted can't be dead before I decrement it.

chippmann (approved):
https://godot-kotl.in/en/stable/contribution/knowledge-base/memory-management/



Should probably also be updated

chippmann (approved):

kt/godot-library/src/main/kotlin/godot/core/memory/MemoryManager.kt
piiertho: Seems not std and cannot specify initial capacity :/

piiertho (approved):

PR #664, opened 2024-08-21
Feature/implement new core types 4.3

### LINKED ISSUE #646: Add new core type PackedVector4Array
Godot 4.3 is going to add a new type to the core family:
https://github.com/godotengine/godot/pull/85474

We should implement it and use the opportunity to simplify the code for all packedArray by creating a common template for them. Aside of PackedByteArray, they all share the exact same API

piiertho:
This implements `PackedVector4Array`.  
This simplify packed arrays code using template and generics.  
This needs test on android and graal.

Edit by @chippmann : Closes #646


src/jvm_wrapper/jvm_instance_wrapper.h
```diff
@@ -34,6 +34,25 @@ public:
                                                                                      \
 private:
 
+#define INIT_JNI_BINDINGS_TEMPLATE(...)                                                       \
```
CedNaru, line 37: The only difference I see with the regular macro is that the fqname became a property instead of a method. Is there a reason why we can't just apply this change to the regular macro instead of just having a near copy of it ?

piiertho: Yes, it is because `get_fully_qualified_name` is not accessible when using a template, it would require to call it this way: `MyTemplate<T>::get_fully_qualified_name()`, which would force to repeat class name in macro

CedNaru (approved):

CedNaru (dismissed):

src/jvm_wrapper/memory/kt_binding.macos.editor.arm64-a1cb275f.o.tmp
chippmann, line 1: why is this here?

src/jvm_wrapper/memory/long_string_queue.macos.editor.arm64-88fb0a13.o.tmp
chippmann, line 1: and this one?

src/jvm_wrapper/bridge/variant_array_bridge.macos.editor.arm64-31425541.o.tmp
chippmann, line 1: same

src/lifecycle/jvm_manager.cpp
```diff
@@ -94,6 +96,8 @@ bool JvmManager::initialize_or_get_jvm(void* lib_handle, JvmUserConfiguration& u
     return true;
 }
 
+inline static constexpr const char packed_color_array_fqdn[] = "godot.core.PackedColorArray$Bridge";
```
chippmann, line 99: The other ones are named directly in the ctor. why this one not?

src/jvm_wrapper/bridge/packed_int_64_array_bridge.macos.editor.arm64-cd375222.o.tmp
chippmann, line 1: some more. i leave it at this :-)

chippmann (changes requested):

piiertho:
Tested for graalvm native-image on iOS.

CedNaru (changes requested):
In `memory_manager.cpp`, you have to update the `unref_native_core_type` method to handle the new coretype.

chippmann (approved):

CedNaru (approved):

PR #668, opened 2024-08-29
Limit core type usage on registered properties

chippmann:
This limits the usage of core types on exported properties to actually allowed cases.

That means that registered core type properties can no longer be lateinit!

Edit: This now also fixes some type checks which failed because callables did not implement the core type interface


CedNaru:
I said it was okay for them to be nullable. In such a case, Godot Variant is going to interpret the null as the "zero value" of the corresponding core type. Unlike lateinit, nullables are not the cause of any known bug right now. 
Maybe we should not allow them, as it's more rigorous but in this case, you also have to check the return type of registered methods, not just properties.

chippmann:
Yeah it was just easy going to directly limit them.
I mean i can also limit the functions right now. Not much work.
But we can also allow them. Both are ok for me.

CedNaru (approved):

CedNaru (approved):

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
```diff
@@ -78,7 +78,7 @@ class Invocation : Node3D() {
 
 	@Export
 	@RegisterProperty
-	lateinit var lateinitString: String
+	var lateinitString: String? = "null"
```
piiertho, line 81: Shouldn't it be `= null` instead of `= "null"` to test nullability ?

kt/godot-library/src/main/kotlin/godot/core/bridge/Callable.kt
```diff
@@ -4,7 +4,7 @@ package godot.core
 import godot.Object
 import godot.core.callable.KtCallable
 
-interface Callable {
+interface Callable: CoreType {
```
piiertho, line 7: So I assume now Callables cannot be lateinit ?

chippmann: no callables are not allowed to be lateinit if registered

harness/tests/src/main/kotlin/godot/tests/Invocation.kt
chippmann: yeah was a mistake from testing

CedNaru (approved):

piiertho (approved):

PR #671, opened 2024-08-30
Don't fetch default value for Object properties

CedNaru:
C++ counterpart to the lateinit change #668. The editor won't try to get the default value for Object anymore.
Object is the only allowed lateinit, but Godot doesn't handle default values for that type anyway.

Here a simple proof using GDScript, a property is declared with a default value, but the inspector can't see it:
image

### DIFF
--- a/src/jvm_wrapper/registration/kt_property.cpp
+++ b/src/jvm_wrapper/registration/kt_property.cpp
@@ -67,17 +67,3 @@ void KtProperty::call_set(jni::Env& p_env, KtObject* instance, const Variant& p_
     jvalue args[1] = {jni::to_jni_arg(instance->get_wrapped())};
     wrapped.call_void_method(p_env, CALL_SET, args);
 }
-
-#ifdef TOOLS_ENABLED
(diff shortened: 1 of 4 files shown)


chippmann (approved):

src/jvm_wrapper/registration/kt_property.cpp
```diff
@@ -67,17 +67,3 @@ void KtProperty::call_set(jni::Env& p_env, KtObject* instance, const Variant& p_
     jvalue args[1] = {jni::to_jni_arg(instance->get_wrapped())};
     wrapped.call_void_method(p_env, CALL_SET, args);
 }
-
-#ifdef TOOLS_ENABLED
-void KtProperty::safe_call_get(jni::Env& p_env, KtObject* instance, Variant& r_ret) {
```
piiertho, line 72: why method is removed from cpp but not from header ?

CedNaru: mistake

piiertho (approved):

chippmann (approved):

PR #675, opened 2024-08-31
Fix default value retrieval for resources and node paths

chippmann:
Resources should still gather default values.
For node types (which are treated as node paths in the inspector) we return an empty variant so we can set them.

@CedNaru please look at this carefully. I'm very unsure if this is the proper way of dealing with this, but it works


piiertho (dismissed):

piiertho (approved):

chippmann:
@CedNaru as i opened this PR i cannot approve it. But i approve :-)

CedNaru (approved):

PR #679, opened 2024-08-31
Add tests on exported projects to our CI/CD pipeline

chippmann:
This adds tests on exported projects to our ci/cd pipeline. This addition has two main benefits:
- We automatically test exports
- We can detect issues which would only arise on exported projects

Atm only tests on linux, macos, and windows are added as this represents our current test suite.

GraalVM for all three platforms will follow in a separate PR.

Once we add android (and maybe even iOS) tests, we could create export tests for them as well. But... baby steps :-)


.github/workflows/test_linux_exports.yml
```diff
@@ -0,0 +1,101 @@
...
+      matrix:
+        include:
+          - name: dev tests
+            target: dev
+            bootstrap-target: debug
+          - name: release tests
+            target: release
+            bootstrap-target: release
+
+    steps:
```
CedNaru, line 28: Do we even need to clone Godot here ? Everything is already built after all.

piiertho (dismissed):

.github/workflows/test_linux_exports.yml
CedNaru: The sources for the api gen ? Previously we needed Godot to get the xml documentation, but now even that is contained inside the api.json, a file that we directly commit to this repo. Piertho made the change, we no longer use the xml at all, it's all purely from the json.

chippmann: done

CedNaru (approved):

PR #680, opened 2024-08-31
Fix closing of module

CedNaru:
The issue was caused by some Godot objects with a Kotlin binding being destroyed later than the module. Our binding callbacks needs to access one of the singleton of our module, which was already destroyed at this point.
I just slightly changed the cleanup operation to also include Object and not just decrement RefCounted.


piiertho (approved):

chippmann (approved):
But please change the title of the PR before merging as this will be the commit message ;-)

PR #682, opened 2024-09-01
Optimize memory manager

CedNaru:
- ThreadLocalLazy had no real purpose. A thread local is already a lazy object.
- Removed GodotBinding, it's no longer necessary now that JNI references to it are gone.
- Reused the name GodotBinding for the entries of ObjectDB and CoreMap.
- Script and Wrapper directly reference each other now.
- Syncing memory is smoother over time.
- Simplified KtObject constructor.
- CoreType are now batched.
- Updated the knowledge doc.

I also replaced the ObjectDB Array by a HashMap. It's technically less efficient, but we were wasting a large amount of memory.
The size of the objectDB is 2^24 references that have to be preallocated, which are 4 Bytes each on the JVM, this takes 64MB alone. On top of that, most of that array will be empty anyway given how sparse Kotlin wrappers are compared to the total number of Godot objects. Memory space is limited on Android so we shouldn't waste it.
It shouldn't have any meaningful performance impact. If it becomes a bottleneck later, we can still create a custom hashmap with a straightforward hashing (Godot C++ map doesn't even hash when the key is a pointer, it just simply uses the memory address as a hash directly).


docs/src/doc/user-guide/advanced/cleanup-operations.md
```diff
@@ -12,7 +12,7 @@ fun foo() {
 ```
 
 !!! warning
-Be mindful that this operation happens when Godot has already been partially closed. 
+    Be mindful that this operation happens when Godot has already been partially closed. 
 The SceneTree is no longer present and a part of the Godot API has been unregistered.
 The order of execution is not guaranteed. Make sure that those callbacks don't depend on each other.
```
chippmann, line 17: These two lines also need to be indented

kt/godot-library/src/main/kotlin/godot/core/memory/GodotBinding.kt
```diff
@@ -0,0 +1,41 @@
...
+}
+
+internal class RefCountedBinding(
+    instance: KtObject,
+    queue: ReferenceQueue<KtObject>,
+    override val objectID: ObjectID
+) : WeakReference<KtObject>(instance, queue), GodotBinding {
+
+    override val instance
+        get() = get()
```
chippmann, line 34: Please explicitly write the type and use `get() = thi.get()`
this hurts my eyes just looking at `RefCountedBinding` alone ^^

```suggestion
    override val instance: KtObject?
        get() = this.get()
```

chippmann (approved):
Very nice work. Some NIT's nothing dealbreaking

chippmann (approved):

piiertho (approved):

PR #684, opened 2024-09-03
Remove nullables from Variant::toKotlin

CedNaru:
I started just wanting to fix an error when we send from Godot to Kotlin a signal.
Signal contains an object and a string name. The buffer was actually writing the variantType 2 times, one for the signal and one contained in the append_object method.
I had to get this VariantType out of the method, but there was also the issue of writing a Nil variant if the pointer was null.
One from one change triggering another, I ended up removing the whole nullable system on the Kotlin side.
Now, Variant::NIL returns a null directly and Variant::OBJECT can also return a null if the ptr is 0.
The rest of the type are non nullable.

The main consequence is that we no longer have to write those Array<Boolean> everywhere.


src/jvm_wrapper/registration/kt_property.h
```diff
@@ -54,14 +54,12 @@ JVM_INSTANCE_WRAPPER(KtProperty, "godot.core.KtProperty") {
 
     INIT_JNI_BINDINGS(
         INIT_JNI_METHOD(GET_KT_PROPERTY_INFO, "getKtPropertyInfo", "()Lgodot/core/KtPropertyInfo;")
-        INIT_JNI_METHOD(IS_REF, "isRef", "()Z")
```
piiertho, line 57: This method should also be removed from graal jni config.

src/script/jvm_script.cpp
```diff
@@ -235,7 +235,9 @@ void JvmScript::update_script() {
         const String& property_name {exported_property.name};
 
         if(exported_property.type != Variant::OBJECT) {
+            LOG_DEV_VERBOSE(vformat("Get default value for %s property from %s:",exported_property.name, kotlin_class->registered_class_name));
```
piiertho, line 238: Are those logs wanted ?

piiertho (approved):

src/jvm_wrapper/registration/kt_property.h
CedNaru: Done

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/GenerationService.kt
```diff
@@ -881,10 +881,9 @@ class GenerationService(
                 )
             } else {
                 addStatement(
-                    "return·(%T.readReturnValue(%T, %L)·as·%T)${callable.getFromBufferCastingMethod()}",
+                    "return·(%T.readReturnValue(%T)·as·%T)${callable.getFromBufferCastingMethod()}",
```
chippmann, line 884: The braces are not needed anymore

kt/godot-library/src/main/kotlin/godot/util/Nullable.kt
```diff
@@ -0,0 +1,3 @@
+package godot.util
+
+inline fun <reified T> isNullable(): Boolean = null is T
```
chippmann, line 3: Not sure if we should expose this utility method. I'd either make it internal or if it is used in another module, I'd move that into the tools-common module

chippmann (changes requested):
And as discussed on discord the default property usage flag :-)

Great work!

PR #690, opened 2024-09-15
Reenable Editor Plugin.

CedNaru:
The editor plugin has been disabled since Godot 4 because we didn't have the time to go over it and adapt it to Godot 4.
 On top of that, the code was made extremely early when we weren't used to Godot's internal (we only knew the GDNative API before) so we had a lot of aberrations in it.
 
 I removed most of it to keep only the core features:
 - Add a button to build the JVM project using gradle. The result of the build is then displayed in a window.
 - Automatically build the JVM project when the game is about to be run. It displays a window only in case of build error. 
 - "About" Dialog, that can optionally be displayed the first time a user open the editor to warn them about the module.

As a new feature, you can also generate the JVM/Gradle project from the editor now.
image
image

For that purpose, I directly encoded the files into a header using a python script. The script has been saved so we can use it whenever the template files need to be updated.  Note that there is a new version.h to update, it will automatically replace the version used in the template.


docs/src/doc/getting-started/setting-up.md
```diff
@@ -6,6 +6,17 @@ or [manually](#setting-up-manually).
...
+
+You can simply create a regular new Godot project. 
+Once done, you can go to `Project/Tools/Kotlin/JVM/Generate JVM Project`.
+![Godot menu](../assets/img/editor-plugin/generation_menu.png)
+
+The following choice will appear:
+![Project dialog](../assets/img/editor-plugin/generation_choice.png)
+
+After this action, all required files should be generated, and you can safely build the project using the `Build` button at the top-right of the editor window.
+
```
chippmann, line 19: I'd add that section after the Setting up using Intellij IDEA project wizard or befor the line above.
As the line above suggests the next section rather than this one IMO

src/editor/template/.gitignore.template
```diff
@@ -0,0 +1,7 @@
+.godot/
+/android/
+.gradle
+.kotlin
+.idea
+build
+jre
```
chippmann, line 7: TBH i would move the templates to the build tools module and use that one from the IDE plugin and use path ref here (as you would execute the python script when our project is checked out anyways). or directly do a path ref to the template files in the IDE plugin project.

Now we have a duplication of the template files and we need to remember to update both

chippmann (changes requested):

chippmann:
As a question; could that pythong script not be executed via scons as part of the build?

docs/src/doc/getting-started/setting-up.md
CedNaru: I'll move it before, I think it's more natural to start with what is directly withing Godot.

src/editor/template/.gitignore.template
CedNaru: I can't do that as it is
The Intellij templates are for the WHOLE godot project, not just the additional JVM files,  and they are all in a flat directory.
For theGodot editor plugin I just need a subset of those and on top of that the python script work recursively so I just set the template files in the same relative directories as the final output.
The intellij plugin doesn't work that way so as it is right now, it's not really compatible. 
We can unify it in a separate PR.

CedNaru:
Added template generation directly when building the engine.

src/editor/project/templates.h
```diff
@@ -0,0 +1,41 @@
+// Auto-generated templates from src/editor/template directory
```
piiertho, line 1: This should be in TOOLS_ENABLED only

piiertho (changes requested):

src/editor/build/build_manager.cpp
```diff
@@ -4,85 +4,17 @@
 #include "build_manager.h"
```
Sombombus, line 4: > ``

chippmann (changes requested):
After creating the files from within the godot editor plugin, the gradlew file still needs to be made executable.
We should do this after we created the files

piiertho (approved):

chippmann (approved):

piiertho (approved):

PR #691, opened 2024-09-16
Improve JNI exception handling and logging

CedNaru:
- Added Typed `MethodID` and made the `JObject's` method calls use them for type safety. Updated all `JvmWrapper` jni methods.
- Logging methods have been rewamped to allow formatting without` vformat(`).
- Crash methods are gone, the only thing left that can explicitly crash the module are the `DEV_ASSERT` statements. As we progress toward a stable release, we already removed most of the crashes. I simply finished the last remaining ones. The new `DEV_ASSERT` statements are mostly related to JNI errors, which is not the type users should have to care about.
- `LOG_ERROR_WITH_ALERT` is also gone. Since Godot 4.3, errors are automatically displayed in the editor as small popups. It's not necessary to spawn our own alert window anymore.
- A new way to handle exceptions. The Env's methods are now much simpler, without difference in behavior between DEBUG and release. You can now set a custom "exception handler" that will be called instead of `exceptionDescribe()` when set. This handler is now set to be the method of a new singleton wrapper that takes a `Throwable ` instance and print its stacktrace directly.

This is what it looks like when a JVM exception is handled from the C++ side with the new handler:
image


kt/godot-library/src/main/kotlin/godot/core/bridge/ExceptionHandler.kt
```diff
@@ -0,0 +1,12 @@
+@file:Suppress("PackageDirectoryMismatch")
+
+package godot.core
+
+object JvmStackTrace {
+    fun getExceptionStackTrace(throwable: Throwable): String {
```
piiertho, line 6: Isn't there any method in kotlin to do this ?

piiertho (dismissed):

kt/godot-library/src/main/kotlin/godot/core/Properties.kt
```diff
@@ -30,12 +31,21 @@ open class KtProperty<T : KtObject, P : Any?>(
     protected val variantConverter: VariantConverter,
 ) {
     open fun callGet(instance: T) {
-        TransferContext.writeReturnValue(kProperty.get(instance), variantConverter)
+        try {
+            TransferContext.writeReturnValue(kProperty.get(instance), variantConverter)
+        } catch (t: Throwable) {
+            GD.printErr("Error calling a JVM getter from Godot:", t.stackTraceToString())
```
chippmann, line 37: Could we also print which instance it is and which property we tried to get?

kt/godot-library/src/main/kotlin/godot/core/Properties.kt
```diff
@@ -30,12 +31,21 @@ open class KtProperty<T : KtObject, P : Any?>(
...
+        }
     }
 
     open fun callSet(instance: T) {
         val arg = extractSetterArgument<P>()
-        kProperty.set(instance, arg)
+        try {
+            kProperty.set(instance, arg)
+        } catch (t: Throwable) {
+            GD.printErr("Error calling a JVM setter from Godot:", t.stackTraceToString())
```
chippmann, line 47: Same here?

chippmann (changes requested):

kt/godot-library/src/main/kotlin/godot/core/Properties.kt
CedNaru: Yeah good point. I just copied-pasted it from the method calls. I forgot to try catch the properties last time I changed the exception checking.
I'll also add more info for method calls.

chippmann (approved):

Issue #694, opened 2024-09-21
Proposal: Reduce Signal memory usage


CedNaru:
Signals are implemented this way in Kotlin:

```kt 
open class Signal internal constructor(
    val godotObject: Object,
    val name: StringName
)
```

And if you want to create one for your script, or even if we want to generate one for wrappers, we use the following delegate:

```kt 
// How to declare a signal. Example taken from Object.kt
public val scriptChanged: Signal0 by signal()


// What happens behind the hood
class SignalDelegate<T : Signal>(val factory: () -> T) {
    @PublishedApi
    internal var signal: T? = null
    @Suppress("NOTHING_TO_INLINE")
    inline operator fun getValue(thisRef: Object, property: KProperty<*>): T {
        if (signal == null) {
            signal = factory()
        }
        return signal!!
    }
}

```

So let's analyse this. By first decompiling the result:

```java

   // $FF: synthetic field
   static final KProperty[] $$delegatedProperties;
   @NotNull
   private final SignalDelegate scriptChanged$delegate;
   ```
We can ignore the `static KProperty[]` as it doesn't cost any memory per instance.
On the other side we have a SignalDelegate instance, one for each signal, for each instance of the class.
This delegate contains a lazy Signal instance, itself containing a reference to the signal owner + a StringName of the property name.

It means that in the best case, each instance is going to have a SignalDelegate allocated and stored per signal + a reference to itself and a String name if that signal is used.

Godot classes can have a lot of signals, which takes a lot of additional memory. Ideally, a simple wrapper should only cost in memory the pointer and id of the native object.

Let's simply count the number of signals of a basic Node2D:
Inherited from Object: 2
Inherited from Node: 10
Inherited from CanvasItem: 4
Own: 0

Total: 16 signals.  

We need to allocate that many Delegate + possible Signal and StringName instances for every Node2D and child class (script or wrapper) used in the code. In my opinion, this is not acceptable for simple wrappers.
A base Object without any field is 16 Bytes on the JVM and a reference is 8 bytes. The delegate having one field to store the signal, its size would be 24 bytes (+ 8 to store a reference to it in the class).

It means that each signal delegate is using 32 bytes of memory (16 * 32 = 512 bytes taken just for signals in a Node2D).

 On top of that, signals are rarely used relatively to their numbers. Most of the time, you are only going to use them once to connect them to a Callable.

[...]

Issue #606, opened 2024-04-10
A proposal for reducing the need for @Register annotations


MartinHaeusler:
# Introduction and motivation

I've been working with the kotlin API quite a lot in the last couple of weeks. While I'm very happy with it in general, one of the papercuts is the `@Register` family of annotations (`@RegisterClass`, `@RegisterFunction`  and `@RegisterProperty`).

Imagine the following scenario:

```kotlin
@RegisterClass
class MyButton : Button(){

    override fun _onReady(){
        GD.print("hello world!")
    }

}
```

This will never print anything. Why? Because `_onReady` needs to be annotated with `@RegisterFunction`. I've lost count of how many times I was pulling my hair why some action wouldn't trigger in my game, and **in 90% of all cases it was because I was missing an annotation**. The information is also redundant: we're overriding a function of the Godot API. Of course we want Godot to "see" it, why wouldn't we?

To reduce this pain, and just get rid of a pitfall, I propose the following set of rules where the Godot gradle plugin should automatically register classes, functions and properties, no matter if they're annotated or not.

# Rules for auto-registration

1. Every `class` which directly or indirectly inherits from `godot.Node` (and is not `abstract`) should be treated as if it was annotated by `@RegisterClass`. 
> It prevents anyone from forgetting it, and there is no use case for a subclass of `godot.Node` that isn't usable in the editor. Conversely, with this rule, I would argue that the explicit `@RegisterClass` annotation can be deprecated and marked for removal. It is always a mistake to annotate a class with `@RegisterClass` if it doesn't inherit directly or indirectly from `godot.Node` as Godot will not know what to do with it.

2. Every function which is an `override` of a function of `godot.Node` or its subclasses in the `godot` namespace should be treated as if it was annotated by `@RegisterFunction`. 

> For example, `override fun _onReady(){}` should be enough, there's no need for `@RegisterFunction`. There is no use case for overriding a Godot-defined API that shouldn't be visible to Godot.

3. Every property which is annotated as `@Export` should also be treated as if it was annotated by `@RegisterProperty`. 

> There is no use case where we want to export a property to the Godot editor and hide it at the same time. The inverse however is not true: a property which is marked as `@RegisterProperty` should not necessarily show up in the editor, which justifies the distinction between `@Export` and `@RegisterProperty`.

# Implementation considerations

[...]

PR #695, opened 2024-09-22
Make signal delegate static to avoid memory allocation

CedNaru:
Implement #694 

I mostly modified the Signal generation so it can create its own delegate. I used the opportunity to split generation into smaller functions. If you want to see the truly new stuff, look at the `SignalGenerationService::generateSignalDelegateCompanion`.

The api generation for signals went from this:
```kt
  public val childEnteredTree: Signal1<Node> by signal("node")
```
to 
```kt
  public val childEnteredTree: Signal1<Node> by Signal1
```

The reason for the change is that the previous method was actually creating a field in the class for each signal (as explained in the linked proposal), you can only remove that field if the delegation is done on something known at compile time. This sadly doesn't include inline functions. So even if I transform the `signal()` function to
```kt
  inline fun <P0> signal1(p0: String) = Signal1
```
It won't work (I tried, and the bytecode generated still has fields for delegates, even if they all point to the same object).

It doesn't matter that the parameter strings are no longer generated, they were useless anyway. Their only function is to name the user created signals in the entry gen so it can be safely be removed in the api gen.

Note: 
I am not fully satisfied. It makes wrappers much lighter (for example, for Node2D, we will save more than 150 Bytes per instance) but this only applies to wrappers. I didn't find a way to make it work for scripts yet.
The main issue is user created signals need string information (for naming the parameters) from the signal() method, which means we can't directly delegate to the static object, which means creating a field.

The only way around it, that I can imagine so far, is to change the way user signals are written. (NOT IMPLEMENTED, just discussing the idea)
From
```kt
 @RegisterSignal
 val mySignal: Signal1<Int> by signal("number")
```
to 
```kt
 @RegisterSignal("number")
 val mySignal: Signal1<Int> by Signal1
```
The main issue with this approach is that we need a `vararg string` parameter in the annotation, which doesn't provide any safety regarding the number of require parameters to name. It's error-prone, as we can't rely on a method signature directly.
On top of that, it goes against our plan to remove the annotation for signals as proposed in #606


chippmann (approved):

chippmann:
You proposal for user scripts is IMO not an option. IMO safety is more important for user scripts than saving a few bytes in this case.
For the wrappers though, this PR is perfect

CedNaru:
Well there is a way to keep safety even for script signals. We just need a @RegisterSignal0, @RegisterSignal1, but it's just cumbersome to use and, again, against what we want to do.
I also was never satisfied about how we use this `signal(string)` method, the string is pure metadata that have no consequence on the code itself, the parameters are not even used by the method, they are only there for the sake of being read by KSP. 
So even the current state is quite dirty imo. But I don't have any idea that would fix it and not be a pain to use.

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/SignalGenerationService.kt
```diff
@@ -23,256 +24,285 @@ import godot.tools.common.constants.GODOT_CALLABLE
 import godot.tools.common.constants.GODOT_ERROR
 import godot.tools.common.constants.GODOT_OBJECT
 import godot.tools.common.constants.godotCorePackage
+import godot.tools.common.constants.kotlinReflectPackage
 import kotlin.reflect.KCallable
 
 class SignalGenerationService : ISignalGenerationService {
+
+    // Helper that provide utilities to handle generic classes.
+    private class GenericClassNameInfo(
```
piiertho, line 33: Could be cool to have a toTypeSpec method in this class that does:
```kotlin
TypeSpec
            .classBuilder(className)
            .addTypeVariables(genericTypes)
```

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/SignalGenerationService.kt
```diff
@@ -23,256 +24,285 @@ import godot.tools.common.constants.GODOT_CALLABLE
...
+            .addTypeVariables(genericClassNameInfo.genericTypes)
+            .addParameters(
+                (0..<argCount)
+                    .map {
+                        ParameterSpec.builder("p$it", STRING).build()
+                    }
+            )
+            .addCode(
+                CodeBlock.of(
+                    "return %T<%T, %T> { thisRef, property ->  %T(thisRef, property.name)}",
```
piiertho, line 256: replace spaces with `·`

piiertho (changes requested):

kt/tools-common/src/main/kotlin/godot/tools/common/constants/Packages.kt
```diff
@@ -17,6 +17,7 @@ const val variantCasterPackage = "$godotCorePackage.${GodotKotlinJvmTypes.varian
 
 const val kotlinPackage = "kotlin"
 const val kotlinCollectionsPackage = "$kotlinPackage.collections"
+const val kotlinReflectPackage = "$kotlinPackage.reflect"
 const val kotlinTextPackage = "$kotlinPackage.text"
-const val kotlinCoroutinePackage = "kotlin.coroutines"
+const val kotlinCoroutinePackage = "$kotlinPackage..coroutines"
```
piiertho, line 22: Why two dots here ?

piiertho (approved):

chippmann (approved):
Approving under the assumption that this will not be released as is.
As discussed on discord, I'm not fond of the reduction in ease of use of user script signal definitions. The requirement of an explicit type declaration is one thing, but the not very descriptive error message for users not familiar with the internal workings of our module is a completely separate topic.
While the memory reduction is important, ease of use is as well.

Hence on discord we agreed on a different signal definition implementation which will be implemented as a follow up PR.

chippmann:
image
This is one example of such an error message.

From the error, it's not clear that what is missing is an explicit type declaration `: Signal1<Type>`.
Especially since the provided quick fix options by the IDE are misleading

kt/godot-library/src/main/kotlin/godot/gen/godot/core/Signals.kt
```diff
@@ -75,14 +88,17 @@ public class Signal2<P0, P1>(
...
+  public companion object {
+    public operator fun <P0, P1> getValue(thisRef: Object, `property`: KProperty<*>):
+        Signal2<P0, P1> = Signal2(thisRef, property.name)
+  }
 }
 
 public inline fun <reified P0, reified P1> Signal2<P0, P1>.connect(flags: Int = 0, noinline
     method: (p0: P0, p1: P1) -> Unit): Error = connect(method.asCallable(), flags)
 
-@Suppress("UNUSED_PARAMETER")
```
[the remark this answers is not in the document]
CedNaru: It's defined at the file level now.

PR #701, opened 2024-09-25
Bump to 0.10.0

CedNaru:
Bump version to be ready for the next release.

I also commit minor changes that don't deserve their own PR. They are meant to solve issues discovered while updating the 3D demo.

- Window - Editor, JVM export, Native Image export tested
- Android - Export tested

### DIFF
--- a/docs/src/doc/user-guide/advanced/graal-vm-native-image.md
+++ b/docs/src/doc/user-guide/advanced/graal-vm-native-image.md
@@ -11,10 +11,9 @@ On Windows, you should add `VC_VARS_PATH` environment variable to point to vcvar
 In order to build a native image, you should add the following configuration to gradle plugin to enable building of native-image:
 ```kotlin
 godot {
-    isGraalExportEnabled.set(true)
-    graalVmDirectory.set(System.getenv("GRAALVM_HOME"))
-    
-    windowsDeveloperVCVarsPath.set(System.getenv("VC_VARS_PATH"))
+    isGraalNativeImageExportEnabled.set(true)
+    graalVmDirectory.set(File(System.getenv("GRAALVM_HOME")))
+    windowsDeveloperVCVarsPath.set(File(System.getenv("VC_VARS_PATH")))
 }
 ```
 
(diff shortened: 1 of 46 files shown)


docs/src/doc/user-guide/advanced/graal-vm-native-image.md
```diff
@@ -43,9 +42,9 @@ The same applies for resource files which should be added (basically any files i
 ```kotlin
 godot {
     isGraalExportEnabled.set(true)
```
chippmann, line 44: this one needs to change as well no?

chippmann:
Linux editor and exports work on the jvm
will test graalvm later

chippmann:
Tested on linux with graalvm as well.
Everything works except the jvm does not close.

CedNaru:
Confirming the issue with native image.
It seems the daemon thread trick doesn't work with native images.

chippmann:
Works now for graalvm as well

chippmann (dismissed):

chippmann (dismissed):

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupConfigurationsAndCompilations.kt
```diff
@@ -33,6 +33,7 @@ fun Project.setupConfigurationsAndCompilations() {
         with(it.dependencies) {
             add(dependencies.create("org.jetbrains.kotlin:kotlin-stdlib:${kotlinJvmExtension.coreLibrariesVersion}"))
             add(dependencies.create("com.utopia-rise:$godotLibraryArtifactName:${GodotBuildProperties.assembledGodotKotlinJvmVersion}"))
+            add(dependencies.create("com.utopia-rise:tools-common:${GodotBuildProperties.assembledGodotKotlinJvmVersion}"))
```
piiertho, line 36: Why do we add this one ?

src/script/jvm_script_manager.h
```diff
@@ -25,56 +29,54 @@ class JvmScriptManager {
...
     template<class SCRIPT>
-    Ref<SCRIPT> get_or_create_script(const String& p_path);
-
+    Ref<SCRIPT> get_or_create_script(const String& p_path, bool* created);
     void clear();
+
+#ifdef TOOLS_ENABLED
+    int64_t get_last_reload();
+
+    void invalidate_source(String path);
```
piiertho, line 48: Why not a const ref ?

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/projectExt/setupConfigurationsAndCompilations.kt
CedNaru: I need tools-common in the bootstrap, it has utilities required at runtime.
It also fixes an inconsistency. The boostrap generated for the editor had it, but not the one generated by the gradle build.

piiertho: We need to fix this after next release as we add kotlin poet to user classpath with this one.

piiertho (approved):

PR #702, opened 2024-09-26
Fix coroutines gradle flag name

chippmann:
This aligns the naming of the coroutine flag to the ones we already have and to what gradle and kotlin are doing in their plugins


CedNaru (approved):

datouzhu125:
The flag was modified but the document was not changed accordingly, which puzzled me for a long time.
image

CedNaru:
Fixed.
I directly edited master.

PR #704, opened 2024-09-27
Add some coroutines utility functions

chippmann:
This adds the following coroutine utility functions:
- awaitDeferred -> like callDeferred but for cotoutines. helpful to run stuff on the main thread from a coroutine
- awaitLoad -> load resources with progress from coroutines

### DIFF
--- a/kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitDeferred.kt
+++ b/kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitDeferred.kt
@@ -1,0 +1,32 @@
+package godot.coroutines
+
+import godot.core.Callable
+import godot.core.asCallable
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+
+
+/**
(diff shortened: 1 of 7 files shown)


kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitDeferred.kt
```diff
@@ -0,0 +1,32 @@
...
+    Callable(
+        {
+            if (continuation.isActive) {
+                try {
+                    block()
+                } finally {
+                    continuation.resume(Unit)
+                }
+            }
+        }.asCallable()
```
CedNaru, line 30: Why do you use the Callable copy constructor for that ?

CedNaru (dismissed):

kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitDeferred.kt
chippmann: Only possible like this atm as `callDeferred` only exists on `NativeCallables`

CedNaru (approved):

PR #707, opened 2024-10-01
Update IDE plugin to 2024.2

chippmann:
This updates our IDE plugin to 2024.2 and updated the gradle setup to use the new [intellij platform gradle plugin](https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html)

### DIFF
--- a/kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/gradle/GodotKotlinJvmSettings.kt
+++ b/kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/gradle/GodotKotlinJvmSettings.kt
@@ -13,6 +13,7 @@ import com.intellij.ui.EditorNotifications
 import godot.intellij.plugin.GodotPluginBundle
 import godot.intellij.plugin.notification.SettingsFetchingNotification
 import godot.plugins.common.GodotKotlinJvmPropertiesFile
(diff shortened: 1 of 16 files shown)


CedNaru (approved):

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/gradle/GodotKotlinJvmSettings.kt
```diff
@@ -13,6 +13,7 @@ import com.intellij.ui.EditorNotifications
 import godot.intellij.plugin.GodotPluginBundle
 import godot.intellij.plugin.notification.SettingsFetchingNotification
 import godot.plugins.common.GodotKotlinJvmPropertiesFile
+import godot.plugins.common.gradle.action.FetchProjectModelsBuildAction
```
piiertho, line 16: why is this added ?

piiertho (approved):

kt/plugins/godot-intellij-plugin/src/main/kotlin/godot/intellij/plugin/gradle/GodotKotlinJvmSettings.kt
chippmann: because we moved that class from the same package to the common module instead. so an import is needed

PR #710, opened 2024-10-03
Fix signal.await() when signal belongs to a Node

CedNaru:
The native Node implementation overrides the `connect()` method and adds a thread guard to it. It means that you can't connect to a node signals outside the main thread. 
So I added a function connectThreadSafe, which is not part of Godot standard API but make our coroutine implementation of `signal.await()` work. 
Behind the hood I use the Node method callThreadSafe. This methods immediately runs if in the main thread or a "scene thread" (if you use the multithreaded mode for SceneTree), but defers it otherwise.


piiertho (dismissed):

kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitResourceLoad.kt
```diff
@@ -37,51 +39,13 @@ public suspend inline fun ResourceLoader.awaitLoad(
...
-                    )
-                }
-            } while (status == ResourceLoader.ThreadLoadStatus.THREAD_LOAD_IN_PROGRESS)
-
-            if (success) {
-                continuation.resume(this@awaitLoad.loadThreadedGet(path))
-            }
-        }
-    }
+    return this.load(path)
```
chippmann, line 48: As mentioned on discord, I-m not fond of this as we block a coroutine dispatcher thread altough godot has a thread pool for these kind of things. But given the fact that godot ironically does not check multiple loads for their multithreaded resource accesses I'm fine with it for now.

kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitResourceLoad.kt
```diff
@@ -37,51 +39,13 @@ public suspend inline fun ResourceLoader.awaitLoad(
...
-    )
-
-    if (error != Error.OK) {
-        GD.printErr("Could not trigger resource load. Got error: $error")
-        return null
+    // Start a new job so we have a suspension point in case the coroutine is currently in the main thread.
+    val job = GodotCoroutine.launch(Dispatchers.Default) {
+        load(path)
     }
+    job.join()
```
chippmann, line 46: this could be made `async` instead of launch and thus directly return the result instead of calling load 2 times

chippmann (changes requested):

kt/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitResourceLoad.kt
CedNaru: Yeah it's not ideal, but the same way, using the godot pool thread is going to block a godot thread as well. And since 4.3, the pool is responsible for running servers. I think blocking a godot thread is force than blocking a coroutine thread.
This is a short term solution for now, and we'll try to offer people a new dispatcher that use Godot's pool instead.

chippmann (approved):

PR #711, opened 2024-10-03
Update exporting.md

danielperano:
Document conflict with official export templates

If you attempt to export with both the official Godot export templates and the godot-kotlin-jvm export templates installed, Godot will default to using its own and the exported game will not be able to load any scene with Kotlin/Java code. This is avoided by setting the "Custom Template" fields in the export window.

### DIFF
--- a/docs/src/doc/user-guide/exporting.md
+++ b/docs/src/doc/user-guide/exporting.md
@@ -12,7 +12,7 @@ After the export templates have been added, you can export your game. Your game 
 On desktop platforms, this also copies the JRE folder of your project in the exported game folder.
 
 !!!danger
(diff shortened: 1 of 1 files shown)


CedNaru:
This not the proper fix imo.
Godot should naturally pick the correct template. The issue is more that we messed up properly "naming" the custom build, so the editor still "thinks" it's an official build, therefore it will try to use the 4.3-stable release instead of our custom templates. I'm not talking about the editor's file name but of the name generated automatically as part of the build process.

If it's okay with you, you can close this PR and I'll do the proper fixes in the engine so it works as intended for the next release.

danielperano:
Sounds good!

PR #712, opened 2024-10-04
Make API more friendly to Java

CedNaru:
After some feedback, it appears that many part of the API are not easy to use from Java, mostly with Callables and Singleton.
I reworked the api generation with a bunch of renames and JvmStatic.

If you want to see what it looks like in Java, check the additions to the JavaTest.java file.

Note that I also renamed KtCallable to LambdaCallable. In Kotlin, we don't never need directly write that type because of all utilities and type-inferences but you have to be more explicit in Java so I picked a named that I think better represent the intent.
On top of that the name "KtCallable" was a leftover from when the class was also the parent of "KtFunction" used for registration. But now those classes are unrelated and KtCallable was its own thing.

Registration classes are meant to be used internally for us and not by the users so we have been using that "Kt" prefix for a while, but I don't think it provides any hint for user, specially when that name is mostly going to be used in Java 

Aside of the name change, the additions to "LambdaCallable" is a helper static method to create an instance from Java, the others being inline reified, they can't be used from Java.

I also renamed the kotlin inline functions `callable()` to `callableX()` to be consistent with the naming scheme used by signals.


chippmann (dismissed):

chippmann (approved):

src/jvm_wrapper/bridge/kt_callable_bridge.h
```diff
@@ -5,12 +5,12 @@
 #include "jvm_wrapper/jvm_singleton_wrapper.h"
 
 namespace bridges {
-    JVM_SINGLETON_WRAPPER(KtCallableBridge, "godot.core.KtCallable$Bridge") {
-        SINGLETON_CLASS(KtCallableBridge)
+    JVM_SINGLETON_WRAPPER(LambdaCallableBridge, "godot.core.LambdaCallable$Bridge") {
```
piiertho, line 8: maybe we should also rename cpp and header files ?

piiertho (dismissed):

PR #713, opened 2024-10-04
Fix template usage

CedNaru:
Currently, the editor is not able to automatically use our templates because there is a mismatch between its name and the name of the template. We have so far build our custom editor using the same "name" as the official godot, which creates a conflict.

I made the changes so that the next release is going to use the name `4.3.stable.jvm.0.11.0`.
We can't change the prefix `4.3-stable` unless we directly Godot files, but we can append anything we want to it, so I picked `jvm.ourcurrentversion`

Without that change, users can't properly export without manually setting a release and a debug export template in their export preset.
This issue didn't happened before because we were using the same name as the official template for ours. I changed last before the last release so people could have both the official and the koltin versions of a template without name conflict. I just didn't realise that the editor name had to exactly the template to work..

So it's either this change or we revert all names to the same as the official release, preventing users to use both Kotlin and non kotlin versions of the templates.

Note that because, this PR changes the names of the build, opening the template manager create this warning:
image
It doesn't prevent anything, it's just an annoying warning because the editor is hardcoded to search the official godot website.

I created a proposal that would allow us to set a link to our GitHub, so users can download templates directly from there:
https://github.com/godotengine/godot-proposals/issues/10894

### DIFF
--- a/docs/src/doc/user-guide/exporting.md
+++ b/docs/src/doc/user-guide/exporting.md
@@ -14,7 +14,7 @@ On desktop platforms, this also copies the JRE folder of your project in the exp
 !!!warning
     The official export templates from Godot will not work! You have to use our export templates or build your own!
     Note that because this is not an official build, the template manager will fail to find a link to our templates and display a warning.
-    Downloading our templates from the Godot editor is currently not possible, we opened a [proposal](godotengine/godot-proposals#10894) for it.
+    You can simply ignore it. Downloading our templates from the Godot editor is currently not possible, we opened a [proposal](godotengine/godot-proposals#10894) for it.
 
 
 ## Requirements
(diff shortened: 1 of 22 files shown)


piiertho (approved):

chippmann (changes requested):
Please also add why the warning appears to the docs and possibly link to your proposal as well as part of the explanation.

piiertho:
Why CI is failing on test export ?

piiertho (approved):

docs/src/doc/user-guide/exporting.md
```diff
@@ -11,8 +11,10 @@ To export your game, you need to add our export templates to Godot. After you've
 After the export templates have been added, you can export your game. Your game `jar` will be included in `pck`.
 On desktop platforms, this also copies the JRE folder of your project in the exported game folder.
 
-!!!danger
-    The official export templates from Godot will not work! You have to use our export templates or build your own! Therefore, you cannot just hit the "Download and Install" button in the export template manager!
+!!!warning
+    The official export templates from Godot will not work! You have to use our export templates or build your own!
+    Note that because this is not an official build, the template manager will fail to find a link to our templates and display a warning.
+    Downloading our templates from the Godot editor is currently not possible, we opened a [proposal](godotengine/godot-proposals#10894) for it.
```
chippmann, line 17: I would also indicate that this warning can be ignored safely

chippmann (dismissed):
One nit. I leave it up to you if you want to change that or not.

PR #726, opened 2024-10-13
Add new common module

CedNaru:
Currently, the `godot-library` module has a dependency on `tools-common` (string operations and some constant).
I move those dependencies to a new `common` module now used by `godot-library` and  `tools-common`.

I also decided to move the utils package from  `godot-library` to this new module. This serves as a first step toward making the API split easier. 
So far the plan is to have a 4 module chain `Common `=> `Internal `=> `Core `=> `API`, it means that I need a bunch of utilities in that common module. To that end, I also created a new "interop" package with basic logic to handle C++ pointers.

The number of line changes is scary but it's 99% just renames and package updates. 
The only new stuff is the module itself + its "interop" package.


kt/godot-library/build.gradle.kts
```diff
@@ -37,7 +37,7 @@ dependencies {
     // added here as a transitive dependency so the user can use reflection
     // we need to add it here so reflection is available where the code is loaded (Bootstrap.kt) otherwise it will not work
     api(kotlin("reflect", version = libs.versions.kotlin.get()))
-    implementation("com.utopia-rise:common:$fullGodotKotlinJvmVersion")
+    api("com.utopia-rise:common:$fullGodotKotlinJvmVersion")
```
chippmann, line 40: Why should this be a transitive dependency?

chippmann (changes requested):

kt/godot-library/build.gradle.kts
CedNaru: Some of those needs to be usable in the user code, so `implementation` is not enough, otherwise I can't even compile the tests. It technically exposes more things than before to the user, but they are all utilities without any global state, so there is no risk. Also, I can still move a few of them to the new Internal module in my next PR if we really want to hide some. But otherwise, everything in common is supposed to be 100% public.

chippmann (approved):

kt/tools-common/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ kotlin {
 
 dependencies {
     implementation(libs.kotlinPoet)
+    api("com.utopia-rise:common:$fullGodotKotlinJvmVersion")
```
piiertho, line 21: Why is it transitive ?

piiertho (approved):

PR #727, opened 2024-10-13
Refine registration annotation handling

### LINKED ISSUE #606: A proposal for reducing the need for @Register annotations

chippmann:
Closes #606 

This refines how we register annotations according to the linked proposal (specifically [this comment](https://github.com/utopia-rise/godot-kotlin-jvm/issues/606#issuecomment-2052530824) and refined by [this comment](https://github.com/utopia-rise/godot-kotlin-jvm/issues/606#issuecomment-2408631941))

It includes the following changes:
- `RegisterClass` is renamed to `GodotScript`
- `RegisterSignal` is dropped
- `RegisterProperty`, `RegisterConstructor` and `RegisterFunction` are combined into `GodotMember`
- Entry generation updated accordingly
- Compile time checks updated accordingly
- IDE plugin checks and warnings updated accordingly

To summarize the implemented registration changes from the proposal:

## All
If a fine grained annotation is used (like `Export`, a type hint annotation, `Rpc` or `Tool`) a class or member does not have to be registered explicitly anymore. Annotations inherit from each other now

## Classes
Classes still need to be registered manually (reason for this is outlined in [this comment](https://github.com/utopia-rise/godot-kotlin-jvm/issues/606#issuecomment-2048465300)).

## Signals
Signals no longer need to be registered at all. As they are only of use in Godot anyway, their sole existence in code is already documentation and hint enough that they should be exposed to godot

## Properties
Still need to be annotated explicitly. If a property is registered in a parent class, it "inherits" that registration status even if not explicitly registered again in a child class (no matter if the child class overrides the property or not).

## Functions
User defined functions still need to be registered manually. Like properties, they inherit their registration status from parent classes.
Godot api functions which are overridden in child classes (like `_ready` or `_process`) do not need to be registered explicitly anymore. They are godot api functions and thus are always relevant to godot.


CedNaru (dismissed):

chippmann:
Will update this PR after #761 is merged

PR #732, opened 2024-10-18
feat(coroutine): Add LambdaCallable::onCancel to cancel coroutine when callable is cancelled

piiertho:
This adds a `onCancel` callback on `LambdaCallable` to cancel coroutine when a callable used within is canceled.

### DIFF
--- a/kt/api-generator/src/main/kotlin/godot/codegen/services/impl/AwaitGenerationService.kt
+++ b/kt/api-generator/src/main/kotlin/godot/codegen/services/impl/AwaitGenerationService.kt
@@ -14,6 +14,7 @@ import com.squareup.kotlinpoet.TypeVariableName
 import com.squareup.kotlinpoet.UNIT
 import godot.codegen.services.IAwaitGenerationService
 import godot.common.constants.Constraints
+import godot.tools.common.constants.AS_CALLABLE_UTIL_FUNCTION
 import godot.tools.common.constants.GODOT_OBJECT
 import godot.tools.common.constants.GodotKotlinJvmTypes.signal
 import godot.tools.common.constants.godotCorePackage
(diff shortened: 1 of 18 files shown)


kt/godot-library/src/main/kotlin/godot/gen/godot/core/LambdaCallables.kt
```diff
@@ -25,27 +27,30 @@ public class LambdaCallable0<R> @PublishedApi internal constructor(
     @JvmStatic
     @JvmName("create")
     public fun <R> javaCreate(returnClass: Class<R>, function: () -> R) =
-        LambdaCallable0(variantMapper.getOrDefault(Reflection.getOrCreateKotlinClass(returnClass), NIL), function)
+        LambdaCallable0(variantMapper.getOrDefault(Reflection.getOrCreateKotlinClass(returnClass), NIL), null, function)
   }
 }
 
-public inline fun <reified R> callable0(noinline function: () -> R) =
```
CedNaru, line 32: I think the onCancelCall should have a default to null, so users are not forces to write it

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/AwaitGenerationService.kt
```diff
@@ -13,17 +13,14 @@ import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.TypeVariableName
 import com.squareup.kotlinpoet.UNIT
 import godot.codegen.services.IAwaitGenerationService
-import godot.tools.common.constants.GODOT_OBJECT
+import godot.tools.common.constants.*
```
CedNaru, line 16: Don't transform imports into stars

CedNaru (changes requested):

kt/godot-library/src/main/kotlin/godot/gen/godot/core/LambdaCallables.kt
piiertho: done

kt/api-generator/src/main/kotlin/godot/codegen/services/impl/AwaitGenerationService.kt
piiertho: done

CedNaru (dismissed):

CedNaru (approved):

Issue #488, opened 2023-08-13
Add the possibility to register scripts without constructor


CedNaru:
In my project, I have many cases where I need to create classes inheriting a Godot base type (well, most of the time inheriting a Node). 
Except, the construction of the class itself requires parameters that are not Variant types (can be any other internal class). Those classes are not supposed to be instantiated from Godot itself, only from within Kotlin code directly.

Right now, my cheap trick to go around that is just to leave the 0 parameter constructor there and have a bunch of lateinit properties. But it's unnecessary, I'm never going to use the default 0 parameter constructor, and it also adds the script to the list of Node I can add directly in the editor, which obviously won't work.

The goal is that I want to be able to use dependency injection easily with my scripts. I know that Godot can accept scripts that can't be instantiated, but I don't know how flexible it is outside abstract scripts.

The requirement for such feature is to check if the Godot editor properly behave when dealing with scripts without constructor. 
In the best case, it shouldn't be possible to attach the script in the editor at all (manually, by custom type creation or GDScript). 
In the worst case, attaching that script would result in an error or undefined behaviour and writing such script in Kotlin should be an opt in.

PR #736, opened 2024-11-03
Abstract class registration

### LINKED ISSUE #730: Unable to assign subclass node in the editor when it inherits an abstract class.
`@RegisterClass
class Sample : Node2D() {
    @RegisterProperty @Export lateinit var node: Parent
}

@RegisterClass
abstract class Parent : Node2D() {
}

@RegisterClass
class Child : Parent() {
}
`
image

Modify Parent to a normal class, then you can assign.
`
@RegisterClass
open class Parent : Node2D() {
}
`
image

By the way, when I made the change, after building, the editor crashed,here's the error log.
[hs_err_pid23860.log](https://github.com/user-attachments/files/17415619/hs_err_pid23860.log)

chippmann:
This adds support for abstract class registration (both regular abstract classes as well as registered interfaces).

Should resolve #730 and #488


chippmann:
Will update this PR after #761 is merged

CedNaru:
Closed because it will be part of the registration rework

PR #739, opened 2024-11-03
Use java dir instead of kotlin dir for initial project generation

### LINKED ISSUE #578: Game Crashing if gdj files from Java sources are being used
When you use gdj files which have been generated from Java Source Files, the game crashes on start with a NoClassDefFoundError, which is only shown when running the editor itself from the terminal.  The same gdj file, but generated from a Kotlin class works fine.

The project is basically just the standard project generated by the IntelliJ plugin with renamed packages.

Here's both the log and the project I've been using:
[log.txt](https://github.com/utopia-rise/godot-kotlin-jvm/files/14270778/log.txt)
[SuikaGame.zip](https://github.com/utopia-rise/godot-kotlin-jvm/files/14271152/SuikaGame.zip)

chippmann:
Resolves #578


CedNaru:
Why do we have to create our Kotlin scripts in a Java directory now ? So far, it has been working perfectly fine to have each language in its own directory. This feels like a terrible and nonsensical change.
Isn't that just up to the users to create their java files in the correct directory ? We don't have any responsability regarding that. 
If we want to improve the IDE plugin to help people making less mistake, I'd rather just add an option in the setup wizard to generate a template script, not only for Kotlin, but also for Java. And they would be both directly created in their respective directory.

chippmann:
closing; as discussed on discord, this is not really something we should do. Now that we soon will support scala as well, it does not make sense just generating the sample code in a java dir
i was a bit misguided by the default android project setup.
But the convention is to have a dir per language

hence we either leave that up to the user, or we provide a setting in the project wizard where one can select all the languages one want's to have in the project. I'll have a look at that as part of #776

Issue #163, opened 2021-04-21
Implement JRE generation task


chippmann:
Best would be a task triggered by the Editor plugin which generates the jre when a new kotlin script is created (similar to the C# plugin) or when run is pressed but no jre is present (for the case when scripts are solely created outside the editor with Intellij for example). Also some good error messages would be nice (for example no JAVA_HOME found, JAVA <= 9 installed, and so on...).

Another possibility could be to handle this with a gradle task which get's called by the editor plugin.

Best might be to have both and execute the gradle task upon build and it checks on each build if a jre is present. So we cover the case where the godot editor is never started.

piiertho:
I agree we should make it as a gradle task, this would be easier for packaging games from CI

PR #740, opened 2024-11-03
Add GenerateEmbeddedJreTask to generate JRE

chippmann:
Resolves https://github.com/utopia-rise/godot-kotlin-jvm/issues/163

### DIFF
--- a/kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
+++ b/kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
@@ -9,7 +9,10 @@ import java.util.*
 open class GenerateEmbeddedJreTask : DefaultTask() {
 
     @Input
-    var modules: String = "java.base,java.logging"
+    var modules: Array<String> = arrayOf(
+        "java.base",
+        "java.logging",
+    )
 
     @Input
     var outputDir: String = "jvm/jre-${getArch()}-${getOs()}"
@@ -30,13 +33,13 @@ open class GenerateEmbeddedJreTask : DefaultTask() {
         project.exec {
             it.commandLine(
                 "$javaHome/bin/jlink",
-                "--add-modules", modules,
+                "--add-modules", modules.joinToString(","),
                 "--output", outputDir,
                 *arguments,
             )
         }
 
-        logger.lifecycle("Custom JRE created in $outputDir using modules: '$modules', arguments: '${arguments.joinToString(" ")}' and java home: $javaHome")
+        logger.lifecycle("Custom JRE created in $outputDir using modules: '${modules.joinToString(",")}', arguments: '${arguments.joinToString(" ")}' and java home: $javaHome")
     }
 
     private fun getArch(): String {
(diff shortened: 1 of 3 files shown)


chippmann:
I opted against implementing more sophisticated tasks like an export task (like we have for the harness/tests project) as these are highly project specific and not hard to implement. Hence in my opinion, it's not worth shipping such tasks as part of our project

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
```diff
@@ -0,0 +1,62 @@
...
+open class GenerateEmbeddedJreTask : DefaultTask() {
+
+    @Input
+    var modules: Array<String> = arrayOf(
+        "java.base",
+        "java.logging",
+    )
+
+    @Input
+    var outputDir: String = "jvm/jre-${getArch()}-${getOs()}"
```
CedNaru, line 18: Wonder if we should declare the jvm directory path in the tool-common directory, to avoid hardcoding it here.

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @Input
+    var arguments: Array<String> = arrayOf(
+        "--strip-debug",
+        "--no-header-files",
+        "--no-man-pages"
+    )
+
+    @Input
+    var javaHome: String = System.getProperty("java.home")
```
CedNaru, line 28: Can we potentially use the JVM running the gradle task if JAVA_HOME is not set ? Also can we make sure it's at least 11 ?

CedNaru (changes requested):
We should probably add a new task in the godot editor plugin to generate it as well, right next to the Kotlin project generation.

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
chippmann: done

kt/plugins/godot-gradle-plugin/src/main/kotlin/godot/gradle/tasks/GenerateEmbeddedJreTask.kt
chippmann: JAVA_HOME selection done.
The check if we're on java 11 is needless imo as jlink was introduced in java 9 and we simply set the requirement for java 11 as it was the next lts release after 9
if the user uses an older java version than java 11 i think he cannot build the project anyways

chippmann:
> We should probably add a new task in the godot editor plugin to generate it as well, right next to the Kotlin project generation.

I'd like to do that in a separate PR as I'd like to rework how the editor plugin executes gradle tasks as now we're executing more than just the build task

CedNaru:
I was talking about the Godot editor plugin, but sure it can also be a different PR.

chippmann:
> I was talking about the Godot editor plugin, but sure it can also be a different PR.

I did as well. I'd like to rewrite the build manager to be more generic in regards to gradle task execution

CedNaru (approved):

PR #746, opened 2024-12-31
Split core into core, api and extension modules

CedNaru:
The godot library is now split into several modules with the following stack each depending on the previous ones:

- Common
- Internal
- Core
- Api
- Extension

Api now purely contains generated code. It will make the custom api generator much easier to implement as we will only need to swap our default one with a local, created by the users.

I'll probably make another PR on that matter. So far I have focused on making the split functional, but there is a bit of cleaning to do. This PR is already big enough so I'll leave that for later.

Don't be scared by the number of files, most of them are just moved to different directory. 90% of the changes are in the gradle scripts and the API generation.


kt/api-generator/src/main/kotlin/godot/codegen/services/impl/ApiService.kt
```diff
@@ -120,4 +130,46 @@ class ClassService(
             }
         }
     }
+
+    override fun findEnumValue(enumClassName: ClassTypeNameWrapper, enumValue: Long): DefaultEnumValue {
```
piiertho, line 134: I would rename `findEnumValue` to `findDefaultEnumValue`

piiertho (approved):

PR #750, opened 2025-01-15
chore: Move signals parameter names to annotation

piiertho:
This moves signal parameter names to annotations.


kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/SignalRegistrationGenerator.kt
```diff
@@ -50,7 +50,7 @@ object SignalRegistrationGenerator {
                 add(ClassName(godotRegistrationPackage, GodotKotlinJvmTypes.ktFunctionArgument))
                 add(argumentType.toKtVariantType())
                 add(argumentType.fqName)
-                add(argumentName)
+                add("\"$argumentName\"")
```
chippmann, line 53: Please change it here so it's a string and not a literal anymore: https://github.com/utopia-rise/godot-kotlin-jvm/blob/5fb0c6aadd64ed9f06e0706853121a2a845bb4e0/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/SignalRegistrationGenerator.kt#L34

With this, the additional quotes are not necessary anymore

chippmann (changes requested):
I'd also change https://github.com/utopia-rise/godot-kotlin-jvm/blob/5fb0c6aadd64ed9f06e0706853121a2a845bb4e0/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/SignalRegistrationGenerator.kt#L49 the default name with this change to be p0 p1 and so on based on the param index. Previously this fallback was almost impossible to reach because of the argument type safety. But now it's much easier to get into this case as there are no static type checks anymore.

so possible arg name conflicts are much more likely to occure

CedNaru (approved):

CedNaru (changes requested):
You have to update the documentation.

piiertho:
> I'd also change
> 
> https://github.com/utopia-rise/godot-kotlin-jvm/blob/5fb0c6aadd64ed9f06e0706853121a2a845bb4e0/kt/entry-generation/godot-entry-generator/src/main/kotlin/godot/entrygenerator/generator/SignalRegistrationGenerator.kt#L49
> 
> the default name with this change to be p0 p1 and so on based on the param index. Previously this fallback was almost impossible to reach because of the argument type safety. But now it's much easier to get into this case as there are no static type checks anymore.
> so possible arg name conflicts are much more likely to occure

done !

CedNaru (approved):

CedNaru (approved):

PR #762, opened 2025-03-05
Implement typed dict support

### LINKED ISSUE #686: Support typed dictionaries
This PR was merged recently: https://github.com/godotengine/godot/pull/78656
We already have typed dictionnaries implemented on the Kotlin side, but we now have official support in C++ as well. It mostly means adding a few more native callls in our Dictionary implementation, and update the API gen to make use of them.

chippmann:
This implements support for typed dictionaries.
~~It has the same limitations and bugs as typed arrays for now.~~

~~Meaning, exporting a typed array/dictionary with nodes using node paths does not yet work.~~ <-- this is now also fixed in this PR

Base branch is the 4.4 update branch as it depends on it.

This also fixes a bug with registering resources in typed arrays and dicts as before i was wrongly using the enum ordinals for type hint generation instead of the enum id. That was a leftover from early in the entry gens history and is resolved now.
This also fixes the mapping of nodes to nodepaths on the godot side. We used the wrong property hint there

Resolves #686


chippmann:
Note: I did not yet clean up the whole hint string generation. I plan to do that in a later PR maybe together with the "node in container fix".
For now this basically just extends the array hint string generation to support dicts as well (it's the same format just the hint strings separated by `;`. So `VariantArray<Int>()` -> `2:int` and `Dictionary<Int, Button>()` -> `2:int;24/34:Button` to give a simpler example

CedNaru (approved):

piiertho (approved):

PR #769, opened 2025-03-09
Add documentation on known issue with microsofts jdk on windows

chippmann:
This adds a bit of warning documentation on the problems encountered while using microsofts jdk builds on windows when trying to build our project

The issue is outlined [here](https://github.com/microsoft/openjdk/issues/339) and also applicable to us.

Even if specifying a jvm vendor manually using jvmToolchain, the task `instrumentCode` from the intellij plugin sdk does not adhere to that constraint.

So there is nothing else we can do on our side to fix this issue.


chippmann:
I also took the opportunity to update the ways on how to install the jdk in general

chippmann:
@aaronjyoder FYI

CedNaru (approved):

piiertho (approved):

PR #771, opened 2025-03-10
 Run android builds in parallel and minor improvements

chippmann:
This runs the android builds in parallel and adds some minor improvements to the CI/CD pipeline, like only defining a fixed linux version where necessary.

**Note:** I would only merge this after the next release is done in case there is an issue i did not see

### DIFF
--- a/.github/workflows/assemble_android.yml
+++ b/.github/workflows/assemble_android.yml
@@ -1,0 +1,100 @@
+name: 🤖 Assemble Android
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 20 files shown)


chippmann:
The two checks are pending as they do not exist in my PR anymore but are a requirement for PR workflows to pass. So they can be ignored in this case.
After this PR is merged, the PR requirements are updated automatically

CedNaru (approved):

CedNaru (approved):

PR #775, opened 2025-03-15
Shorten enum names

CedNaru:
we remove from enum values the name of the enum itself according to 3 patterns:

- Partial match of the prefix
- Partial match of the suffix
- The entire name in the middle


kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
```diff
@@ -48,6 +48,48 @@ fun String.convertToSnakeCase(): String =
         else accumulator.append(character)
     }.toString()
 
+fun String.toScreamingSnakeCase(): String {
```
piiertho, line 51: Can you change this method name ? Like `toUpperSnakeCase` ?

piiertho (approved):

kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
CedNaru: Done

kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
```diff
@@ -48,6 +48,48 @@ fun String.convertToSnakeCase(): String =
         else accumulator.append(character)
     }.toString()
 
+fun String.toUpperSnakeCase(): String {
+    if (this.isEmpty()) return this
+
+    val result = StringBuilder()
```
chippmann, line 54: why use `StringBuilder` instead of just `buildString{}`? IMO that's more concise

kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
```diff
@@ -48,6 +48,48 @@ fun String.convertToSnakeCase(): String =
...
+}
+
+fun String.removePrefixWords(wordWithDashes: String): String {
+    // We split the 2 strings into different "words"
+    val otherWords = wordWithDashes.split('_')
+    val valueWords = this.split('_')
+
+    var index = 0
+    for (enumWord in otherWords) {
+        if (index >= valueWords.size) break
```
chippmann, line 77: just a `otherWords.forEach` would do the trick?

kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
```diff
@@ -48,6 +48,48 @@ fun String.convertToSnakeCase(): String =
...
+            index++
+        } else {
+            break
+        }
+    }
+    // Join the remaining words with underscores.
+    return valueWords.drop(index).joinToString("_")
+}
+
+fun String.removeWords(other: String) =replace("_" + other + "_", "_")
```
chippmann, line 91: formatting ` =<space>`
and just using string templation would be nicer IMO

kt/common/src/main/kotlin/godot/common/extensions/StringExtensions.kt
CedNaru: You can't break in a foreach, if you write a return@ForEach, it's continue to be equivalent to a continue.

chippmann: didn't register you looped over `otherWords` and checked against `valueWords` sorry about that

chippmann (approved):

PR #783, opened 2025-03-23
Use PagedAllocator for native core types

CedNaru:
We profiled a long time ago that the constant use of memnew in kt_variant.h was a potential bottleneck in situation where a lot of data were exchange between Godot and the JVM. It wasn't high priority, but 2 years past, and we still didn't take the time to adress it.
I created our own little VariantAllocator, taking heavy inspiration of what Godot itself does in its implementation of Variant.
https://github.com/godotengine/godot/blob/2303ce843a362cf5497ca3336451de23793eb711/core/variant/variant.h#L164

### DIFF
--- a/src/variant_allocator.h
+++ b/src/variant_allocator.h
@@ -1,0 +1,67 @@
+#ifndef GODOT_JVM_VARIANT_ALLOCATOR_H
+#define GODOT_JVM_VARIANT_ALLOCATOR_H
+
+#include <core/templates/paged_allocator.h>
+#include <core/variant/variant.h>
+
+class VariantAllocator {
+    union BucketSmall {
+        BucketSmall() {}
+        ~BucketSmall() {}
+
+        StringName string_name;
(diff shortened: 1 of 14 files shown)


src/variant_allocator.h
```diff
@@ -0,0 +1,69 @@
...
+    template<typename T>
+    static T* alloc(T&& variant) {
+        // Trigger compile-time error if T is larger than 16 bytes
+        static_assert(sizeof(T) <= sizeof(BucketLarge), "Variant doesn't fit inside the VariantAllocator");
+
+        if constexpr (sizeof(T) <= sizeof(BucketSmall)) {
+            T* ret = reinterpret_cast<T*>(bucket_small.alloc());
+            memnew_placement(ret, T(variant));
+            return ret;
+        } else {
```
piiertho, line 48: Else is not useful here

piiertho (approved):

PR #784, opened 2025-03-23
Add batched conversion to PackedVector arrays

CedNaru:
Like we did with the byte/int/float versions of PackedArrays, I added another batch conversion for Vector PackedArray.
It's the same pattern as before, except there is an extra conversion step to/from a FloatArray (because obviously the JVM doesn't have any native Vector2 arrays).
For example, we pass a Vector2 has a pair of floats, so an array of Vector2 is converted into an array of floats with double the size.

### DIFF
--- a/harness/tests/src/main/kotlin/godot/tests/packedarray/PackedArrayTest.kt
+++ b/harness/tests/src/main/kotlin/godot/tests/packedarray/PackedArrayTest.kt
@@ -8,6 +8,8 @@ import godot.core.PackedFloat32Array
 import godot.core.PackedFloat64Array
 import godot.core.PackedInt32Array
 import godot.core.PackedInt64Array
+import godot.core.PackedVector2Array
+import godot.core.Vector2
 
 @RegisterClass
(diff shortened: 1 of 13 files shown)


harness/tests/src/main/kotlin/godot/tests/packedarray/PackedArrayTest.kt
```diff
@@ -8,67 +8,109 @@ import godot.core.PackedFloat32Array
...
+    }
+
+    @RegisterFunction
+    fun getVector3ArrayValue(arr: PackedVector3Array, index: Int): Vector3 {
+        val kotlinArr = arr.toVector3Array()
+        return kotlinArr[index];
+    }
+
+    @RegisterFunction
+    fun getVector5ArrayValue(arr: PackedVector4Array, index: Int): Vector4 {
```
chippmann, line 112: typo 5 instead of 4.
Probably the source of the failing tests

chippmann (changes requested):

Issue #744, opened 2024-12-24
Add the possibility to override toString.


CedNaru:
Godot uses the toString method and so that function is implemented by the api generation and made final, preventing users to customize it.
_toString should be used instead as planned by Godot (to do in ScriptInstance)

PR #792, opened 2025-03-25
Add missing core Object methods.

CedNaru:
Solve #744
It was technically possible to make those methods works, but they are special cases (like notification() and are not part of the regular Godot API so they are not generated alongside Object.kt.
I add them as unimplemented methods in KtObject and fixed toString in the Script instance.

### DIFF
--- a/kt/godot-library/godot-core-library/src/main/kotlin/godot/core/KtObject.kt
+++ b/kt/godot-library/godot-core-library/src/main/kotlin/godot/core/KtObject.kt
@@ -55,6 +55,15 @@ abstract class KtObject : NativeWrapper {
 
     protected abstract fun new(scriptIndex: Int)
 
+    open fun _get(property: StringName): Any = throw NotImplementedError("_get is not implemented for Object")
+    open fun _getPropertyList(): VariantArray<Dictionary<Any, Any>> = throw NotImplementedError("_getPropertyList is not implemented for Object")
+    open fun _propertyCanRevert(name: StringName) : Boolean = throw NotImplementedError("_propertyCanRevert is not implemented for Object")
+    open fun _propertyGetRevert(name: StringName): Any = throw NotImplementedError("_propertyGetRevert is not implemented for Object")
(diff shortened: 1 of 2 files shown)


kt/godot-library/godot-core-library/src/main/kotlin/godot/core/KtObject.kt
```diff
@@ -55,6 +55,15 @@ abstract class KtObject : NativeWrapper {
 
     protected abstract fun new(scriptIndex: Int)
 
+    open fun _get(property: StringName): Any = throw NotImplementedError("_get is not implemented for Object")
+    open fun _getPropertyList(): VariantArray<Dictionary<Any, Any>> = throw NotImplementedError("_getPropertyList is not implemented for Object")
+    open fun _propertyCanRevert(name: StringName) : Boolean = throw NotImplementedError("_propertyCanRevert is not implemented for Object")
+    open fun _propertyGetRevert(name: StringName): Any = throw NotImplementedError("_propertyGetRevert is not implemented for Object")
+    open fun _set(name: StringName, value: Any) : Unit = throw NotImplementedError("_set is not implemented for Object")
+    open fun _toString(): String = throw NotImplementedError("_toString is not implemented for Object")
```
chippmann, line 63: Shouldn't the default delegate to the regular `toString` method instead?

In the current state, doesn't that crash when one tries to print a jvm object from gdscript for example and does not override this method?
or does it also need to be registered for godot to be called in the first place like the rest?

chippmann (approved):
More of a question than anything else. Otherwise looks good

kt/godot-library/godot-core-library/src/main/kotlin/godot/core/KtObject.kt
CedNaru: The way it works.
The native C++ implementation of `toString()` is the one that call `_toString()` IF it's implemented in the script, in the sense it is registered. It's the same for all case of `method()` and `_method()`.
Right now the JVM side just have a dummy implementation, so the class is not abstract and can be override but the script itself is not aware of it.
The user is then supposed to override it to make it useable.

Issue #765, opened 2025-03-07
[Feature] Add signal connection handle


chippmann:
When one connects to a signal using a lambda:
```kotlin
mySignal.connect { arg1, arg2 ->
    // do stuff
}
```

one receives a godot Error object as a return type from the `connect` function.

While this allows to check if the connection was successful, one can never disconnect that lambda again from the signal.

After some discussions on discord, we settled on a design which returns a handle for that connect function call. The handle will look something like this:

**Note: names not final and subject to change. Probably they will be called `Connector` or similar**
```kotlin
class SignalHandle(
    private val signal: Signal,
    private val callable: Callable,
) {
    fun connect(flags: Int = 0): Error = signal.connect(callable, flags)
    fun disconnect(): Unit = signal.disconnect(callable)
    fun isConnected(): Boolean = signal.isConnected(callable)
    fun isNull(): Boolean = signal.isNull()
}
```

The connect call taking in the lambda will then return such a handle which allows to later disconnect and reconnect the lambda at will while still retaining the fire and forget (or better setup and forget) way of working which we have today and is quite easy to use:
```kotlin
// setup and forget:
mySignal.connect { arg1, arg2 -> }

// ---

// using handle:
val handle = mySignal.connect { arg1, arg2 -> }

// later
if (handle.isConnected()) {
    handle.disconnect()
}

// later
handle.connect()

// later
handle.disconnect()
```

The generated code would look like something this:
```kotlin
public inline fun <reified P0> Signal1<P0>.connect(flags: Int = 0, noinline method: (p0: P0) -> Unit): SignalHandle = SignalHandle(
    signal = this,
    callable = method.asCallable(),
).also { it.connect(flags) }
```

This would ideally be implemented after #760 to take advantage of the api gen's new design.

chippmann:
This also requires proper memory management to be implemented for custom callables, as currently we create a new wrapper each time we send a callable to gdscript. So from the cpp point of view, the callable is a different one when we call disconnect

PR #803, opened 2025-04-06
Improve Custom Callables and add SignalConnector

CedNaru:
Implement #765

First of a series of 3 PRs aiming at improving the Signal and Callable.
This one only focus on improving the Callable interface and fixing memory issues with CustomCallable.

For now, I disabled Java. I will remake it in the next PRs. 
Don't worry, I'm targeting a working branch, not master.


kt/godot-library/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitResourceLoad.kt
```diff
@@ -1,48 +1,32 @@
...
  * @param cacheMode The cache mode to be used while loading the resource.
- * @param onProgress A callback function to track the progress of resource loading.
  * @return The loaded resource, or null if there was an error.
  */
 public suspend inline fun ResourceLoader.awaitLoad(
     path: String,
     typeHint: String = "",
-    useSubThreads: Boolean = false,
-    cacheMode: CacheMode = ResourceLoader.CacheMode.REUSE,
-    crossinline onProgress: (RealT) -> Unit = {},
```
chippmann, line 36: I'd love for this to make a return at some point :-) Just a note not really related to this PR

chippmann (approved):

kt/godot-library/godot-coroutine-library/src/main/kotlin/godot/coroutines/awaitResourceLoad.kt
CedNaru: The parameters were actually unused, we forgot to remove them when we realised we couldn't use the threaded requests for the ResourceLoader.
As a reminder, despite being threaded, it doesn't support multiple threads requesting the same resource at the same time >< which is an issue when we try to use coroutines to load several instances of the same scene.

PR #815, opened 2025-04-19
GH-801 Exclude jvm dir from export

### LINKED ISSUE #801: Failure during export for Android
Godot Kotlin version: 0.12.3. 
Plugin version: 0.12.3-4.4.1-IJ2024.3.3

Image

Image

chippmann:
Resolves #801 

This explicitly excludes our jvm dir by default. The jars we need we add manually in the export plugin, and this takes predecence over what we exclude here. So these files are still present.

Before this change, these jar files were added twice, breaking the android export.

@CedNaru This change kinda prevents your idea of loading jars dynamically with the jar resource loader.
We can either worry about this once we implement multi jars or we can instead, not explicitly add the bootstrap jar and the main jar in the export plugin anymore and instead let it happen atomatically through the resource loader.

Because of this decision; this is still a draft

### DIFF
--- a/src/kotlin_editor_export_plugin.cpp
+++ b/src/kotlin_editor_export_plugin.cpp
@@ -177,10 +177,16 @@ void KotlinEditorExportPlugin::_generate_export_configuration_file(jni::JvmType 
         get_export_preset()->set_exclude_filter(get_export_preset()->get_exclude_filter() + "," + JVM_CONFIGURATION_PATH);
     }
 
-    // exclude build folder
-    get_export_preset()->set_exclude_filter(get_export_preset()->get_exclude_filter() + "," + BUILD_DIRECTORY + "/*");
-    // exclude any jars in the embedded jre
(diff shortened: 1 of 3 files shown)


chippmann:
As discussed on discord, will change it so that we no longer add the bootstrap and main jar manually but rely on automatic add and instead manually exclude the build dir and jars in the jre

src/kotlin_editor_export_plugin.cpp
```diff
@@ -179,6 +177,17 @@ void KotlinEditorExportPlugin::_generate_export_configuration_file(jni::JvmType
         get_export_preset()->set_exclude_filter(get_export_preset()->get_exclude_filter() + "," + JVM_CONFIGURATION_PATH);
     }
 
+    if (const String build_dir = String {BUILD_DIRECTORY}.path_join("*"); !get_export_preset()->get_exclude_filter().contains(build_dir)) {
```
CedNaru, line 180: You are inside the function that generate the configuration file here, so I don't think it's the best place.

CedNaru (approved):

PR #819, opened 2025-04-22
Improve Gradle Task Runner

CedNaru:
The build task was only displaying something when the entire task was done running.
I added streaming to it so you can see what is happening in case the task takes a while. On top of that, I added a small animation to show that there is activity.
You also have now a selector to pick which task you want to run.

So far, only 3 options are present:
- Regular build (debug by default)
- Release build
- Generate Embedded JRE.

I also fixed a small bug, the editor was still complaining about the lack of bootstrap (if the project hasn't been built once yet).
It was an error I made in the PR removing the editor bootstrap, I only wanted to display an error if the game itself was launched without bootstrap, but I used the project manager hint instead of the editor hint.

https://github.com/user-attachments/assets/7ecd8f41-1df7-4bed-9b18-07592390fa2f

### DIFF
--- a/src/editor/build/gradle_task_runner.cpp
+++ b/src/editor/build/gradle_task_runner.cpp
@@ -45,6 +45,8 @@ Error GradleTaskRunner::run_task(int task_id, String& log, bool blocking) {
         case Task::GENERATE_EMBEDDED_JVM:
             args.push_back("generateEmbeddedJre");
             break;
+        default:
+            return Error::ERR_INVALID_PARAMETER;
     }
 
     String gradlew_path = get_gradlew_path();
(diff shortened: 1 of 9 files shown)


src/editor/build/gradle_task_runner.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (blocking) {
+        int exit_code;
+        Error result = OS::get_singleton()->execute(gradlew_path, args, &log, &exit_code, true, &build_mutex, false);
+        return result;
+    } else {
+        Dictionary info = OS::get_singleton()->execute_with_pipe(gradlew_path, args, /*blocking=*/false);
+
+        if (info.is_empty()) { JVM_ERR_FAIL_V_MSG(Error::ERR_CANT_CREATE, "Failed to start process"); }
+
+        log = vformat("Running %s build task...", gradlew_path);
```
chippmann, line 61: not necessarily the build task

chippmann (approved):
Just a minor nit. Otherwise looks good!

chippmann (approved):

piiertho (approved):

PR #824, opened 2025-04-23
Add support for jniLibs

chippmann:
Some libraries and user code require to load jni libs (dynamic libraries).

Our own [DexClassLoader](https://developer.android.com/reference/dalvik/system/DexClassLoader) has had only access to system libraries. Hence, any libraries one might want to bundle with the code, could not be used.

This PR makes use of the parameter `librarySearchPath` of the constructor from `DexClassLoader` and thus provides the `DexClassLoader` the ability to properly link the libraries.

An example of a library requiring such a case: https://github.com/xerial/sqlite-jdbc/blob/master/USAGE.md#how-to-use-with-android

In a plain android project, one would add the libraries to a jniLibs dir and the AGP plugin would bundle these with the apk.
But even if one, adds such libraries by hand using godot's gradle build, our classloader would not have access to it. For this, it's parent class loader would need to be the godot app class loader to which we do not have access from inside the cpp code.

Hence this PR provides a dir `jvm/android/jniLibs` to which a user can add all libraries which should be shipped with the app. And that folder is then passed in as a possible path for jniLibs to be loaded from for our `DexClassLoader`.

This PR is already in a working state but remains a draft for the following reasons:
- CPP code needs to be refined
- dir location needs to be defined
- general concept of the feature and whether and how we want to support it, needs to be discussed further

### DIFF
--- a/src/gd_kotlin.cpp
+++ b/src/gd_kotlin.cpp
@@ -218,7 +218,7 @@ String GDKotlin::copy_new_file_to_user_dir(const String& file_name) {
     unlink(file_user_path_global.utf8().get_data()); // we do not really care about errors here
 #endif
         Ref<DirAccess> dir_access {DirAccess::open(USER_DIRECTORY)};
-        dir_access->make_dir(JVM_DIRECTORY);
+        dir_access->make_dir_recursive(JNI_LIBS_PATH);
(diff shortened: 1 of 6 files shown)


CedNaru:
Closing for being a dead PR. We can discuss the feature and resume if the neet ever arise.

PR #828, opened 2025-05-09
Install dependencies before building on Linux 

### LINKED ISSUE #827: GH Actions builds without Wayland
As far as I can tell, the OS image used for GH Actions lacks `libwayland-dev`, so it should be installed before building the rest of the program.

signalno11:
Fixes #827


CedNaru:
You are adding a lot of dependency, but in Godot's own CI I only see this single install:
https://github.com/godotengine/godot/blob/19bb18716ef08b811fec330cba4d35fbdb027bcc/.github/workflows/linux_builds.yml#L111

Is there a reason for the extra stuff ? Hard Disk space is really limited in runner and we already ran into issues with it more than once. We shouldn't bloat it with so many libs if we are not sure to use them.

signalno11:
The instructions seem to be very different depending on the distro but these should be good. I think its a good idea just to make sure libalsa and libpulse are installed, so I left those, and the wayland and udev dev files. (Sidenote: we should probably be building with udev=yes)

CedNaru:
udev is already true by default, as you can see here:
https://github.com/godotengine/godot/blob/19bb18716ef08b811fec330cba4d35fbdb027bcc/platform/linuxbsd/detect.py#L48

signalno11:
Alright. The dependency is probably still smart to install, but if we haven't seen any other issues with the build yet, I suppose it's safe to simply just install wayland-devel.

CedNaru (approved):

PR #833, opened 2025-05-13
Rework Java API for Signal/Callable

### LINKED ISSUE #754: Utilize Java functional interfaces to improve signal and callable code from Java and other JVM languages
Right now, the callable interface is generated in a way that makes it very difficult to use in Java.

The Java example given [here](https://godot-kotl.in/en/stable/user-guide/signals_and_callables/#callables) is incorrect, but besides that it could also be much simplier if using Java's functional interface feature, introduced in Java 8.

## Functional Interfaces

Java has 5 built-in functional interfaces:

[`Runnable`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Runnable.html): Takes in nothing, returns nothing. A function from Void -> Void.
[`Consumer<T>`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/Consumer.html): Takes in a value `T`, returns nothing. A function from T -> Void.
[`Predicate<T>`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/Predicate.html) => Takes in a value `T`, returns a `boolean`. A function from T -> Boolean.
[`Function<T, R>`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/Function.html) => Takes in a value `T`, returns a value `R`. A function from T -> R.
[`Supplier<R>`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/Supplier.html) => Takes in nothing, returns a value `R`. A function from Void -> R.

You can also create your own functional interfaces using the `@FunctionalInterface` annotation, like so:
```java
@FunctionalInterface
interface Logger<T> {
    void log(T message);
}
```

A functional interface can only have one abstract method.

You would utilize it like so:
```java
Logger<String> logger = message -> { // Creating the functional definition
	String timestamp = java.time.LocalDateTime.now().toString();
	System.out.println("[" + timestamp + "] ERROR: " + message);
};

// logger = System.out::println; // Can reference a method to pass the data directly to as well.

logger.log("An error occurred while processing the request."); // Using it
```

You can also create functional interfaces that take in multiple parameters:
```java
@FunctionalInterface
interface Calculator {
	int calculate(int a, int b);
}
```

Utilizing it looks like:
```java
Calculator calculator = (a, b) -> a + b; // You can do this
calculator = Integer::sum; // Or this
calculator = (a, b) -> a * b; // Or whatever else you like!
```

## Outcome

Currently, the example linked above looks like this in Kotlin:
```kt
    val regularCallable = NativeCallable(myObject, MyObject::myMethod)
    val customCallable = callable1<String> { println(it) }
```

and it looks like this in Java:
```java
    NativeCallable regularCallable = Callable.create(myObject, StringNames.toGodotName("myMethod"));
    LambdaCallable1<String, Void> customCallable = LambdaCallable1.create(
            Void.class,
            String.class,
            (string) -> {
                System.out.println(string);
                return "";
            }
    );
```
I am not entirely sure if this even functions correctly, but this is what's required to get it to compile.

Using functional interfaces, the Java could instead look like:
```java
final var regularCallable = Callable.create(myObject, MyObject::myMethod);
final var customCallable = LambdaCallable1<String>.create(System.out::println);
```

Or, maybe a more useful example:
```java
final var regularCallable = Callable.create(myObject, MyObject::myMethod);
final var customCallable = LambdaCallable1<String>.create((string -> {
	System.out.println(string);
	// Do something else if you like
}));
```

This is just one example of potentially many more. The Java code can look almost identical to the Kotlin code if functional interfaces are utilized. This does require a little bit more code on the project side in order to support it (you have to specify what functional interfaces are being used, whereas in Kotlin-land it's all inferred), but the benefit is that it would allow similar usage of lambdas across the entire JVM ecosystem. For the end-user, lambdas would function identically in Kotlin as they do now, AND Java/Scala users would be able to use callables properly.

I am unsure where else this type of thing might be useful in this project in order to bring together the language APIs across the JVM ecosystem, but this one stood out to me the most.

CedNaru:


piiertho (approved):
I think we can do better for scala by making use of scala macros

Issue #269, opened 2021-09-03
Provide snapshot versions


chippmann:
Follow up of #267

It might be a good idea to build and publish snapshot versions on a regular basis.
For example a snapshot build each week.

The only module with which we would have a problem atm is the gradle plugin as the pluginPortal does not allow `-SNAPSHOT` versions. But we could provide the necessary documentation on how to use gradle plugins distributed to mavenCentral.

CedNaru:
Local deployement was made super easy ( don't need any gradle trick anymore) and PR CI expose all artifacts and project stable enough that snapshot are less necessary

PR #838, opened 2025-06-08
Update jvm artifact deployment for new maven central portal

chippmann:
This updates our jvm deployment plugin and ci/cd setup for the new maven central plugin portal as the old one is discontinued and we cannot publish updates anymore.

As there is no official gradle plugin yet to publish to the new plugin portal using gradle, this [third party gradle plugin](https://github.com/vanniktech/gradle-maven-publish-plugin) is used and configured instead as it's one ofe the mentioned third party alternatives in the [official documentation](https://central.sonatype.org/publish/publish-portal-gradle/#alternatives).

Open tasks:
- [ ] @piiertho Move the newly created secrets to org level (i do not have he rights to do so)
    - `GODOT_KOTLIN_MAVEN_CENTRAL_PORTAL_TOKEN_PASSWORD`
    - `GODOT_KOTLIN_MAVEN_CENTRAL_PORTAL_TOKEN_USERNAME`
- [ ] Remove `GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII` and rename and move `GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII_CLEANED` to org level
    - This was needed as the old ASCII private key representation is/was malformed and the old setup could just handle that. But the new plugin cannot and expects a correctly formatted key. But this should only be done once we did at least one deployment with the new setup to ensure it's working correctly. Also if you @piiertho use this key in other repos as well, i would wait with the removal until you've migrated them as well.


chippmann:
After this one is merged, we can setup the Snapshot repository as requested in #269

tommyettinger:
Good choice on VannikTech as the plugin to use, it's been my go-to for Maven Central releases for a while now and the update to the new repo was fairly painless. It's certainly *much* faster to release now! I also **had to generate new tokens** in the Sonatype web UI, the old ones would not work for me. I know some stuff has changed since the initial news that we need to migrate, and I had migrated all my projects about a month ago, so my experience may not match yours.

.github/workflows/deploy_jvm.yml
```diff
@@ -47,6 +47,14 @@ jobs:
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/master' }}
 
+      - name: Setup publish gradle properties
+        shell: sh
+        run: |
+          echo "mavenCentralUsername=$GODOT_KOTLIN_MAVEN_CENTRAL_PORTAL_TOKEN_USERNAME" >> ~/.gradle/gradle.properties
```
piiertho, line 53: Why do we write those to properties ?

.github/workflows/deploy_jvm.yml
```diff
@@ -13,10 +13,10 @@ on:
         type: string
 
 env:
-  GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII: ${{ secrets.GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII }}
+  GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII: ${{ secrets.GODOT_KOTLIN_GPG_PRIVATE_KEY_ASCII_CLEANED }} # TODO: remove old key and rename this one once we're sure that the new deployment works
```
piiertho, line 16: I would rename `GODOT_KOTLIN` prefix to `UTOPIA_RISE`

PR #850, opened 2025-08-13
Use binary Gradle distribution in project template

DSteve595:
This changes the Gradle project template to use the `-bin` distribution of Gradle, which contains just Gradle's binary, rather than `-all`, which also contains its sources. The sources aren't needed for building, so this results in a smaller download when first building the project.


chippmann (changes requested):
Personally I'm a fan of having the gradle sources out of the box as well but i see that it's not the default in most cases and that people might prefer a slimmer download. So I'm fine with this.

But in any case; the cpp templates need to be updated as well: https://github.com/utopia-rise/godot-kotlin-jvm/blob/5e6be50e511dd1516b8f051f822174c105d77579/src/editor/project/templates.h#L36

@CedNaru IIRC they should be autogenerated when doing a scons build locally and the relevant file has changed, do i remember that correctly?
If I remembered that correctly on to top of my head, then you just need to build godot locally once after editing the file which you already changed in the PR then you can just commit this header file

DSteve595:
> Personally I'm a fan of having the gradle sources out of the box as well but i see that it's not the default in most cases and that people might prefer a slimmer download. So I'm fine with this.

I am too, for sure. But most don't care, and those that do know how to change this.

Will take a look at updating the cpp template.

CedNaru:
For the headers, you can just run that script:
https://github.com/utopia-rise/godot-kotlin-jvm/blob/master/generate_templates.py

PR #859, opened 2025-09-22
Update to Godot 4.5

CedNaru:


src/editor/dialog/about_dialog.cpp
```diff
@@ -7,7 +7,9 @@
 #include "editor/strings.h"
 
 #include <editor/editor_interface.h>
+#include <editor/settings/editor_settings.h>
```
piiertho, line 10: Why imports have been added ?

piiertho: ah nevermind, just saw the header

piiertho (approved):

PR #862, opened 2025-09-23
Enable Gradle configuration cache in project template

DSteve595:
With this and the changes in https://github.com/utopia-rise/kotlin-preprocessors/pull/4, [Gradle Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html) can be enabled for downstream projects, which should give a nice speed boost since the same task is run over and over again during development.

~This does not yet enable the config cache in the template project. Need to bring in that kotlin-preprocessors change first.~ Done!


CedNaru (approved):

chippmann (approved):

CedNaru (approved):

DSteve595:
Bump @CedNaru @chippmann, would love to get this in!

PR #883, opened 2025-10-23
Rename to Godot-JVM

CedNaru:
There is probably other thing to rename like the repositories we publish to and the github repo itself.


CedNaru:
Will do it as part of the GDExtension transition

PR #917, opened 2026-05-20
Fix Bunnymark for current Kotlin API

falcon4ever:
Was trying to use the demo but it was broken for me.

Before
Screenshot 2026-05-20 at 1 14 29 PM

After
Screenshot 2026-05-20 at 1 23 18 PM


CedNaru:
Closing sorry, registration has been remade from scratch so the update for bunnymark was done differently

Issue #941, opened 2026-07-10
Every freed non-RefCounted Object leaks its JVM binding: free callback queues instance id 0


stan220:
**Version:** 0.16.2-4.6.3 (introduced by the MemoryManager rework #661, first released in 0.10.0-4.3.0; the code is unchanged on current master)

## Summary

Every non-RefCounted `Object` (i.e. every `Node`) that dies at runtime leaves its `Binding` in the JVM-side `MemoryManager.ObjectDB` forever. For non-RefCounted objects the binding holds a **strong** reference to the wrapper, so the JVM heap grows without bound in any game that keeps creating and freeing nodes. Native memory is unaffected (`Performance.OBJECT_NODE_COUNT` stays flat), only the JVM side leaks, which makes it easy to miss.

How the node is freed does not matter: `free()` from Kotlin, `queueFree()`, or engine-side deletion all leak.

## Root cause

`KotlinBindingManager::_instance_binding_free_callback` queues the dead object for the next memory sync:

```cpp
// src/jvm_wrapper/memory/memory_manager.cpp
void MemoryManager::queue_dead_object(Object* obj) {
    dead_objects_mutex.lock();
    dead_objects.push_back(obj->get_instance_id());
    dead_objects_mutex.unlock();
}
```

But Godot invokes the instance-binding free callbacks at the very end of `Object::~Object`, *after* the instance id has been cleared (core/object/object.cpp, same order in every 4.x release since at least 4.3):

```cpp
	if (_instance_id != ObjectID()) {
		ObjectDB::remove_instance(this);
		_instance_id = ObjectID();
	}
	_predelete_ok = true;

	if (_instance_bindings != nullptr) {
		for (uint32_t i = 0; i < _instance_binding_count; i++) {
			if (_instance_bindings[i].free_callback) {
				_instance_bindings[i].free_callback(_instance_bindings[i].token, this, _instance_bindings[i].binding);
```

So `obj->get_instance_id()` always returns the null ObjectID at that point. `dead_objects` fills with zeros, and `MemoryManager.removeDeadObjects` on the JVM side removes `ObjectID(0)` instead of the real ids. The real entries are never removed.

[...]

PR #945, opened 2026-07-25
Backport #942 for stable 0.16.3-4.6.3

CedNaru:
Backports the changes from #942 and bumps the release version to `0.16.3-4.6.3`.

This carries the non-RefCounted object binding leak fix and its accompanying macOS CI updates into a branch based on the previous stable release.

The leak occurred because non-refcounted bindings were not released through the appropriate lifecycle path. Users of this release receive the corrected binding cleanup behavior.

The release metadata and CI build variables now consistently use `0.16.3-4.6.3`.

Validation: `git diff --check` and `git diff --cached --check` passed; the cherry-pick applied cleanly.


stan220:
Validated this backport locally (macOS arm64, editor build of this branch: `modules/kotlin_jvm` at 34cdf0a on top of the `0.16.2-4.6.3` tag, everything else stock 4.6.3).

Repro is the public one linked from #941 — https://github.com/stan220/gkj-binding-leak-poc at 39d4bbf, unmodified, same JVM side (`godot-bootstrap.jar` 0.16.2) in both runs:

```
0.16.2-4.6.3 stock : [POC] freed=20000 nativeObjects=1472 jvmObjectDB=20039  -> LEAK CONFIRMED (~18567 zombie bindings)
0.16.2-4.6.3 + #945: [POC] freed=20000 nativeObjects=1472 jvmObjectDB=39     -> no leak detected
```

So the fix fully closes it on the stable tag: every freed non-RefCounted `Object` now leaves `MemoryManager.ObjectDB`.

I also ran my game's headless functional suite (75 tests, one shared Godot process) on the patched binary — 3 runs, 2 fully green and 1 with a single test failing; that test passes 3/3 in isolation on both the patched and the stock binary, and the suite is green on stock, so I can't attribute it to this change yet. I'll report back if it turns out to be reproducible.

One question: is a patch release (editor binary + export templates for `0.16.2-4.6.3`) planned once this lands, or should people on stable build the editor themselves in the meantime?

CedNaru:
Next release is the 4.7 one, I'm just having that PR to generate the artifacts so you can pick the ones you need, or are you seeing it's not just you who needs that ?

stan220:
Could you make a downloadable stable build? Tagging this branch is enough — its `trigger_on_tag.yml` already declares `0.16.2-4.6.3` / `4.6.3-stable` and produces the full editor + templates. Refreshing the binaries on the existing release works too; the fix is native-only, so the jar and Gradle plugin are unaffected.

CedNaru:
Ha yeah my bad, I forgot that I opened the PR targeting master so it doesn't work...
I can't really just tag it with the same version number as before because its builds/publish everything, including the JVM artifacts and you can't override them once published once. There is no "C++ only" release
I guess I could make a quick `0.16.3-4.6.3` release.

stan220:
Sounds good, thanks. Heads-up for the tag: the branch still hardcodes 0.16.2 in `src/version.h`, `trigger_on_tag.yml` (`godot-kotlin-jvm-version` + `build-version`) and `kt/gradle/libs.versions.toml`, so those need bumping to 0.16.3 or the release will be labelled with the old version.

PR #967, opened 2026-08-29
Add group annotations

### LINKED ISSUE #605: Adding missing group Godot annotations
While working on manager classes, it is useful to group related properties, using the `@export_group` (https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html#grouping-exports) annotation. 

However, at the moment, this annotation is not available for Kotlin. It'd be nice to have it!

```kt
@RegisterClass
class Simple : Node() {
    @ExportGroup("Bonus Properties")
    var bonusMultiplier: Double
    var bonusPoints: Int
}
```

### LINKED ISSUE #965: @Export var loots = VariantArray<LootData>() crashes editor on adding elements
Version: Godot 4.7.2.stable.jvm.0.17.1
JDK:21

Description:
In version 4.6.3, exporting VariantArray<LevelData> worked fine – the inspector showed Array[LevelData] and I could add elements without issues.
After upgrading to 4.7.2, the shows only Array[] (no type hint). When I click the "add element" button and assign any resource, the editor crashes immediately.

**Steps to reproduce:**

1. Create a Kotlin script (Node) with:

kotlin
`@Export var loots = VariantArray<LootData>()`

2. Attach it to a node in a scene.

3. In the inspector, see Array[] instead of Array[LootData].

4. Click [Add Element] to add an element, then drag a resource into it – editor crashes.

**Additional info:**

Single resource export (@Export var loot: LootData? = null) works fine.

This used to work in 4.6.3 (with the old .gdj system).

Attempting to wrap the array inside a custom Resource also crashes, so that workaround is not valid.

Image

Image

### LINKED ISSUE #966: Duplicate property registration when property name starts with capital letter
Versions: 0.17.1-4.7.2, 1.0.0-dev3

When exporting a property, if your property starts with a capital letter the build fails with `Cannot infer type for type parameter 'P'. Specify it explicitly.`

I had a camera property in one of my scripts that was still PascalCase after porting from C#:
```kotlin
@Export
lateinit var Camera: Camera3D
```
Which results in a duplicate property registration in generated registrar class after building:
```kotlin
property(CharaProto::Camera, OBJECT, "Camera3D", NODE_TYPE, "Camera3D", PropertyUsageFlags.DEFAULT)
property(CharaProto::camera, OBJECT, "Camera3D", NODE_TYPE, "Camera3D", PropertyUsageFlags.DEFAULT)
```

If I rename the property from "Camera" -> "camera", the `CharaProto::Camera` property registration goes away and the build succeeds.

This seems to be caused by a change in 0.17.1. This script was working in 0.17.0.

CedNaru:
Implement #605
Fix #965, #966

image

### DIFF
--- a/.github/actions/detect-path-changes/action.yml
+++ b/.github/actions/detect-path-changes/action.yml
@@ -78,6 +78,12 @@ runs:
           exit 0
         fi
 
+        if ! git merge-base --is-ancestor "$BASE" "$HEAD_SHA"; then
+          echo "Base SHA $BASE is not an ancestor of $HEAD_SHA (rewritten branch) - forcing full run."
+          echo "changed=true" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+
         echo "Base: $BASE"
         echo "Head: $HEAD_SHA"
         echo "Paths: ${PATH_ARGS[*]}"
--- a/kt/godot-library/godot-bootstrap-library/src/main/kotlin/godot/registration/KtClass.kt
+++ b/kt/godot-library/godot-bootstrap-library/src/main/kotlin/godot/registration/KtClass.kt
@@ -19,7 +19,7 @@ data class KtClass<T : KtObject>(
     val kClass: KClass<out KtObject>,
     private val _registeredSupertypes: List<String>,
     val constructor: KtConstructor<T>?,
-    private val _properties: Map<String, KtProperty<T, *>>,
+    private val _properties: List<KtProperty<T, *>>,
     private val _functions: Map<String, KtFunction<T, *>>,
     private val _notifications: List<KtNotification<T>>,
     private val _signalInfos: Map<String, KtSignalInfo>,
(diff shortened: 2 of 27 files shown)


.github/actions/detect-path-changes/action.yml
```diff
@@ -78,6 +78,12 @@ runs:
           exit 0
         fi
 
+        if ! git merge-base --is-ancestor "$BASE" "$HEAD_SHA"; then
```
piiertho, line 81: What is this for ?

kt/godot-library/godot-bootstrap-library/src/main/kotlin/godot/registration/KtClass.kt
```diff
@@ -31,7 +31,7 @@ data class KtClass<T : KtObject>(
     val functions: Array<KtFunction<T, *>>
         get() = _functions.values.toTypedArray()
     val properties: Array<KtProperty<T, *>>
-        get() = _properties.values.toTypedArray()
+        get() = _properties.toTypedArray()
```
piiertho: Do we still need it ?

piiertho (changes requested):

piiertho (approved):

PR #969, opened 2026-09-01
Add custom api generator.

CedNaru:
Provide your own api.json from a different godot with custom module or other gdextension, or even a newer version and you will be able to use matching Kotlin wrappers.


docs/src/doc/reference/gradle-plugin/packaging-and-tasks.md
```diff
@@ -8,6 +8,31 @@ These settings affect how the project is packaged, which optional runtime helper
...
+    isCustomApiEnabled.set(true)
+}
+```
+
+The plugin reads `api.json` from the root Gradle project by default. Set `apiJsonFile` when the file is elsewhere; it has no effect while `isCustomApiEnabled` is `false`:
+
+```kotlin
+godot {
+    isCustomApiEnabled.set(true)
+    apiJsonFile.set(file("godot-api/custom-api.json"))
```
piiertho, line 26: Maybe we can merge both properties and say custom api is enabled when apiJsonFile is set ?

piiertho (changes requested):

docs/src/doc/reference/gradle-plugin/packaging-and-tasks.md
CedNaru: What I had in mind was that you could set the api.json but still wish to run some test using the official published library for some tests instead of the custom one, so you could do it by just flipping a flag instead of erasing a path to a file

piiertho (approved):

PR #983, opened 2026-09-15
Yet again the memory model is redone

CedNaru:
Reworks how a Godot object and its JVM counterpart keep each other alive, plus the correctness and tooling work that came out of it.

## Memory model

- **A scripted `RefCounted` is now born weak and stays weak for its whole life.** Strength is no longer expressed by swapping the reference, but by a second reference — a *pin* — held beside it. A holder other than the JVM create the pin, if only the JVM is left, the pin is deleted. The handle the rest of the binding reads never moves, so no thread can catch it mid-swap.
- **Demotions are deferred, and for a busy object never happen at all.** Handing an object back to JVM-only ownership marks it at one synchronization and only acts at the next, and any use in between clears the mark. An object Godot touches every frame therefore keeps its pin forever: the promote/demote ping-pong that used to cost a pair of JNI operations per object per frame stops happening rather than merely getting cheaper. Objects that genuinely go quiet pay one extra frame of latency, which is invisible next to how long the collector takes to notice them anyway.
- **Reference-count decrements wait a frame**, so an object being passed around between threads cannot be destroyed in the window where one side has let go and the other has not yet taken hold.
- **All synchronization happens on the main thread**, which is what lets these transitions run without locking against the engine.
- **Object creation states its intent.** A plain `Object`, a `RefCounted` shared with the engine, and a `RefCounted` the JVM alone owns are now three separate, named constructions instead of one path that built the wrong kind of reference and immediately corrected it.
- **A scripted object is one allocation fewer.** The JVM handle lives inside the instance data rather than beside it, so the per-object footprint and the pointer chasing on every call both shrink.

## Correctness

- **JNI references that were never released** — both long-lived ones held for registration data and short-lived ones accumulating inside calls — are now accounted for.
- **Strings crossing the boundary were being read with the wrong encoding.** Anything outside the Basic Multilingual Plane (emoji, rarer CJK, historic scripts) and any string containing an embedded null was silently corrupted in one direction or the other. Both directions now agree. Harness tests cover it, for short strings and for ones large enough to take the separate long-string path.
- **Per-thread native storage is released when its thread ends**, instead of living until the process exits.

## Structure

- The per-thread exchange machinery is split into three pieces along ownership lines — the argument stack, the shared buffer borrowed from the JVM, and the long-string queue — each shaped by what it actually owns rather than all three sharing one shape.
- `cpp/engine/` sources are wrapped in the `godot` namespace, in line with the rest of the tree.

## Tooling, tests and docs

- Benchmark project became a stress test project and redone from 0
- Bunnymark builds and runs again.
- `--jvm-custom-args` accepts several arguments in one flag, quoted the way a shell quotes them — the same rule the JDK itself uses for `JDK_JAVA_OPTIONS`. Documented in the runtime configuration reference.

image


docs/src/doc/contribute/how-it-works/memory-management.md
```diff
@@ -1,58 +1,177 @@
...
+    [*] --> Pinned: created while Godot owns it too
+    [*] --> Weak: created by the JVM alone
+    Weak --> Pinned: Godot takes a reference
+    Pinned --> Marked: the count falls back to one
+    Marked --> Pinned: the object is used again
+    Marked --> Weak: a whole frame passes unused
+    Weak --> [*]: collected, native reference released
+```
+
+Read the two arrows out of `Marked` and the behaviour is easy to state:
```
chippmann: `Marked` does not exist as a rendered name in the mermaid diagram. Use `Pinned, demotion marked` instead

